Issue #338, opened 2026-06-13
OAuth callback consumes Link state before fallible completion work


ftchvs:
## Summary

The Hosted Link OAuth callback consumes the pending one-time `state` before it finishes Plaid completion lookup, public-token exchange, account lookup, and local item storage. If any of those post-callback steps throws, the browser receives a 500 error and the same callback URL cannot be retried because the state has already been removed.

## Affected area/files

- `Sources/PlaidBarServer/Routes/LinkRoutes.swift`
- `Sources/PlaidBarServer/Auth/PendingLinkSessionStore.swift`
- `Tests/PlaidBarServerTests/PlaidBarServerTests.swift`

## Reproduction or static proof

Static path on current `main`:

1. `OAuthCallbackRoute.handleCallback` reads `state` and immediately calls `pendingLinkSessions.consume(state:)` before entering the Plaid completion work.
2. `PendingLinkSessionStore.consume` removes the session from `sessions` and persists that removal before returning it.
3. The callback then calls `plaidClient.getLinkToken`, `exchangePublicToken`, `getAccounts`, and `tokenStore.saveItem`.
4. Any thrown error in those calls returns the callback error page with HTTP 500.
5. Reloading the same browser callback URL then hits the `consume` guard with no stored session and returns “Missing or expired link session state”, so the completed Link session is no longer recoverable through retry.

Relevant lines checked:

```console
$ nl -ba Sources/PlaidBarServer/Routes/LinkRoutes.swift | sed -n '108,178p'
111  guard let stateParam = request.uri.queryParameters.get("state"),
112        let pendingSession = await pendingLinkSessions.consume(state: String(stateParam))
...
124  let publicTokenResults = try await publicTokenResults(from: request, pendingSession: pendingSession)
...
145  try await exchangeAndStore(publicTokenResult: publicTokenResult)
...
153  } catch {
154      return Response(status: .internalServerError, ...)
```

```console
$ nl -ba Sources/PlaidBarServer/Auth/PendingLinkSessionStore.swift | sed -n '41,49p'
41  func consume(state: String) -> PendingLinkSession? {
43      guard let session = sessions.removeValue(forKey: state),
48      persist()
49      return session
```

Current tests cover the happy path (`OAuth callback stores Hosted Link institution names`) and pending-session single-use behavior, but there is no regression coverage for a transient Plaid/storage failure after the state has been consumed.

## Expected vs actual behavior

[...]

Issue #309, opened 2026-06-13
Primary CI and review-gate workflows are disabled


ftchvs:
## Summary

The repository's primary CI and required review-gate workflows are present on `main`, but GitHub Actions currently reports them as `disabled_manually`. As a result, new pushes/PRs are not automatically getting the macOS build/test/smoke/package validation or the configured review/merge gates.

## Affected area/files

- `.github/workflows/ci.yml`
- `.github/workflows/claude-code-review.yml`
- `.github/workflows/codex-code-review.yml`
- `.github/workflows/otto-openclaw-merge-gate.yml`
- Repository workflow state in GitHub Actions

## Reproduction or static proof

From a clean checkout of `origin/main`:

```console
$ gh workflow list --repo ftchvs/PlaidBar --all
Changelog active 248187167
CI disabled_manually 248151304
Claude Code Review disabled_manually 248185546
Claude Code active 248185592
Codex CI active 294988346
Codex Code Review disabled_manually 295004382
Otto OpenClaw Merge Gate disabled_manually 291158689
```

The disabled `CI` workflow is the one whose `ci.yml` file is configured for both pushes and pull requests to `main` and runs the release build, strict-concurrency build, `swift test`, sandbox smoke test, and app bundle validation.

## Expected vs actual behavior

Expected: the `CI`, `Claude Code Review`, `Codex Code Review`, and `Otto OpenClaw Merge Gate` workflows should be enabled so pushes/PRs targeting `main` receive the checks described by the workflow files and PR template.

Actual: GitHub marks those workflows as manually disabled, so the repository configuration silently bypasses the current safety checks until they are re-enabled.

## Risk/impact

Bugs in the macOS build, strict Swift concurrency, tests, sandbox smoke flow, bundle packaging, or autonomous merge handoff can reach `main` without automated proof. This also contradicts the PR template's requirement that CI, Claude review, Codex review, and OpenClaw gates be green before merge.

## Suggested fix direction

Re-enable the disabled workflows after confirming whether they were intentionally paused. If they were disabled because of earlier failures, keep them enabled and fix the underlying workflow failures so missing checks are visible on PRs.

## Evidence commands/results

- `gh workflow list --repo ftchvs/PlaidBar --all` showed the disabled workflow states above.
- `gh workflow view 248151304 --repo ftchvs/PlaidBar --yaml` confirmed the disabled `CI` workflow is the active `ci.yml` build/test/smoke/package workflow on `main`.
[...]

PR #348, opened 2026-06-13
fix: preserve OAuth callback retry state

ftchvs:
## Summary
- Keeps pending Hosted Link state retryable until callback completion succeeds, instead of consuming it before fallible completion work.
- Adds a short in-flight completion lease plus completed-result progress so concurrent/retried callbacks do not replay already stored Hosted Link results.
- Covers the retry behavior with synthetic server tests only.

## Changes
- `Sources/PlaidBarServer/Auth/PendingLinkSessionStore.swift`: adds completion lookup/release, success consume, and a backward-compatible persisted completed-result count.
- `Sources/PlaidBarServer/Routes/LinkRoutes.swift`: consumes pending state only after update/add completion succeeds; releases retry state on failure; skips completed result indexes on retry.
- `Tests/PlaidBarServerTests/PlaidBarServerTests.swift`: adds fail-then-retry coverage plus multi-result partial-progress replay coverage using synthetic IDs/tokens.

## Test plan
- [x] `git diff --check`
- [x] `swift test --filter 'PlaidBarServerTests/oauthCallback' --skip-update --disable-keychain -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` (4 tests passed)
- [x] `swift test --filter 'PlaidBarServerTests' --skip-update --disable-keychain -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` (68 tests passed)
- [x] `swift build --target PlaidBar --skip-update --disable-keychain -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors`
- [x] Diff-only privacy/security scan for Plaid secrets, public/access tokens, raw Plaid payloads, real account/transaction IDs, balances, local SQLite data, logs, screenshots/generated artifacts, app-target Plaid access, and local-first boundary drift. Findings were expected server/test symbols and synthetic fixtures only.
- [x] `codex review --uncommitted`; first pass found partial-result replay risk, fixed here; rerun reported no actionable findings and its own server test run passed.

## Privacy/security notes
- No SwiftUI app-target changes and no direct Plaid app access.
- No real Plaid credentials, provider tokens, raw Plaid payloads, real account IDs, transaction IDs, balances, local data, logs, screenshots, or generated artifacts added.
- Persisted retry progress is only an integer completed-result count; public/access tokens are not added to pending-session storage.

## CI/check state
- Draft PR because primary CI/review-gate workflows are disabled under #309 and checks may be missing.

## Risks
- This preserves retryability for failures before completed result storage and prevents completed multi-result replay. A failure after Plaid public-token exchange succeeds but before local item storage/progress marking remains a harder edge; fully recovering that would require a larger token-handoff design.

Refs #338

### DIFF
--- a/Sources/PlaidBarServer/Auth/PendingLinkSessionStore.swift
+++ b/Sources/PlaidBarServer/Auth/PendingLinkSessionStore.swift
@@ -47,23 +47,36 @@ actor PendingLinkSessionStore {
         purgeExpired()
         let currentDate = now()
         guard let session = sessions[state],
-              currentDate.timeIntervalSince(session.createdAt) <= ttl,
-              completionLeases[state] == nil
+              currentDate.timeIntervalSince(session.createdAt) <= ttl
         else {
+            return nil
+        }
+        // A live lease means another handler is mid-flight: refuse, preserving
+        // single-flight. The lease is reclaimable only once it is older than
+        // `completionLeaseTTL`, i.e. the original handler is presumed dead (it
+        // would otherwise have called releaseCompletion/consume). This bounds
+        // recovery without dropping a fresh in-flight lease — unlike purging by
+        // TTL, which could race a still-running handler.
+        if let leaseStartedAt = completionLeases[state],
+           currentDate.timeIntervalSince(leaseStartedAt) <= completionLeaseTTL {
             return nil
         }
         completionLeases[state] = currentDate
         return session
     }
 
-    func markPublicTokenResultStored(state: String) {
+    /// Records that the result with this stable `identity` has been consumed
+    /// (its single-use Plaid public token exchanged), so a retry never replays
+    /// it. `identity` is a token-free fingerprint (see `resultIdentity`), never
+    /// the raw public token — pending-session storage holds no Plaid tokens.
+    func markResultCompleted(state: String, identity: String) {
         purgeExpired()
         guard var session = sessions[state],
               now().timeIntervalSince(session.createdAt) <= ttl
         else {
             return
         }
-        session.completedPublicTokenResultCount += 1
+        session.completedResultIdentities.insert(identity)
         sessions[state] = session
         persist()
     }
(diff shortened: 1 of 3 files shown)


ftchvs:
@codex please review.

ftchvs:
Addressed all three Codex findings in 310555c:

- **P1 (replay of consumed public token):** the result is now marked consumed the instant `exchangePublicToken` returns — a post-exchange handoff barrier in `exchangeAndStore`, before the fallible `getAccounts`/`saveItem` steps. A retry after a post-exchange failure skips the spent token instead of replaying it (which Plaid rejects). New test `oauthCallbackDoesNotReplayConsumedPublicTokenAfterPostExchangeFailure` asserts the token is exchanged exactly once.
- **P2 (ordinal vs identity progress):** `PendingLinkSession` now records a `Set<String>` of token-free identities (SHA-256 of the public token; institution id for token-less update results) instead of a count, so a reordered multi-item retry skips exactly the already-stored results. Backward compatible — older persisted sessions decode to an empty set.
- **P2 (lease expiry mid-flight):** completion leases now live as long as their session and are cleared only by `releaseCompletion`/`consume`; `purgeExpired` no longer drops a fresh in-flight lease. The short `completionLeaseTTL` survives only as a reclaim clamp in `beginCompletion` for a presumed-dead handler. New tests `completionLeaseHoldsUntilReleased` and `completionLeaseReclaimableWhenStale`.

Full strict-concurrency build clean; suite **618** tests pass.

@codex please review.

ftchvs:
Addressed both round-2 findings in ae4cca4:

- **P1 (false success / mark-before-save — LinkRoutes:172):** the callback now classifies each result: `notExchanged` (token unspent → release + retryable), `stored` (persisted → mark finalized, success), or `spentButNotStored` (exchange succeeded but save failed → mark finalized so the dead token is never replayed, but return a **re-link error**, not a success page). Additionally, `getAccounts` is now non-fatal — it only enriches institution metadata, so a transient failure there salvages the item from the result's own institution and still stores it. New test `oauthCallbackSalvagesItemWhenGetAccountsFails` covers that.
- **P2 (lease reclaim race — PendingLinkSessionStore:64):** removed the timer-based reclaim entirely. A lease is now released only by `releaseCompletion`/`consume`, or it disappears with the session at the session TTL (an in-memory lease for a crashed handler clears on process restart). No slow-but-alive handler can be raced. Test renamed to `completionLeaseIsNotReclaimedByTimer`.

**Residual, flagged for you:** a `saveItem` failure *after* a successful exchange is still an unrecoverable item needing a manual re-link — fully recovering it requires the larger token-handoff design noted in #338. This change makes that case honest (error + no replay + no false success) rather than solving it end-to-end.

Full strict build clean; suite **618** tests pass.

@codex please review.

PR #351, opened 2026-06-14
feat: fix Hosted Link 500 + detachable desktop window + consumer-prod foundation

ftchvs:
Three lanes (dynamic workflow: author → adversarial review; all **APPROVE-WITH-NITS, no must-fix**). **Not merged — your review/merge call.**

## 1. Fix sandbox/consumer Plaid link **500** ✅ verified live
`PlaidClient` sent a top-level OAuth `redirect_uri` **and** Hosted Link → Plaid `INVALID_FIELD` ("OAuth redirect URI must be configured in the developer dashboard") → opaque empty **500** (the Connect-screen error). Fix: `redirectUri` optional + omitted; dropped from both Hosted Link bodies; `LinkRoutes` surfaces Plaid's real `error_code`/`error_message` (secret-free). **Verified:** `POST /api/link/create` → **200** + `https://secure.plaid.com/hl/…`. Same fix applies to production.

## 2. Detachable desktop dashboard (AND-384) — opt-in
`MainPopover` hosted in a floating, draggable `NSPanel` — "not linked to the nav menu, drag across desktop." Opt-in Detach control + Settings toggle; re-dock; menu-bar item raises the panel; one shared `AppState`; Reduce-Motion aware. **Default popover unchanged** (render-verified).
⚠️ Detach *interaction* needs a hands-on pass (drag / app-switch survival / re-dock / ⌘-shortcuts + Esc / VoiceOver) — headless can't exercise NSPanel lifecycle.

## 3. Consumer production-readiness **foundation** (inert)
`deployment` seam (`.local` default), `AccessTokenResolver` (local + fail-closed stub), `EntitlementMiddleware` shell, `Entitlement` model, **`consumer-production-checklist.md`**. Local mode byte-for-byte unchanged.
🔒 Gated (NOT here): hosted-bridge deploy, prod Plaid creds, prod redirect/Hosted-Link registration, Plaid production approval, live Stripe billing.

## Validation
Strict build clean; `swift test` **658**; `/api/link/create` **200 + Hosted Link URL** (live sandbox); dashboard unchanged light/dark.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -24,6 +24,7 @@ final class AppState {
         static let setupCompletedOnce = "setup.completedOnce"
         static let setupCompletedContextPrefix = "setup.completedOnce.context"
         static let lastTransactionCacheContext = "cache.lastTransactionCacheContext"
+        static let dashboardDetached = DetachedDashboardPreferences.detachedStorageKey
     }
 
     // MARK: - State
@@ -51,6 +52,20 @@ final class AppState {
         }
     }
     var isPopoverPresented = false
+
+    /// When true, the dashboard lives in a floating desktop window instead of
+    /// the menu-bar popover (AND-384). Persisted so the window reopens on the
+    /// next launch. While detached, a click on the menu-bar item raises the
+    /// floating window rather than opening the popover; re-docking flips this
+    /// back to false and the popover resumes. Mirrors the `dashboard.detached`
+    /// `@AppStorage` key the Settings toggle writes, so the toggle and the
+    /// in-dashboard pin/re-dock controls stay in sync.
+    var isDashboardDetached = false {
+        didSet {
+            guard isDashboardDetached != oldValue else { return }
+            UserDefaults.standard.set(isDashboardDetached, forKey: Keys.dashboardDetached)
+        }
+    }
 
     /// Persisted across launches so configured installs boot straight into
     /// the dashboard instead of flashing first-run onboarding until the
@@ -235,6 +250,10 @@ final class AppState {
         loadPersistedBalanceHistory()
         // Launch at login
         launchAtLogin = LaunchService.isEnabled
+        // Detached-dashboard intent (AND-384)
+        if defaults.object(forKey: Keys.dashboardDetached) != nil {
+            isDashboardDetached = defaults.bool(forKey: Keys.dashboardDetached)
+        }
     }
 
     // MARK: - Computed
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -250,10 +250,19 @@ final class AppState {
         loadPersistedBalanceHistory()
         // Launch at login
         launchAtLogin = LaunchService.isEnabled
-        // Detached-dashboard intent (AND-384)
-        if defaults.object(forKey: Keys.dashboardDetached) != nil {
-            isDashboardDetached = defaults.bool(forKey: Keys.dashboardDetached)
-        }
+        // Detached-dashboard intent (AND-384). A headless snapshot render
+        // ignores the persisted intent so the popover-capture path stays
+        // deterministic regardless of host/CI defaults — otherwise a stale
+        // `dashboard.detached = true` would spawn the floating window and
+        // intercept the renderer's popover open. The stored value is left
+        // untouched (no write-back) so the real user preference survives.
+        let storedDetached = defaults.object(forKey: Keys.dashboardDetached) != nil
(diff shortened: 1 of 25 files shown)


ftchvs:
@codex please review.

ftchvs:
Addressed all three Codex findings in 6afc4d4:

- **P1 (raw Plaid error payload → SwiftUI):** `LinkRoutes.linkErrorMessage` no longer accepts or echoes Plaid's free-form `error_message`. Known `error_code`/`error_type` identifiers now map to a curated, locally-authored description; unknown codes degrade to generic guidance carrying only the bounded code. Closes the `AGENTS.md` high-priority leak path through `ServerClient` → `AppState.error`.
- **P2 (detached restore inside lazily-mounted popover):** moved the restore-at-launch `.task` and the `isDashboardDetached` `.onChange` from `MainPopover` content onto the always-mounted menu-bar label. A persisted `dashboard.detached = true` now reopens the floating window at launch, and Settings-only toggles take effect immediately.
- **P2 (inspector width in detached minimum):** added `PopoverGeometry.detachedMinContentWidth(isInspectorOpen:)` and wired it into `MainPopover`; the resizable window's floor now reserves the inspector column (982pt) so selecting an account near the minimum width no longer clips the inspector.

Strict-concurrency build clean; full suite **663** tests pass (added coverage for the allowlist mapping, the new geometry helper, and snapshot-render determinism).

@codex please review.

ftchvs:
Addressed Codex's round-2 review in 3b905d8:

- **P1 (validate error identifiers — LinkRoutes:131):** added `sanitizedErrorIdentifier` — only a code matching Plaid's stable enum shape (uppercase/digits/underscore, ≤64 chars) is surfaced; prose, request data, or an over-long string collapses to `PLAID_ERROR`. New test `nonIdentifierCodeIsSanitized` proves `request_id`/URLs never leak.
- **P2 (click interceptor on lazy popover — PlaidBarApp:83):** moved the `isPopoverPresented` interceptor onto the always-mounted menu-bar label, alongside the restore/sync observers. A status-item click while detached (before the popover ever mounted) now raises the floating window instead of opening the popover.
- **P2 (panel resize floor — DetachedDashboardWindowController:147):** the controller now tracks `dashboard.selectedAccountId` and raises the NSPanel `contentMinSize` (growing the frame if needed) to the three-column 982pt floor when an account is selected — the SwiftUI `frame(minWidth:)` alone didn't move the panel's real floor. `minContentWidth(isInspectorOpen:)` shares `PopoverGeometry` so both floors agree.
- **P2 (stale hide completion — DetachedDashboardWindowController:96):** a `presentationGeneration` counter (bumped on show/raise) makes a pending fade-out skip its `orderOut` when superseded, so re-enabling mid-hide can't leave a detached-but-invisible state.

**Needs your call — P1 redirect_uri for OAuth Hosted Link (PlaidClient:70):** Codex is correct that OAuth/app-to-app institutions use the top-level `redirect_uri` for the bank-auth handoff. But re-adding it *unconditionally* would re-introduce the exact `INVALID_FIELD` 500 this PR fixes for any setup whose `redirect_uri` isn't registered in the Plaid dashboard. The correct fix is to thread a *dashboard-registered* OAuth redirect URI from config and include it only when set — but the value and its config surface are a Plaid-dashboard decision. I left the code omitting it (keeps non-OAuth Hosted Link working, no 500) and updated the comment to capture the constraint. Want me to add a `PLAID_OAUTH_REDIRECT_URI` config knob (defaulting unset → current behavior) so OAuth institutions work for users who register one?

Strict build clean; suite **664** tests pass.

@codex please review.

ftchvs:
Addressed Codex round-3 in 6c367e3:

- **P1 (reject unknown error identifiers):** `linkErrorMessage` now surfaces a code only when it's in the curated allowlist; any other value — including an enum-shaped but unmapped token — collapses to the generic `PLAID_ERROR`. New/updated tests confirm `REQ_9F3A1B2C`-style tokens never leak.
- **P2 (detached-restore appearance):** the controller now resolves `AppAppearanceMode` (CLI `--appearance` → stored Light/Dark preference → follow system) for both the panel chrome and the hosted root, so a launch-restored detached window honors the user's setting before any `.appliesAppAppearance()` scene mounts.

**Still flagged for you — OAuth redirect_uri (P1, recurring):** I'm intentionally not implementing this without your call. It needs a dashboard-registered OAuth redirect URI threaded from config and included only when set; re-adding `redirect_uri` unconditionally would re-break the `INVALID_FIELD` 500 this PR fixes. If you confirm you want a `PLAID_OAUTH_REDIRECT_URI` config knob (defaulting unset → current behavior), I'll add it in a focused follow-up. This is a Plaid-dashboard product decision rather than a pure code fix.

Strict build clean; suite **664** tests pass.

@codex please review.

PR #383, opened 2026-06-15
AND-414: Add managed bank-link broker enforcement

ftchvs:
## Summary

Implements AND-414 as the next stacked managed-link slice on top of #382:

- adds a backend-only managed link broker endpoint at `/api/link/managed/create`
- adds `/api/link/managed/entitlement` for safe plan/status/usage summaries
- enforces managed institution limits server-side before Plaid Link creation and again before OAuth callback token exchange
- records managed vs BYO item origin, defaulting existing/local rows to BYO so demo/BYO mode remains ungated
- keeps Plaid public tokens/access tokens/provider secrets entirely server-side; managed create response returns only Hosted Link URL + entitlement summary
- documents managed bridge responsibilities and token-handling boundaries

Stacking note: this PR is based on `otto/and-411-managed-link` / #382 because the broker depends on the managed OAuth/webhook/status stack.

## Verification

```text
git diff --check
swift test --disable-sandbox --skip-update --filter 'ConsumerFoundationTests|SubscriptionPlanTests|LinkTokenRequestTests'
# Test run with 55 tests passed

swift build --target PlaidBarServer --skip-update -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors
# Build of target: 'PlaidBarServer' complete
```

## Privacy / security

- No Plaid access tokens, public tokens, client secrets, raw provider payloads, account IDs, balances, transactions, local SQLite data, logs, or screenshots are exposed to the SwiftUI app.
- Managed entitlement responses are limited to plan/status/count/reason fields.
- BYO/demo/local mode remains ungated.

Linear: AND-414

### DIFF
--- a/Sources/PlaidBarCore/Models/SubscriptionPlan.swift
+++ b/Sources/PlaidBarCore/Models/SubscriptionPlan.swift
@@ -120,3 +120,75 @@ public struct InstitutionUsage: Sendable, Equatable {
         return "\(connectedCount) \(institutionWord) connected"
     }
 }
+
+public enum ManagedLinkBlockReason: String, Codable, Sendable, Equatable {
+    case managedBridgeUnavailable = "managed_bridge_unavailable"
+    case subscriptionRequired = "subscription_required"
+    case subscriptionDegraded = "subscription_degraded"
+    case institutionLimitReached = "institution_limit_reached"
+
+    public var safeMessage: String {
+        switch self {
+        case .managedBridgeUnavailable:
+            "Managed bank linking is available only on the hosted bridge path. Demo and BYO-key linking stay available locally."
+        case .subscriptionRequired:
+            "Choose an eligible managed plan before connecting a bank through VaultPeek."
+        case .subscriptionDegraded:
+            "Your subscription needs attention before a new managed bank can be connected. Local data is not deleted."
+        case .institutionLimitReached:
+            "Your managed institution limit has been reached. Disconnect an institution or choose a higher plan before connecting another bank."
+        }
+    }
+}
+
+/// Secret-free entitlement and usage state returned to clients. This object is
+/// intentionally limited to plan/status/counts/reasons: never add Plaid tokens,
+/// provider secrets, raw provider payloads, account IDs, balances, transactions,
+/// or local database paths.
+public struct ManagedLinkEntitlementSummary: Codable, Sendable, Equatable {
+    public let plan: SubscriptionPlan
+    public let status: BillingSubscriptionStatus?
+    public let institutionLimit: Int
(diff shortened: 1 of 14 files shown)


ftchvs:
## Automated senior review — HOLD (gated; one enforcement bug to fix before un-gating)

Reviewed via an adversarial multi-agent pass (5 dimensions; each finding attacked by 2 skeptics). **Verdict: hold — correctly product-gated foundation seam** (AND-414: "build after approval gates resolve"; BYO/demo stays ungated). **It does NOT change local-first/BYO/demo behavior today** — default `.local` mode forces `blockReason = .managedBridgeUnavailable`, `EntitlementMiddleware.evaluate` is a no-op, and all `/api/*` routes sit behind `APITokenMiddleware` on the 127.0.0.1 bind. Secret/token/PII boundary is **clean** (entitlement responses carry only plan/status/count/limit/trial — no Plaid tokens, raw payloads, account IDs, or balances reach the app).

### Must fix before this is ever un-gated
- **P1 — Managed institution-limit enforcement is a non-atomic TOCTOU race.** `ManagedLinkEntitlementService.summary()` reads `activeInstitutionCount(origin: .managed)`, and the OAuth callback inserts via a *separate* `TokenStore.saveItem` actor call; `PendingLinkSessionStore`'s single-flight lease is keyed per `state`, so two concurrent managed sessions at `count == limit-1` both pass the `count < limit` check and both insert (cap overshoot by N-1). No DB unique constraint backstops it (`institution_id` is plain non-unique). The serial OAuth re-check only closes the *serial* window — contrary to `managed-link-broker-enforcement.md`'s safety claim. **Fix:** a single `TokenStore` actor method (`saveManagedItemEnforcingLimit`) that re-reads the managed count and inserts uninterrupted (throws `.limitReached`), and/or a unique partial index on managed-origin rows. Add a concurrency regression test (seed `limit-1`, two concurrent callbacks via `withTaskGroup`, assert `<= institutionLimit`).

### Follow-up (P3)
- `AddOriginToItems` adds `origin` nullable with no backfill; `activeInstitutionCount`'s raw-column filter disagrees with `ItemModel.itemOrigin`'s NULL→BYO mapping. Harmless today (only the `.managed` branch runs), but backfill existing rows to `'byo'` before any BYO-counting feature relies on the query.

_Automated PR-review loop. Not merging — gated, pending the product/sequencing decision._

ftchvs:
✅ The managed-cap TOCTOU race flagged above is fixed in #386 (atomic FIFO-gated `TokenStore.saveManagedItemEnforcingLimit` + concurrency regression test). Still hold this PR for the gating decision; #386 stacks on #384.

PR #384, opened 2026-06-15
AND-348: Add Stripe subscription entitlement seam

ftchvs:
## Summary

Implements AND-348 as a stacked Stripe-entitlement seam on top of #383 / AND-414:

- adds Stripe-shaped checkout and customer portal session endpoints under `/api/billing`
- adds a normalized metadata-only Stripe webhook endpoint that updates subscription status/plan/trial state idempotently by event ID
- adds `/api/billing/entitlement` returning plan, subscription status, institution limit, active managed institution count, trial state, allowed premium features, and managed-link eligibility
- wires billing entitlement into the AND-414 managed-link count/limit path so canceled/degraded subscriptions block future managed linking without deleting data
- documents Stripe/entitlement privacy boundaries and production follow-ups

Stacking note: this PR is based on `otto/and-414-managed-broker` / #383 because the entitlement API reuses the managed-link broker summary and institution-count enforcement.

## Verification

```text
git diff --check
swift test --disable-sandbox --skip-update --filter 'ConsumerFoundationTests|SubscriptionPlanTests|LinkTokenRequestTests'
# Test run with 58 tests passed

swift build --target PlaidBarServer --skip-update -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors
# Build of target: 'PlaidBarServer' complete
```

## Privacy / security

- Stripe webhook handling stores normalized metadata only: event ID/type, status, plan, and period/trial dates.
- No Stripe secrets, webhook signatures, raw Stripe payloads, customer emails, card/payment details, invoices, Plaid tokens, raw Plaid payloads, account IDs, balances, transactions, database paths, logs, or screenshots are exposed to the app.
- BYO/demo/local mode remains ungated.

## Limitations

This is the tested local/server contract for billing entitlements. Production still needs real Stripe Checkout/Portal API calls, webhook signature verification, Stripe product/price IDs, and durable hosted-bridge webhook event storage.

Linear: AND-348

### DIFF
--- a/Sources/PlaidBarCore/Models/BillingSubscription.swift
+++ b/Sources/PlaidBarCore/Models/BillingSubscription.swift
@@ -93,6 +93,102 @@ public struct SaveBillingSubscriptionRequest: Codable, Sendable, Equatable {
     }
 }
 
+public struct BillingCheckoutSessionRequest: Codable, Sendable, Equatable {
+    public let plan: SubscriptionPlan
+    public let successURL: String
+    public let cancelURL: String
+
+    public init(plan: SubscriptionPlan, successURL: String, cancelURL: String) {
+        self.plan = plan
+        self.successURL = successURL
+        self.cancelURL = cancelURL
+    }
+}
+
+public struct BillingCheckoutSessionResponse: Codable, Sendable, Equatable {
+    public let checkoutURL: String
+    public let plan: SubscriptionPlan
+    public let mode: String
+
+    public init(checkoutURL: String, plan: SubscriptionPlan, mode: String = "subscription") {
+        self.checkoutURL = checkoutURL
+        self.plan = plan
+        self.mode = mode
+    }
+}
+
+public struct BillingPortalSessionRequest: Codable, Sendable, Equatable {
+    public let returnURL: String
+
+    public init(returnURL: String) {
+        self.returnURL = returnURL
+    }
+}
+
+public struct BillingPortalSessionResponse: Codable, Sendable, Equatable {
+    public let portalURL: String
+
+    public init(portalURL: String) {
+        self.portalURL = portalURL
+    }
+}
+
+/// Safe Stripe webhook projection. It intentionally carries only normalized
+/// subscription metadata; never raw Stripe payloads, signatures, customer email,
+/// payment-method details, invoices, or secrets.
+public struct StripeBillingWebhookEvent: Codable, Sendable, Equatable {
+    public let id: String
+    public let type: String
+    public let status: BillingSubscriptionStatus
+    public let plan: SubscriptionPlan
+    public let currentPeriodEnd: Date?
+    public let trialEndsAt: Date?
+
+    public init(
+        id: String,
+        type: String,
+        status: BillingSubscriptionStatus,
+        plan: SubscriptionPlan,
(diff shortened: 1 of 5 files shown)


ftchvs:
## Automated senior review — HOLD (gated; ships FAIL-OPEN, must be fail-closed before exposure)

Adversarial multi-agent review. **Verdict: hold — correctly gated (deferred billing cluster); changes nothing in the beta today** (route is under `/api` behind `APITokenMiddleware` on 127.0.0.1, and managed linking is blocked in default `.local` mode regardless of subscription). Secret/token/PII boundary is **clean** (billing/entitlement responses carry only plan/status/count/limit/trial/features).

**But the Stripe webhook ships fail-open and must be made fail-closed before it is ever exposed to real Stripe delivery.**

### Must fix before un-gating
- **P1 — Stripe webhook is fail-OPEN.** `handleStripeWebhook` (`BillingRoutes.swift:93-111`) applies any decoded event via `billingStore.save` with **zero signature/authenticity verification** — the exact asymmetry vs the Plaid path, whose `UnconfiguredPlaidWebhookSignatureValidator` *throws* until a real validator is wired. A test (`ConsumerFoundationTests.swift:562-616`) even codifies the fail-open grant as intended. Once exposed for production delivery (which requires an open route), a forged `POST {status:active, plan:plus}` is a **remote unauthenticated free-premium / institution-cap-raise** primitive. **Fix:** add a `StripeWebhookVerifier` protocol with a default impl that throws; HMAC-verify over the **raw** request body + `Stripe-Signature` header (the handler currently decodes a *normalized projection* with no raw-bytes/signature field, so real HMAC is structurally impossible — extend the wire contract); reject unsigned/invalid with 400 mutating nothing; invert the test to assert rejection.
- **P2 — Second unverified entitlement-write path:** `PUT /api/billing/subscription` (`BillingRoutes.swift:46-53`) writes status/plan straight to the premium-granting row. Lock down once the verified webhook is the source of truth.
- **P2 — No monotonicity guard:** a stale `active` event after `canceled` silently re-grants premium (no event timestamp/sequence in the projection). Add an ordering field + reject older-than-stored.
- **P2 — In-memory idempotency ledger:** `StripeBillingEventStore` is an in-memory `Set` (never injected from `App.swift`), so replay survives restart. Back it with Fluent/SQLite like the Plaid `WebhookEventStore`.
- **P2 — Tie signature-verification to route exposure** (refuse to start if mounted open without a verifier) and add webhook-driven lifecycle tests for `active`/`past_due`/`expired` (only `trialing`/`canceled` are covered).
- **P2/P3 — Checkout/portal redirect URLs** are reflected with only `.urlQueryAllowed` encoding → open-redirect/query-injection shape once wired to real Stripe. Build with `URLComponents` + an allowlist.

_Automated PR-review loop. Not merging — gated, pending the product/sequencing decision._

ftchvs:
✅ The fail-open Stripe webhook flagged above is fixed in #386: `StripeWebhookVerifier` with a fail-closed `Unconfigured…` default + a production-ready HMAC-over-raw-bytes `StripeSignatureWebhookVerifier`; the handler now verifies the raw body before applying. Remaining P2s (PUT lockdown, monotonicity, durable idempotency, redirect allowlist) stay as follow-ups. Hold for the gating decision.

PR #380, opened 2026-06-15
AND-409: Harden Hosted Link completion handling

ftchvs:
## Summary
- Add metadata-only Hosted Link completion store and `POST /webhooks/plaid/hosted-link` receiver.
- Let OAuth callback recover completion metadata and fetch public token results from Plaid Link session lookup when redirect lacks a public token.
- Require `PLAID_LINK_WEBHOOK_URL` in hosted-bridge config.

## Tests
- `git diff --check`
- `swift test --filter 'LinkTokenRequestTests|oauthCallbackUsesRedirectPublicTokenWithoutSessionLookup|hostedLinkWebhookSuccessAllowsSessionLookup|hostedLinkCompletionStoreIsIdempotentByAnyIdentifier|oauthCallbackDistinguishesUserExitExpiredAndRetryableProviderFailure|hostedBridgeRequiresWebhookURL' --skip-update --disable-keychain` — 21 tests passed

Linear: AND-409

PR #385, opened 2026-06-15
fix: review fixes for managed-link stack (AND-408/409/411)

ftchvs:
## Why

An adversarial multi-agent review of the `otto/and-408..411` managed-link stack (PRs #379–#382) surfaced confirmed defects that ship inside a green build. This PR stacks on **#382** and resolves the **fix-before-merge blockers** plus the one **live-in-beta** finding. It does not change merge readiness of the underlying stack — it makes it mergeable.

The full strict-concurrency build (3 targets, `-warnings-as-errors`) and the complete suite (**766 tests**) pass locally with these changes.

### 1. Blocker — webhook status mis-mapping / dead mapper (AND-411)
The live receiver mapped status via the coarse `statusSignal`, which sent `PENDING_EXPIRATION → .connected`, silently dropped `PENDING_DISCONNECT` / `USER_PERMISSION_REVOKED` / `NEW_ACCOUNTS_AVAILABLE`, and clobbered a hard `.error` on `LOGIN_REPAIRED` — while the correct `ItemStatusMapping` had **zero production callers** and only the dead path was tested. `apply()` now routes through `ItemStatusMapping` against the item's *current* status, and the redundant `WebhookItemStatusSignal`/`statusSignal` is removed (single source of truth). Preserved `ERROR → loginRequired`; mapped `LOGIN_REPAIRED_WITH_NEW_ACCOUNTS → newAccountsAvailable`.

### 2. Blocker — Hosted Link completion-webhook poisoning (AND-409)
The unauthenticated `/webhooks/plaid/hosted-link` receiver writes into the first-writer-wins completion store the OAuth callback reads to gate token exchange. Keyed on the non-secret `link_token`/`link_session_id`, a forged non-success record could veto a victim's genuine link. `publicTokenResults` now fetches Plaid's authoritative link-session result first and prefers a real success; it consults completion metadata only when bound to the unguessable single-use `state`.

### 3. Live-in-beta — WeeklyReview false-nag (AND-411)
`WeeklyReview` counted unhealthy items via `status != .connected`, so a freshly repaired `.loginRepaired` item produced a false "needs a check" nag. Now filters on the shared `.isDegraded` predicate.

### 4. Defense-in-depth — managed-redirect TLS gate (AND-408)
The managed-mode HTTPS requirement keyed on `deployment == .hostedBridge`, so managed + local deployment + production accepted a plaintext `http://` redirect carrying the one-time `state`. Now gates on `production` regardless of deployment (mirrors the `.app` branch).

## Tests added
- Receiver-level: each of the four consent codes persists its modeled `ItemConnectionStatus`; `LOGIN_REPAIRED` does not clobber `.error`.
- Adversarial: a forged `user_exit` keyed by `link_token` cannot veto a real Plaid success.
- Predicate matrix: `isDegraded` / `needsUpdateMode` for all 8 statuses.
- Config: managed + local + production + `http://` now throws.

## Deferred (documented follow-ups, not blockers)
- `HostedLinkCompletionStore` TTL/size-cap/eviction + O(n²) per-insert index rebuild.
- `needsPollingSync` staleness (latest-SYNC-bearing event; compare vs sync-cursor commit time, not `item.updatedAt`).
- Verifier adversarial regression tests (`alg:"none"`, 2-/4-segment tokens, future `iat`, undecodable segment) + missing-header 401 + a doc/tracking marker on `UnconfiguredPlaidWebhookSignatureValidator`.

## Provenance
Generated by an automated PR-review loop. **Hold for human merge decision** — security-sensitive OAuth/webhook code; the underlying stack has no CI gate and no prior human review.

### DIFF
--- a/Sources/PlaidBarCore/Models/WeeklyReview.swift
+++ b/Sources/PlaidBarCore/Models/WeeklyReview.swift
@@ -364,7 +364,10 @@ public enum WeeklyReviewBuilder {
             ))
         }
 
-        let unhealthyItems = itemStatuses.filter { $0.status != .connected }.count
+        // `.loginRepaired` is a healthy, non-degraded state; counting it via
+        // `!= .connected` produced a false "needs a check" nag right after a
+        // successful repair. Use the shared degraded predicate instead.
+        let unhealthyItems = itemStatuses.filter { $0.status.isDegraded }.count
         if unhealthyItems > 0 || isSyncStale {
             // A login-required item needs the reconnect/update-link flow, not
             // another refresh — treat it as reconnectable alongside errors so
--- a/Sources/PlaidBarServer/Config/ServerConfig.swift
+++ b/Sources/PlaidBarServer/Config/ServerConfig.swift
@@ -962,7 +962,12 @@ struct OAuthRedirectConfiguration: Sendable, Equatable {
             guard let configuredURI else {
                 throw ServerConfigError.missingManagedRedirectURI
             }
-            if deployment == .hostedBridge, plaidEnvironment == .production, !isHTTPS(configuredURI) {
+            // Production OAuth redirect URIs must be HTTPS regardless of
+            // deployment mode (mirrors the `.app` branch): a plaintext callback
+            // would carry the one-time `state` over http. Gating on `.hostedBridge`
+            // alone left managed + local deployment + production accepting an
+            // http:// redirect — a TLS-downgrade hole.
+            if plaidEnvironment == .production, !isHTTPS(configuredURI) {
                 throw ServerConfigError.invalidManagedRedirectURI
             }
             return OAuthRedirectConfiguration(mode: .managed, uri: configuredURI)
--- a/Sources/PlaidBarServer/Routes/ItemStatusMapping.swift
+++ b/Sources/PlaidBarServer/Routes/ItemStatusMapping.swift
@@ -14,6 +14,19 @@ enum ItemStatusMapping {
         switch normalized {
         case "LOGIN_REPAIRED":
             return currentStatus.applyingLoginRepaired()
+        case "LOGIN_REPAIRED_WITH_NEW_ACCOUNTS":
+            // Login was repaired AND Plaid has newly available accounts to grant;
+            // surface the actionable new-accounts state rather than silently
+            // clearing the prompt. Like the sibling `LOGIN_REPAIRED` branch (and
+            // `applyingLoginRepaired()`), a repair signal must never clobber a hard
+            // `.error` this receiver cannot resolve — the item still needs a full
+            // reconnect, so preserve `.error` instead of downgrading it.
+            return currentStatus == .error ? .error : .newAccountsAvailable
+        case "ERROR":
+            // A bare ITEM `ERROR` webhook whose granular `error.error_code` this
+            // receiver does not yet parse: keep the item degraded (login needed)
+            // rather than dropping the signal, preserving prior behavior.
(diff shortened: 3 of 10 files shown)


ftchvs:
## PR audit — managed-link stack (held for human merge decision)

Audited all 8 open PRs as one stack. Re-verified locally (strict-concurrency build, `-warnings-as-errors`, + full suite), not relying on CI — the `claude` workflow only ever **skips** (Actions billing gate), so there is no automated build/test gate on any of these.

### The stack
```
main
└─ #379 AND-408 redirect readiness gates
   └─ #380 AND-409 Hosted Link completion
      └─ #381 AND-410 webhook receiver
         └─ #382 AND-411 consent-repair states
            ├─ #385 review fixes (408/409/411)   ← this PR
            └─ #383 AND-414 broker (gated)
               └─ #384 AND-348 Stripe (gated)
                  └─ #386 un-gating security fixes (414/348)
```
All 8 are `MERGEABLE/CLEAN`; main is an ancestor of every branch (no rebase needed). No branch protection on `main`; all merge methods allowed.

### Verified green
- **#385 stack** (408–411 + these fixes): strict build OK, **766/766 tests pass** — matches the PR claim. Closes the two confirmed blockers (webhook status mis-mapping `PENDING_EXPIRATION→connected` / dead `ItemStatusMapping`; Hosted Link completion-store poisoning) + the managed-redirect TLS gate + the WeeklyReview false-nag.
- **#386 stack** (full tree incl. 414/348 + broker/billing fixes): strict build OK, **773/773 tests pass** — matches the PR claim. Closes the two must-fix-before-un-gating P1s (managed institution-limit **TOCTOU race** → now atomic check+insert under a FIFO lock; **Stripe webhook fail-open** → now fail-closed `UnconfiguredStripeWebhookVerifier` + HMAC verifier).

### Merge-sequencing finding
**#385 and #386 are siblings** (both branch off the 408–411 base) and both touch `LinkRoutes.swift`, but **a trial merge auto-resolves with no conflicts.** So the whole stack integrates cleanly.

### Still-open findings (not blockers, documented in the PRs)
- `HostedLinkCompletionStore` has no TTL/size cap → unbounded growth (#380 P2).
- `PUT /api/billing/subscription` remains a second unverified entitlement-write path (local-auth-only today) — #383's "trusted billing state" P1 is only **partially** closed by #386 (webhook side fixed; local write route still authoritative).
- Billing apply has no monotonicity guard (stale `active` after `canceled`) — Stripe event-ordering P2 (#384) still open.
- In-memory Stripe idempotency ledger (not durable) — #384 P2 still open.
- Checkout/portal redirect-URL `.urlQueryAllowed` under-encoding — #384 P2 still open.
- Managed items still use direct local Plaid sync, not fail-closed on cancellation — #383 P1 data-plane finding still open.

### Recommended merge plan (squash, bottom-up) — for your approval
1. `#379 → #380 → #381 → #382` (squash each into the next/main as the base advances), **then**
2. `#385` (carries the 408/409/411 fixes) — **do not merge 379–382 without 385**; 385 fixes a P1 that ships live otherwise.
3. Gated tier — only if you choose to land the broker/Stripe scaffolding now: `#383 → #384 → #386`. #386 must land with them (it's what makes them safe to un-gate later). Both features stay config-gated/inert in the local-first beta regardless.

Holding here — security-sensitive OAuth/webhook/billing code, no CI gate, no prior human review. Recommend: approve+squash 1–2 now (real bug fixes, fully green); decide on the gated 3 separately.

PR #386, opened 2026-06-15
fix: close managed-broker + Stripe un-gating security holes (AND-414/AND-348)

ftchvs:
## Why

The adversarial review of the gated broker (#383) + Stripe (#384) stack flagged two **must-fix-before-un-gating** P1s. This PR stacks on **#384** and fixes both, with tests. It does not change the gating decision — both features remain config-gated and inert in the local-first beta; this makes them safe to un-gate later.

Full strict-concurrency build (3 targets, `-warnings-as-errors`) + the complete suite (**773 tests**) pass locally.

## Fix 1 — managed institution-limit TOCTOU race (#383 / AND-414)

The cap was a non-atomic check-then-insert: the OAuth callback read `activeInstitutionCount` via the entitlement summary, then inserted via a *separate* `TokenStore.saveItem` actor call. Because Fluent calls suspend, **actor reentrancy** let two concurrent managed completions at `count == limit-1` both observe room and both insert (overshoot by N-1).

- New `TokenStore.saveManagedItemEnforcingLimit(...)` runs the count check + insert under an **in-process FIFO lock held across both `await` points** (single local server), throwing `.limitReached` at capacity.
- The OAuth callback keeps the entitlement pre-check (so a degraded subscription never spends a token) but now enforces the cap **atomically at insert**; a race-loser becomes spent-but-not-stored ("re-link").
- Regression test: seed `limit-1`, fire `limit` concurrent inserts → exactly one wins, count never exceeds the limit.

## Fix 2 — Stripe webhook fail-open (#384 / AND-348)

`handleStripeWebhook` applied any decoded event to billing state with **zero signature verification**. Once exposed for real Stripe delivery, a forged `{status:active, plan:plus}` POST would be a remote unauthenticated **free-premium / cap-raise**.

- New `StripeWebhookVerifier` protocol; default `UnconfiguredStripeWebhookVerifier` **throws** until a real validator is wired (mirrors the Plaid `Unconfigured…SignatureValidator` fail-closed default; this is the production default).
- Production-ready `StripeSignatureWebhookVerifier`: HMAC-SHA256 of `"{t}.{rawBody}"` from the `Stripe-Signature` header over the **exact received bytes**, with timestamp tolerance + constant-time compare.
- `handleStripeWebhook` now collects the **raw body → verifies → then decodes/applies** (the wire-contract change). An unsigned/forged event mutates nothing.
- Tests: fail-closed default rejects + grants nothing; verifier accepts valid / rejects wrong-sig, missing header, stale timestamp; the two lifecycle webhook tests inject an accepting verifier (they test application logic).

## Still deferred (documented follow-ups, NOT in scope here)
- `PUT /api/billing/subscription` second unverified entitlement-write path (local-auth-only today).
- Monotonicity/ordering guard on billing apply (stale `active` after `canceled`).
- Durable Stripe idempotency ledger (currently in-memory).
- Checkout/portal redirect-URL allowlist; tie signature-verification to route exposure.

## Provenance
Automated PR-review loop, at the user's direction to fix the two un-gating blockers. **Hold for human merge** — security-sensitive billing/broker code; the underlying stack has no CI gate.

### DIFF
--- a/Sources/PlaidBarServer/Routes/BillingRoutes.swift
+++ b/Sources/PlaidBarServer/Routes/BillingRoutes.swift
@@ -1,5 +1,6 @@
 import Foundation
 import Hummingbird
+import HTTPTypes
 import NIOCore
 import PlaidBarCore
 
@@ -14,17 +15,23 @@ struct BillingRoutes: Sendable {
     let tokenStore: TokenStore?
     let deployment: DeploymentMode
     let webhookEvents: StripeBillingEventStore
+    let verifier: any StripeWebhookVerifier
+    let now: @Sendable () -> Date
 
     init(
         billingStore: BillingSubscriptionStore,
         tokenStore: TokenStore? = nil,
         deployment: DeploymentMode = .local,
-        webhookEvents: StripeBillingEventStore = StripeBillingEventStore()
+        webhookEvents: StripeBillingEventStore = StripeBillingEventStore(),
+        verifier: any StripeWebhookVerifier = UnconfiguredStripeWebhookVerifier(),
+        now: @escaping @Sendable () -> Date = { Date() }
     ) {
         self.billingStore = billingStore
         self.tokenStore = tokenStore
         self.deployment = deployment
         self.webhookEvents = webhookEvents
+        self.verifier = verifier
+        self.now = now
     }
 
     func register(with group: RouterGroup<some RequestContext>) {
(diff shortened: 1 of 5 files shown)


ftchvs:
## Audit verified — gated-tier un-gating fixes

Re-verified locally: full stack (408→411 + 414 + 348 + these fixes) builds under strict-concurrency `-warnings-as-errors` and **773/773 tests pass** — matches the PR claim. Both un-gating P1s are closed:
- Managed institution-limit **TOCTOU race** → atomic count+insert under FIFO lock.
- **Stripe webhook fail-open** → fail-closed `UnconfiguredStripeWebhookVerifier` default + HMAC-SHA256 verifier over raw bytes.

**Sequencing:** this branches off #384 (the gated stack) and is a sibling of #385; both touch `LinkRoutes.swift` but trial-merge auto-resolves cleanly. Land order if un-gating: `#383 → #384 → #386`, and #386 must land with them.

**Still open (per PR's own "deferred"):** `PUT /api/billing/subscription` unverified local entitlement-write, billing monotonicity/ordering guard, durable idempotency ledger, redirect-URL allowlist, managed data-plane not fail-closed on cancellation (#383 P1). Full consolidated audit + merge plan in #385.

**Hold for human merge** — both features stay config-gated/inert in the local-first beta; merging is a product/security call, not auto-dispatched.

ftchvs:
## Deferred un-gating follow-ups (tracked, NOT beta blockers)

The #379–386 managed-link stack merged after two adversarial review passes (0 must-fix-before-merge; every item below is provably inert in the default `.local`/BYO/demo beta and only matters once someone deliberately enables `hosted-bridge` + production). Recording here so they are not lost before any un-gating:

**Hosted-bridge / managed (gate before enabling `.hostedBridge`+production):**
- `LinkRoutes`: legacy `POST /api/link/create` still mints a Hosted Link session and stores items as BYO, bypassing the managed institution cap. (P1)
- `AccountRoutes`/`TransactionRoutes`: managed items would use direct Plaid sync with local creds rather than the blind data-plane proxy. (P1)
- `ManagedLinkEntitlementService`: locally-writable `BillingSubscription` (`PUT /api/billing/subscription`) is authoritative for `canCreateManagedLink` — needs signed entitlement before trust. (P1)
- `WebhookRoutes`: wire a real ES256/JWK Plaid signature validator (today `Unconfigured…` is fail-closed, so the path is inert). (P2)
- Hosted Link completion: redirect `status=SUCCESS` short-circuit fragility; EVENTS/ITEM_ADD_RESULT first-writer poisoning; completion-store TTL/eviction; `EXITED` final-status handling. (P1/P2)

**Billing (before exposing the Stripe webhook for real delivery):**
- Durable Stripe idempotency ledger (in-memory today); mark-processed-after-save; monotonicity/ordering guard (older event overwriting newer); `PUT /api/billing/subscription` lockdown; checkout/portal redirect-URL allowlist + `.urlQueryAllowed` escaping fix. (P2)

**OAuth redirect (`#379`):** `.app` mode has no exchange handler; managed requires a redirect URI unconditionally (affects sandbox hosted-bridge boot); `https://localhost` is accepted as production-ready. (P2)

**Test nicety:** the new `.local`-blocks-managed-create invariant test could additionally vary count/status axes. (P3)

PR #400, opened 2026-06-16
fix: unblock live weekly review state

### LINKED ISSUE #392: Weekly Review card is a permanent "Waiting for AND-399" dead-end outside demo mode
Mirror of Linear [AND-443](https://linear.app/andeslab/issue/AND-443). Severity P2.

**Symptom:** On the live sandbox/production dashboard the **Weekly Review** card is permanently frozen on "Weekly review unlocks after the transaction review inbox can say which transactions are trusted." plus a warning chip **"Waiting for AND-399 review state"** — leaking an internal ticket id into user-facing copy on the primary surface. No action flips it to the real checklist.

**Root cause (HEAD 812f7cf):**
- `Sources/PlaidBar/App/AppState.swift:759` — `weeklyReviewTransactionState`: `guard isDemoMode else { return nil }` (only `--demo` yields non-nil)
- `Sources/PlaidBarCore/Models/WeeklyReview.swift:233` — `guard let transactionState else { return .waitingForTransactionReview }`
- `WeeklyReview.swift:204-212` — that presentation sets `isBlockedByTransactionReviewDependency: true`
- `Sources/PlaidBar/Views/WeeklyReviewCard.swift:17` renders the dead-end; `MainPopover.swift:436` mounts the card unconditionally
- The dependency (the transaction review inbox) is Done and live right below it (`MainPopover.swift:439`), with real metadata on disk (46 needsReview + 2 reviewed)

**Proposed fix:** Derive a real `WeeklyReviewTransactionState` from `transactionReviewMetadata` (status `.reviewed`/`.ignored` ⇒ trusted) instead of returning nil outside demo, or hide the card outside demo. Either way, remove the user-facing "AND-399" string.
\n\n## Additional implementation details\n\n### Fix objective\n- Remove external dependency token text from UI and make Weekly Review state a function of actual review metadata.\n\n### Concrete changes\n1. In `Sources/PlaidBar/App/AppState.swift`, compute `weeklyReviewTransactionState` using real `transactionReviewMetadata` completion fields (`reviewed` / `ignored` / `needsReview`) instead of hardcoding `nil` outside demo mode.\n2. In `Sources/PlaidBarCore/Models/WeeklyReview.swift`, if dependent data is missing, return a neutral state (e.g., `notReady`) and avoid exposing internal AND issue IDs in copy.\n3. In `Sources/PlaidBar/Views/WeeklyReviewCard.swift`, replace the static dead-end text and blocker chip with a real state label and action.\n4. Remove/replace any user-facing "Waiting for AND-*" strings across this path.\n\n### Validation\n- Add a unit test for non-demo mode with `transactionReviewMetadata` containing all reviewed/ignored values and assert card no longer blocks.\n- Add a regression test for the missing-metadata scenario so no hard-coded internal ticket string leaks.\n

## Documentation-backed implementation guidance
- Derive Weekly Review state from persisted review completion signals and remove internal ticket-number copy from user-facing text.
- If the source data is missing, degrade gracefully with neutral UX rather than hard-coded dependency IDs.
- Recommended source for transaction sync state design patterns: [Plaid Transactions API sync pagination behavior](https://plaid.com/docs/api/products/transactions/)
- Validation: assert UI has no string containing `AND-` once user-facing and transitions when `transactionReviewMetadata` is ready.

ftchvs:
## Summary
- Derive live Weekly Review transaction state from transaction-review metadata instead of blocking outside demo mode.
- Treat reviewed/ignored metadata as trusted and needs-review/missing metadata as actionable review work.
- Replace internal `AND-399` blocker copy with neutral not-ready messaging when transaction data is unavailable.

Fixes #392
Linear: AND-443

## Validation
- `git diff --check`
- `grep -R "AND-" -n Sources/PlaidBarCore/Models/WeeklyReview.swift Sources/PlaidBar/Views/WeeklyReviewCard.swift` returns no matches
- `swift test --filter WeeklyReviewTests --scratch-path /tmp/plaidbar-issue392-build`

## Safety
- No Plaid credentials, provider tokens, raw payloads, account IDs, transaction IDs, balances, SQLite data, or logs added.
- User-facing weekly-review copy avoids internal ticket IDs and remains privacy-preserving.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -752,16 +752,21 @@ final class AppState {
     }
 
     private var weeklyReviewTransactionState: WeeklyReviewTransactionState? {
-        // AND-403 is intentionally gated on AND-399. Production must provide
-        // trusted/unreviewed transaction state before this returns a value; raw
-        // Plaid transactions alone are not treated as reviewed. Demo mode uses
-        // synthetic ids so the checklist surface remains locally exercisable.
-        guard isDemoMode else { return nil }
-        let unreviewed = Set(transactions.filter(\.pending).map(\.id))
-        let trusted = Set(transactions.map(\.id)).subtracting(unreviewed)
-        return WeeklyReviewTransactionState(
-            trustedTransactionIds: trusted,
-            unreviewedTransactionIds: unreviewed
+        // AND-403 is intentionally gated on AND-399 review metadata. Raw Plaid
+        // transactions alone are not treated as reviewed; demo mode keeps using
+        // pending flags so the checklist surface remains locally exercisable.
+        guard !isDemoMode else {
+            let unreviewed = Set(transactions.filter(\.pending).map(\.id))
+            let trusted = Set(transactions.map(\.id)).subtracting(unreviewed)
+            return WeeklyReviewTransactionState(
+                trustedTransactionIds: trusted,
+                unreviewedTransactionIds: unreviewed
+            )
+        }
+
+        return WeeklyReviewTransactionState.derived(
+            from: transactions,
+            metadata: transactionReviewMetadata
         )
     }
 
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -48,6 +48,7 @@ final class AppState {
     }
     var transactionReviewMetadata: [TransactionReviewMetadata] = []
     var transactionRules: [TransactionRule] = []
+    var hasLoadedTransactionReviewStorage = false
     var isLoading = false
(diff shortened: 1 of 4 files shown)


ftchvs:
Addressed Codex feedback. Commit: d4fa017\n\nFixes:\n- Weekly Review now derives transaction readiness from the actionable TransactionReviewInbox snapshot.\n- Empty live transaction data remains in the neutral not-ready state instead of being treated as reviewed.\n\nValidation:\n- git diff --cached --check\n- swift test --filter 'WeeklyReviewTests|TransactionReviewInboxTests' --scratch-path /tmp/plaidbar-pr400-review-fixes\n- Result: 21 focused tests passed

PR #434, opened 2026-06-17
fix: reconcile posted↔pending review state and serialize storage writes (AND-399)

ftchvs:
## What

Forward-ports **AND-399** (two P2 follow-ups from PR #370, transaction review inbox) onto current `main`. The work existed only on the stale `worktree-fix+and-399-pending-reconciliation` branch (authored 2026-06-14); it was never merged and is **not** present on `main` today. Rebased/cherry-picked onto `234f261` with conflicts resolved.

### 1. Posted↔pending reconciliation
When Plaid posts a pending charge it removes the pending transaction and adds a new posted one linked by `pending_transaction_id`. The evaluator's metadata lookup under the *posted* id missed the `lastSeenPending`/amount/name recorded under the old *pending* id — so the "settled with a different amount or name" review never fired for the common posting flow.

- `TransactionDTO` and `PlaidTransaction` now carry `pendingTransactionId` (optional, privacy-safe, kept out of UI/logs, decoded from Plaid via `convertFromSnakeCase`).
- The evaluator falls back to the prior pending record as the change baseline; a charge whose pending phase was already approved/ignored reopens (`status → needsReview`) when it settles materially differently.

### 2. Serialize review-storage writes
`persistReviewStorage()` spawned an unstructured `Task` per action, so rapid approve/undo/rule actions could land out of order and overwrite the JSON with stale state. Writes now go through a serial `ReviewStorageWriter` (`AsyncStream`, `bufferingNewest(1)`) that applies full snapshots in enqueue order and coalesces bursts to the latest — last logical state always wins on disk. `MainActor`-safe and strict-concurrency clean.

## Forward-port note
The original branch declared `PlaidTransaction.pendingTransactionId` as `let` with no default, which made it a **required** memberwise-init parameter. A newer `main` test (`PlaidBarServerTests.swift:1107`, added after the branch was cut) constructs `PlaidTransaction` without it. Changed to `var pendingTransactionId: String? = nil` so the field (a) stays in `Decodable` synthesis — a `let … = nil` would be silently excluded from decoding and break reconciliation — and (b) is optional in the synthesized memberwise init. This matches how `TransactionDTO` already defaults the parameter.

## Verification
- `swift test` — **827 tests pass** (incl. the 5 new posted↔pending reconciliation tests in `TransactionReviewInboxTests`)
- `git diff --check` clean; changed-line secret scan clean (the new id field is privacy-safe and kept out of UI/logs)

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -291,6 +291,7 @@ final class AppState {
     private let serverClient = ServerClient()
     private let localDataCache = LocalDataCacheService()
     private let appLockService = AppLockService()
+    private let reviewStorageWriter = ReviewStorageWriter()
     private var localAIInsightsService = LocalAIInsightsService()
     private let notificationService: any NotificationServiceProtocol
     private var refreshTask: Task<Void, Never>?
@@ -2953,16 +2954,22 @@ final class AppState {
         let cacheDirectory = activeStorageDirectoryURL
         let cacheContext = transactionCacheContext
         let cache = localDataCache
-        Task {
+        // Hand the full snapshot to the serial writer instead of spawning an
+        // independent Task. Independent tasks could complete out of order, letting
+        // a stale write overwrite a newer one; the writer applies snapshots in
+        // enqueue order so the last logical state always wins on disk.
+        reviewStorageWriter.enqueue { [weak self] in
             do {
                 try await cache.saveTransactionReviewMetadata(metadata, to: cacheDirectory, context: cacheContext)
                 try await cache.saveTransactionRules(rules, to: cacheDirectory, context: cacheContext)
             } catch {
-                await MainActor.run {
-                    self.error = "Review inbox storage failed to save: \(error.localizedDescription)"
-                }
+                await self?.reportReviewStorageFailure(error.localizedDescription)
             }
         }
+    }
+
+    private func reportReviewStorageFailure(_ description: String) {
+        error = "Review inbox storage failed to save: \(description)"
     }
 
 }
--- a/Sources/PlaidBar/Services/ReviewStorageWriter.swift
+++ b/Sources/PlaidBar/Services/ReviewStorageWriter.swift
@@ -1,0 +1,46 @@
+import Foundation
+
+/// Serializes transaction-review storage writes so the most recent in-memory
+/// state always wins on disk.
+///
+/// Every review action (approve / ignore / rule edit / undo) hands the writer a
+/// closure that persists a *complete* snapshot of the review metadata and rules.
+/// A single consumer applies those closures strictly in the order they were
+/// enqueued, so a slow earlier write can never land after a newer one and
+/// resurrect stale JSON. Because each write carries the full state, rapid bursts
+/// are coalesced to the newest snapshot — intermediate ones are dropped rather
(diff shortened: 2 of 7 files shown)


ftchvs:
## Addressing the Codex review (commit `7fd30b0`)

### ✅ P2 — "Carry prior pending metadata onto posted transactions" (`TransactionReview.swift:192`)
Fixed. You're right that `AppState.seedReviewMetadataForNewTransactions` plants a fresh `.needsReview` record under the posted id *before* the evaluator runs, so `ownMetadata` almost always exists and the old `ownMetadata ?? priorPendingMetadata` always picked the seed — dropping the status/overrides of a charge the user already reviewed while pending. The evaluator now treats the posted charge as resolved-on-its-own **only once the user has acted on it directly** (`reviewed`/`ignored` under the posted id); until then it prefers the prior pending record:

```swift
let ownResolved = ownMetadata.map { $0.status == .reviewed || $0.status == .ignored } ?? false
let metadata = ownResolved ? ownMetadata : (priorPendingMetadata ?? ownMetadata)
```

Regression test: `postedUnchangedWithSeededOwnRecordStaysReviewed` (seeded own record + reviewed pending + unchanged settle ⇒ stays out of the inbox, overrides preserved).

### ✅ P2 — "Stop reusing the pending baseline after review" (`TransactionReview.swift:361`)
Fixed by the same `ownResolved` gate on the baseline:

```swift
let pendingBaseline = ownResolved ? nil : pendingBaseline(own: ownMetadata, prior: priorPendingMetadata)
```

Once the posted charge is resolved under its own id we stop comparing against the old pending amount, so `.pendingChanged` no longer re-fires and reopens it on every refresh. Regression test: `resolvedPostedDoesNotReopenFromPendingBaseline` (reviewed posted record + stale differing pending record ⇒ does not reopen).

Both fail on the previous logic and pass now; full suite green under `-strict-concurrency=complete -warnings-as-errors` (829 tests).

### ⚠️ P1 — "Keep pending transaction ids out of the app DTO" (`TransactionDTO.swift:17`)
Want your call on this one — there's a real tension with the two P2 fixes above. The metadata store and the review evaluator both live app-side (`PlaidBarCore`/`AppState`), so reconciling posted→pending requires the app to look up the prior record **by its transaction id**. The DTO already carries the raw Plaid `transactionId` (as `id`) and `accountId`; `pendingTransactionId` is the same sensitivity class, and it's confined to the data/logic layers — it is **not** referenced by any `View` and **not** in any log statement (verified).

Fully honoring "non-provider local mapping / keep it in `PlaidBarServer`" means moving reconciliation to a sync-time metadata migration (carry the pending record onto the posted id at ingest, drop the raw id from the persisted transaction cache). That's a larger, separable change to money-trust logic that I'd rather not fold into this PR silently.

**Two options — happy to take either:**
1. Land this PR as-is (correctness fixes in, `pendingTransactionId` confined to local storage), and I open a follow-up to migrate reconciliation off the persisted raw id.
2. I do the migration here before merge.

Which do you prefer?

ftchvs:
Merged. The two P2 correctness comments are fixed in this PR (commit `7fd30b0`). The P1 (raw `pending_transaction_id` in the persisted cache) is tracked as a follow-up: #435 — move reconciliation to a sync-time metadata migration so the raw id leaves the cache.

PR #452, opened 2026-06-17
fix: keep pending reconciliation out of transaction cache

### LINKED ISSUE #435: Move posted↔pending reconciliation off the persisted raw pending_transaction_id (Codex P1 from #434)
## Context

Follow-up to the Codex **P1** review comment on #434 (merged as `1d09fcc`). That PR added posted↔pending review reconciliation. To link a posted charge back to its pending phase, `TransactionDTO` now carries Plaid's raw `pending_transaction_id` (`TransactionDTO.pendingTransactionId`, set in `TransactionRoutes.toDTO`), and the SwiftUI app **persists `TransactionDTO`s to its local cache** (`AppState` → `LocalDataCacheService.saveTransactions`). This puts a *second* raw Plaid transaction identifier into the app/cache.

It is currently confined to local storage — never referenced by any `View`, never logged — and is the same sensitivity class as the `transactionId` (`TransactionDTO.id`) and `accountId` the DTO already persists. But per the repo's local-first privacy posture, raw provider ids in the app cache should be minimized.

## Why it wasn't fixed in #434

Reconciliation currently lives in the **pure app-side evaluator** (`TransactionReviewInbox.evaluate` in `PlaidBarCore`), which needs `pendingTransactionId` at evaluation time — including on app launch when transactions are re-loaded from cache. So removing the id from the cache requires moving *where* reconciliation happens, which is a larger change to money-trust logic and was kept out of the correctness-focused PR.

## Proposed approach

Move reconciliation from the evaluator to a **sync-time metadata migration** in `AppState` (where review metadata is owned and persisted):

1. In `seedReviewMetadataForNewTransactions` (or a dedicated reconcile step), when a newly-synced posted transaction has a `pendingTransactionId` matching an existing metadata record:
   - Migrate that prior pending record onto the posted id (carry `status`, `userCategory`, `userMerchantName`, `isTransferOverride`, `reviewedAt`), set `lastSeen*` to the posted values, and **reopen** (`status = .needsReview` + a durable reason marker) iff it settled materially differently after having been `reviewed`/`ignored`.
   - Remove the consumed pending record.
2. Drop `pendingTransactionId` from the **persisted** transaction cache (strip before `saveTransactions`, or make it transient on the cache path) — it's only needed transiently during the live-sync reconciliation pass.
3. Simplify the evaluator: it no longer needs `pendingTransactionId`; it reads the already-migrated `ownMetadata`. Surface `.pendingChanged` from a durable marker on the metadata rather than recomputing from the raw link.
4. Re-home the existing evaluator-level reconciliation tests in `TransactionReviewInboxTests` to `AppState`-level (sync/migration) tests.

## Acceptance criteria

- [ ] Raw `pending_transaction_id` is no longer written to the persisted transaction cache.
- [ ] Posted-unchanged-after-review stays settled across an app restart (durable, not just within a session).
- [ ] Posted-changed-after-review reopens once and does not re-reopen after the user resolves it.
- [ ] `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` + `swift test` green.

## References
- PR #434 (merged `1d09fcc`); Codex P1 comment on `TransactionDTO.swift:17`
- Files: `Sources/PlaidBarCore/Models/TransactionReview.swift`, `Sources/PlaidBarCore/Models/TransactionDTO.swift`, `Sources/PlaidBar/App/AppState.swift` (`seedReviewMetadataForNewTransactions`), `Sources/PlaidBarServer/Routes/TransactionRoutes.swift`, `Sources/PlaidBar/Services/LocalDataCacheService.swift`

ftchvs:
## Summary

- moves posted↔pending transaction review reconciliation from evaluator-time fallback to sync-time metadata migration
- strips Plaid `pending_transaction_id` from the persisted transaction cache while preserving transient server DTO support
- persists changed-posted-charge reopen state via `TransactionReviewMetadata.reviewReasonCodes` so it survives restart and clears after user resolution

Fixes #435
Linear: AND-476
Kanban: `t_98d1bd91`

## Verification

- `git diff --check`
- `swift test --filter TransactionReviewInboxTests` — 20 tests passed
- `swift test --filter transactionCacheDoesNotPersistPendingTransactionIdentifiers` — 1 test passed
- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — passed
- `swift test --filter PlaidBarCoreTests` — 641 tests passed

## Privacy/security

No Plaid credentials, access tokens, public tokens, account IDs, transaction IDs, balances, raw payloads, prompts, or response bodies were printed or added to docs/tests. The app remains local-first; this change removes a raw Plaid pending transaction identifier from the persisted app transaction cache.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -2823,6 +2823,7 @@ final class AppState {
         updateReviewMetadata(id: id) { metadata, transaction in
             metadata.status = .ignored
             metadata.reviewedAt = Date()
+            metadata.reviewReasonCodes = []
             metadata.lastSeenAmount = transaction.amount
             metadata.lastSeenName = transaction.name
             metadata.lastSeenPending = transaction.pending
@@ -2936,8 +2937,14 @@ final class AppState {
             )
             changed = true
         }
+        let seeded = byId.values.sorted { $0.id < $1.id }
+        let reconciled = TransactionReviewMetadataReconciler.reconcilePostedPendingTransactions(
+            transactions: transactions,
+            metadata: seeded
+        )
+        if reconciled != seeded { changed = true }
         guard changed else { return }
-        transactionReviewMetadata = byId.values.sorted { $0.id < $1.id }
+        transactionReviewMetadata = reconciled
         persistReviewStorage()
     }
 
--- a/Sources/PlaidBarCore/Models/TransactionReview.swift
+++ b/Sources/PlaidBarCore/Models/TransactionReview.swift
@@ -182,28 +182,9 @@ public enum TransactionReviewInbox {
         let recurringByMerchant = Dictionary(grouping: recurring, by: { normalizedKey($0.merchantName) })
 
         let items = transactions.compactMap { transaction -> TransactionReviewItem? in
-            let ownMetadata = metadataById[transaction.id]
-            // When Plaid posts a previously-pending charge it arrives under a
-            // brand-new transaction id that links back to the pending id via
-            // pending_transaction_id. The pending-phase record (the user's
-            // category/transfer choices and the last-seen amount/name) still lives
-            // under that old id, so carry it forward to reconcile the two.
-            let priorPendingMetadata = transaction.pendingTransactionId.flatMap { metadataById[$0] }
-            // Production seeds a fresh `.needsReview` record under the posted id
-            // before this runs, so `ownMetadata` almost always exists. Treat the
-            // posted charge as resolved on its own only once the user has acted on
-            // it directly (reviewed/ignored under the posted id). Until then the
-            // own record is just the seeded baseline, so prefer the pending-phase
-            // record — otherwise the seeded `.needsReview` would mask a charge the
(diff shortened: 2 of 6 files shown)


ftchvs:
Closing without merging after adversarial review. Two problems:

1. **Over-scoped / risky.** Beyond the ~20-line cache-stripping fix its title describes, this PR rewrites `TransactionReview.swift` (deleting the live `priorPendingMetadata`/`ownResolved`/`pendingBaseline(own:prior:)` carry-forward logic that ships on current main) and adds a *destructive* sync-time `TransactionReviewMetadataReconciler`. It was authored off pre-#460 main and would conflict with / regress the posted↔pending review reconciliation main already ships.

2. **Even the minimal cache-strip is not safe alone.** `TransactionReview.swift:196` reads `transaction.pendingTransactionId` during review-inbox evaluation, and at cold start the transaction set is **cache-loaded**. Stripping the field from the persisted cache therefore breaks cold-start pending→posted review-metadata carry-forward until the first live sync re-populates it.

See #435 for the constraint + recommended approach.

Issue #437, opened 2026-06-17
[Audit 2026-06-16] Wire mandated release-diff secret scan into CI + release.sh (AND-461)


ftchvs:
**Tracked in Linear: AND-461** · Source: pre-release audit 2026-06-16 · **Do not implement yet.**

**Severity:** High · **Launch impact:** release-checklist hardening for a Plaid-handling finance app.

`docs/release-checklist.md:63` mandates a secret scan over the release diff, but nothing in `Scripts/release.sh` or `.github/workflows/ci.yml` runs one — only the opt-in local `Scripts/pre-push-gate.sh`. (A manual scan during the audit was clean; the automation is the gap.)

**Recommendation:** add a hard secret-scan step in CI and/or `release.sh` (invoke `pre-push-gate.sh` scan-only via `PLAIDBAR_GATE_SKIP_BUILD=1`, or add gitleaks).

ftchvs:
Fixed in #454 (squash-merged to main 2026-06-17). Linear: AND-461. Verified locally: strict-concurrency build + full test suite green (CI is billing-blocked).

PR #454, opened 2026-06-17
ci: run mandated secret scan in release.sh and CI (AND-461)

ftchvs:
## Summary
Wired the mandated release-diff secret scan into both the release script and CI by reusing the existing scanner in pre-push-gate.sh (no duplicated patterns). In Scripts/release.sh, added a hard scan step after the bash -n syntax check and before packaging: `PLAIDBAR_GATE_SKIP_BUILD=1 bash "$SCRIPT_DIR/pre-push-gate.sh"`. Because release.sh runs under `set -euo pipefail`, the scanner's non-zero exit (exit 2 on any finding) fails the release. In .github/workflows/ci.yml, added a "Secret scan" step after "Packaging checks" running the same scan-only invocation so it is correct for when CI returns. The scanner's no-stdin fallback derives an origin/main..HEAD (release-diff) range, matching the docs/release-checklist.md:63 requirement; PLAIDBAR_GATE_SKIP_BUILD=1 keeps it scan-only since release.sh and CI already run the strict build separately.

## For the reviewer
- CI scan range under shallow checkout: actions/checkout@v6 defaults to fetch-depth:1, so origin/main may be absent and pre-push-gate.sh falls back to scanning the single checked-out HEAD commit (git log -p HEAD) rather than the full origin/main..HEAD diff. This does not error (no fail-closed trip) and the task accepts CI being off, but if a full release-diff scan in CI is desired later, the checkout step should set fetch-depth: 0 (or fetch origin/main). No Swift test added: shell-only change per task scope (bash -n covers syntax; pre-push-gate.sh --selftest already exercises the scanner).

## Source
Pre-release audit 2026-06-16 — **AND-461** / #437. **Draft — do not merge yet** (part of the audit-fix stack; merge per the wave plan).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,9 @@ jobs:
         run: |
           bash -n Scripts/*.sh Scripts/vaultpeek-run Scripts/plaidbar-run
 
+      - name: Secret scan
+        run: PLAIDBAR_GATE_SKIP_BUILD=1 bash Scripts/pre-push-gate.sh
+
       - name: Version metadata alignment
         run: ./Scripts/verify-version-alignment.sh
 
--- a/Scripts/release.sh
+++ b/Scripts/release.sh
@@ -92,6 +92,9 @@ else
 fi
 bash -n Scripts/*.sh Scripts/vaultpeek-run Scripts/plaidbar-run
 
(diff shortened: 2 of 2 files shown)


ftchvs:

Issue #447, opened 2026-06-17
[Audit 2026-06-16] Server sync concurrency & atomicity: per-item in-flight gate + atomic cursor commit (AND-471)


ftchvs:
**Tracked in Linear: AND-471** · Source: pre-release audit 2026-06-16 · **Do not implement yet.**

**Severity:** Low. Correctness held by the idempotent id-keyed reducer; latent races.

- #24 No per-item in-flight gate on `GET /api/transactions/sync` (TransactionRoutes.swift:128,188-227) — overlapping syncs double Plaid calls.
- #25 `commitSyncCursors` non-atomic vs `deleteItem`: existence check (:75) + `saveSyncCursor` (:78) are separate awaits, can resurrect a `sync_cursors` row for a deleted item.

**Recommendation:** per-item async coalescing gate keyed by itemId; atomic existence-check+save in TokenStore; prune orphan cursors at startup.

ftchvs:
Fixed in #457 (squash-merged to main 2026-06-17). Linear: AND-471. Verified locally: strict-concurrency build + full test suite green (CI is billing-blocked).

PR #457, opened 2026-06-17
fix: per-item sync coalescing + atomic cursor commit (AND-471)

ftchvs:
## Summary
Fixed server sync concurrency and cursor-save atomicity. (1) Added a per-item async coalescing gate: a new generic actor ItemSyncCoalescer<Value: Sendable> keys in-flight sync Tasks by itemId, so concurrent GET /api/transactions/sync requests for the same item await the single in-flight result instead of racing the persisted cursor; distinct items still run concurrently. Wired it into TransactionRoutes.syncItems (extracted the per-item work into syncItemRecordingStatus). Added an explicit internal initializer to TransactionRoutes because the new private stored property would otherwise lower the synthesized memberwise init below internal and break the App.swift call site; the (plaidClient:tokenStore:maxConcurrentItemRefreshes:) signature used by existing tests is preserved. (2) Made the item-existence check + cursor save atomic: new TokenStore.saveSyncCursorIfItemExists(itemId:cursor:) runs the find+save under a new cursor-write FIFO lock (mirroring the existing managed-insert lock pattern), and deleteItem now removes the cursor+item rows under the same lock, so a cursor row can no longer be resurrected for a concurrently-deleted item. commitSyncCursors now calls the atomic method and returns the same 400 ('Cannot commit cursor for unknown item') when it no-ops, preserving the external contract. The reducer was not touched.

- Tests added/updated: PlaidBarServerTests.conditionalCursorSaveSkipsDeletedItem (atomic save no-ops + does not resurrect a sync_cursors row after the item is deleted; also verifies it persists while the item exists); PlaidBarServerTests.concurrentItemSyncsCoalesce (two concurrent syncTransactions for the same item produce exactly one underlying Plaid sync call, maxActive==1, both callers see the same result)

## For the reviewer
- Could not run swift build/test in the worktree (orchestrator compiles). Highest-risk compile point is the unstructured Task inside the new actor under -strict-concurrency=complete -warnings-as-errors; the operation closure and all captures (self/item/itemId) are Sendable so it should be clean, but worth confirming in CI.
- The coalescing test relies on a 200ms Plaid-call delay to keep the first sync in flight while the second lands on the gate; if a slow/loaded CI host serializes the two async-let calls before overlap, the maxActive==1 assertion still holds (coalescing keys by itemId), but the syncs==[token-a] assertion is the load-bearing check.
- deleteItem now holds the cursor-write lock across its Fluent find/delete calls (Keychain delete stays outside the lock); this serializes deletes against conditional cursor saves only, not against unrelated reads, so contention impact is negligible for a single-user local server.

## Source
Pre-release audit 2026-06-16 — **AND-471** / #447. **Draft — do not merge yet** (part of the audit-fix stack; merge per the wave plan).

### DIFF
--- a/Sources/PlaidBarServer/Routes/ItemSyncCoalescer.swift
+++ b/Sources/PlaidBarServer/Routes/ItemSyncCoalescer.swift
@@ -1,0 +1,30 @@
+/// Coalesces concurrent transaction syncs keyed by Plaid item id so only one
+/// sync runs per item at a time. Concurrent callers for the same item await the
+/// in-flight result instead of issuing a second overlapping `/transactions/sync`
+/// pagination against Plaid (which races on the persisted cursor and can double
+/// the work). Distinct items still run concurrently.
+///
+/// The companion server is a single local process, so an in-process actor fully
+/// orders the per-item gate. The in-flight entry is cleared once the work
+/// completes (success or failure), so the next sync re-runs normally.
+actor ItemSyncCoalescer<Value: Sendable> {
+    private var inFlight: [String: Task<Value, Error>] = [:]
+
+    /// Runs `operation` for `itemId`, coalescing with any sync already in
+    /// flight for the same item. The first caller starts the work; subsequent
+    /// callers await the same task and observe its result (or error).
+    func run(
+        itemId: String,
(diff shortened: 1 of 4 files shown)


ftchvs:

Issue #448, opened 2026-06-17
[Audit 2026-06-16] Security defense-in-depth: CSPRNG fallback, localhost Origin/rate-limit, status contract (AND-472)


ftchvs:
**Tracked in Linear: AND-472** · Source: pre-release audit 2026-06-16 · **Do not implement yet.**

**Severity:** Low. Core boundary is strong (256-bit CSPRNG token, 0600 file, constant-time compare, loopback-only); these are hardening + doc reconciliation.

- #17 Auth-token gen silently downgrades to non-CSPRNG UUID if `SecRandomCopyBytes` fails (ServerConfig.swift:178-189) — no warning/throw.
- #18 Localhost API has no rate-limit/lockout or Origin/Referer check (APITokenMiddleware.swift:11-24).
- #19 `/api/status?include=items` + `/api/items` expose item IDs/institution names (StatusRoutes.swift:22/39/60,80-81), contradicting `SECURITY.md:59-63` and `docs/privacy.md:104-108`.

**Recommendation:** make CSPRNG failure fatal/logged; reject browser Origin/Referer + add failed-auth backoff; reconcile the status contract (update docs OR move itemStatuses to /api/items only).

ftchvs:
Fixed in #458 (squash-merged to main 2026-06-17). Linear: AND-472. Verified locally: strict-concurrency build + full test suite green (CI is billing-blocked).

PR #458, opened 2026-06-17
fix: fail-loud CSPRNG, reject foreign Origin, reconcile status contract (AND-472)

ftchvs:
## Summary
Security defense-in-depth plus status-contract doc reconciliation. (#17) ServerConfig secureRandomBytes/generateAuthToken now throw a new ServerConfigError.secureRandomUnavailable on CSPRNG failure instead of silently downgrading to UUID material; both call sites (auth token, link client user id) were already throws and now propagate it, and the happy path is unchanged. (#18) APITokenMiddleware adds a cheap origin check after the constant-time bearer check: any /api request whose Origin or Referer resolves to a non-loopback host gets 403; native app (no headers) and loopback origins pass. (#19) SECURITY.md and docs/privacy.md now document that /api/status?include=items and /api/items expose authenticated, opt-in item-health metadata (item_id, institution name, status, last sync/webhook, pending flag), matching the code and ServerStatus doc-comment, while keeping the no-tokens/no-account-IDs/no-balances/no-transactions guarantees intact.

- Tests added/updated: PlaidBarServerTests.apiMiddlewareRejectsForeignBrowserOrigin; PlaidBarServerTests.apiMiddlewareAllowsNativeAppAndLoopbackOrigins; PlaidBarServerTests.apiOriginAllowlistClassifiesHosts; PlaidBarServerTests.makeRequest (helper extended with origin/referer params)

## For the reviewer
- #17: No Swift Testing test added for the secureRandomUnavailable throw path. secureRandomBytes is private static and unreachable on the macOS-only build (canImport(Security) is always true on macos-15 CI), so the throw is effectively defensive/dead on the supported platform and cannot be exercised without making the function internal or injecting a failing RNG. Left the API surface unchanged to keep the diff minimal; flagging in case a testability seam is desired.
- #18: The origin allowlist permits any loopback host (127.0.0.1/localhost/[::1]) on any port, not strictly the configured server port. This is intentional so a local dev tool on a different loopback port still works, but if a stricter same-port-only policy is wanted that is a product decision.

## Source
Pre-release audit 2026-06-16 — **AND-472** / #448. **Draft — do not merge yet** (part of the audit-fix stack; merge per the wave plan).

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -56,11 +56,15 @@ Use GitHub private vulnerability reporting if available, or contact the reposito
 - Existing legacy `plaidbar.sqlite` data, SQLite sidecar files, and matching transaction cache are copied into a scoped database only when the legacy environment is explicit (`PLAIDBAR_MIGRATE_LEGACY_DATABASE=sandbox|production`) or can be inferred from the existing transaction-cache context. Ambiguous legacy databases are left untouched to avoid sandbox/production token crossover. Explicit migration can replace an existing scoped store, backs up the previous scoped SQLite store and transaction cache before copying legacy data, and writes a migration marker so restarts do not reapply stale legacy data.
 - The app/server auth token is generated locally under `~/.vaultpeek/auth-token`
   (or `$PLAIDBAR_DATA_DIR/auth-token`).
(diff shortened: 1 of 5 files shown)


ftchvs:

Issue #440, opened 2026-06-17
[Audit 2026-06-16] Offline-at-boot never starts the background refresh loop (AND-464)


ftchvs:
**Tracked in Linear: AND-464** · Source: pre-release audit 2026-06-16 · **Do not implement yet.**

**Severity:** Medium · Verified against current code.

If the local server is unreachable during `loadInitialData`, the background polling loop never bootstraps:
- `startBackgroundRefresh()` is called only inside the `serverConnected` branch (AppState.swift:1916) and at :172 (guarded by `if refreshTask != nil`, a no-op when never started).
- The offline `else` branch (1917-1923) only runs `evaluateNotifications()` once; `prewarmBundledServer()` (1929-1950) never starts the loop.

**Recommendation:** start the refresh loop unconditionally at the end of `loadInitialData()`; it self-heals via `checkServerConnection`.

ftchvs:
Fixed in #462 (squash-merged to main 2026-06-17). Linear: AND-464. Verified locally: strict-concurrency build + full test suite green (CI is billing-blocked).

PR #462, opened 2026-06-17
fix: start background refresh loop unconditionally at boot (AND-464)

ftchvs:
## Summary
Fixed the offline-at-boot bug where the background refresh loop never started. Previously startBackgroundRefresh() was called only inside the serverConnected branch of loadInitialData(), so an offline cold start only ran evaluateNotifications() once and then sat frozen forever. Added an idempotent startBackgroundRefreshIfNeeded() helper (guards against double-start when refreshTask is already running) and moved a single unconditional call to the end of loadInitialData() so both the online and offline boot paths leave the loop running. The loop re-probes connectivity via refreshDashboard -> checkServerConnection and is gated by the app-lock predicates, so it self-heals once the server comes up. The connected-branch's previous direct startBackgroundRefresh() call was removed to avoid double-start; the refreshInterval didSet force-restart and stopBackgroundRefresh paths remain unaffected. Demo mode still returns early before the bootstrap (intentional — static fixtures, no live loop needed).

## For the reviewer
- No test added: the bootstrap is not testable in isolation. AppState, startBackgroundRefresh(), startBackgroundRefreshIfNeeded(), and loadInitialData() all live in the PlaidBar executable target (@main SwiftUI app), which cannot be @testable imported. The Tests/PlaidBarTests target only exercises PlaidBarCore shared logic (per its own header comment, lines 8-11). There is no PlaidBarCore seam for the refresh-loop lifecycle, so the fix cannot be proven via Swift Testing without introducing a larger testability refactor (extracting loop-bootstrap state into a Sendable core type) that would exceed this issue's minimal scope.

## Source
Pre-release audit 2026-06-16 — **AND-464** / #440. **Draft — do not merge yet** (part of the audit-fix stack; merge per the wave plan).

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -1761,6 +1761,14 @@ final class AppState {
         return result
     }
 
+    /// Starts the background refresh loop only if one is not already running.
+    /// Used by the boot bootstrap so both the online and offline `loadInitialData()`
+    /// paths leave the self-healing loop running without restarting an existing one.
+    func startBackgroundRefreshIfNeeded() {
+        guard refreshTask == nil else { return }
+        startBackgroundRefresh()
+    }
+
     func startBackgroundRefresh() {
         refreshTask?.cancel()
         refreshTask = Task {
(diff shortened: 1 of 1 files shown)


ftchvs:

Issue #439, opened 2026-06-17
[Audit 2026-06-16] Menu-bar render performance: cache presentations, fix O(n²) inbox & date-parse hot paths (AND-463)


ftchvs:
**Tracked in Linear: AND-463** · Source: pre-release audit 2026-06-16 · **Do not implement yet.**

**Severity:** High · Top perf win for the always-on menu bar (non-blocking).

- `weeklyReviewPresentation` (AppState.swift:855), `categoryBudgetPresentation` (891-915), `transactionReviewInboxSnapshot` (989) are uncached and rescan all transactions each `MenuBarLabel.body`; inbox `isUnusual` → `transactions.filter` per transaction = **O(n²)** (TransactionReview.swift:289,295).
- Heatmap `currentLayout()` re-aggregates the full year on every cell selection/arrow move (MainPopover.swift:694-704).
- `cashflowSummary` parses a DateFormatter per transaction (WealthSummaryPresentation.swift:225-232); `cellLabel` parses+formats per ~365 cells (MainPopover.swift:1010).
- `ServerClient.authorizedRequest` reads the token file from disk every request (ServerClient.swift:280-293).

**Recommendation:** cache via the existing `didSet`-invalidated pattern; canonical-string date compare; precompute cell labels; cache the token in memory (invalidate on 401).

ftchvs:
Fixed in #463 (squash-merged to main 2026-06-17). Linear: AND-463. Verified locally: strict-concurrency build + full test suite green (CI is billing-blocked).

PR #463, opened 2026-06-17
perf: cache menu-bar presentations and remove O(n^2) inbox scan (AND-463)

ftchvs:
## Summary
Killed the O(n^2) review-inbox scan and the repeated per-render full-transaction recomputes. In TransactionReview.swift, TransactionReviewInbox.evaluate now builds a single UnusualSpendPeerIndex (per-merchant and per-category count+sum aggregates over spend transactions) up front, and isUnusual does O(1) lookups instead of two .filter scans per row. Exact output semantics are preserved by subtracting the current transaction's own contribution to reproduce the original id != self exclusion, including the subtlety that the category fallback keys on the raw category while the lookup uses the effective category. In AppState.swift, categoryBudgetPresentation and transactionReviewInboxSnapshot are now memoized with the existing didSet-invalidation pattern (transactions/categoryBudgets for the budget cache; transactions/transactionReviewMetadata/transactionRules for the inbox cache), so MenuBarLabel.body no longer re-runs those scans multiple times per frame. weeklyReviewPresentation was deliberately left uncached (see open questions) but now benefits indirectly because it already routes through the now-memoized categoryBudgetPresentation and recurringTransactions. Added a Swift Testing test asserting the cached-aggregate unusual-amount decision is byte-for-byte identical to a frozen brute-force replica of the original scan on a mixed fixture exercising both the merchant-peer and category-fallback paths.

- Tests added/updated: unusualAmountMatchesBruteForceScan

## For the reviewer
- weeklyReviewPresentation was NOT given its own cache. Its inputs span ~15 fields (accounts, transactions, weeklyReviewState, transactionReviewMetadata, hasLoadedTransactionReviewStorage, isDemoMode, categoryBudgets, itemStatuses, serverConnected, serverCredentialsConfigured, serverItemCount, serverSyncedItemCount, lastSyncDate, error, creditUtilizationThreshold, balanceHistory, isBooting, refreshInterval) plus Date()/isSyncStale which is purely time-derived. Most of those have no didSet and are assigned at many scattered sites with no choke point, so a didSet-invalidated cache would be fragile and risk staleness (e.g. a sync going stale by elapsed time would never refresh until some unrelated field mutated). Left uncached per the smallest-correct-change rule; it now reuses the two newly memoized vars so its per-access cost is already much lower. Caching it safely likely needs a small data-version token bumped at the central status-refresh choke point plus an explicit time-bucket key, which is a follow-up.
- Caching transactionReviewInboxSnapshot means the recurringChanged/isStale(asOf: now) signal no longer re-evaluates on elapsed time alone; it refreshes when transactions/metadata/rules change (i.e. on each sync). This matches the pre-existing _cachedRecurringTransactions tradeoff and is consistent with the codebase's memoization philosophy, but confirm it is acceptable that a charge crossing the recurring-stale time boundary between syncs is reflected only after the next data mutation.
- Floating-point note: the aggregate path sums all qualifying spend then subtracts the current row's magnitude, vs the original which summed only the peer subset. Differences are ~1e-13 relative and far below the max(avg*1.75, avg+50) threshold for realistic dollar amounts, so no boundary flip is expected; the equivalence test passes on the fixture, but a pathological near-threshold dataset is theoretically distinguishable.

## Source
Pre-release audit 2026-06-16 — **AND-463** / #439. **Draft — do not merge yet** (part of the audit-fix stack; merge per the wave plan).

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -50,11 +50,17 @@ final class AppState {
         didSet {
             _cachedTransactionDerivedIndex = nil
             _cachedRecurringTransactions = nil
+            _cachedCategoryBudgetPresentation = nil
+            _cachedTransactionReviewInboxSnapshot = nil
             invalidateLocalAIActivitySummaries()
         }
     }
-    var transactionReviewMetadata: [TransactionReviewMetadata] = []
-    var transactionRules: [TransactionRule] = []
+    var transactionReviewMetadata: [TransactionReviewMetadata] = [] {
+        didSet { _cachedTransactionReviewInboxSnapshot = nil }
+    }
+    var transactionRules: [TransactionRule] = [] {
+        didSet { _cachedTransactionReviewInboxSnapshot = nil }
+    }
     var hasLoadedTransactionReviewStorage = false
     var isLoading = false
     /// True from launch until the first `loadInitialData()` pass completes.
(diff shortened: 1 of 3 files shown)


ftchvs:

PR #469, opened 2026-06-17
test: stabilize glance snapshot debounce assertion

### LINKED ISSUE #468: [Bug hunter] Stabilize GlanceSnapshot debouncer test under full-suite load
## Summary
`swift test --filter PlaidBarCoreTests` intermittently fails on `GlanceSnapshotTests.debouncerCoalescesBurstWritesToLatestSnapshot` because the test sleeps for a fixed 80 ms and then assumes the debounced task has executed. Under full-suite load, the scheduled write may not run before the assertion, producing an empty recorder array even though the one-test focused run passes.

## Evidence
- Command: `swift test --filter PlaidBarCoreTests 2>&1 | tee /tmp/vaultpeek-coretests.log`
- Failure: `Expectation failed: (snapshots.map(\.netWorth) → []) == ([300] → [300.0])` at `Tests/PlaidBarCoreTests/GlanceSnapshotTests.swift:129`.
- Focused command `swift test --filter debouncerCoalescesBurstWritesToLatestSnapshot` passed, consistent with scheduler/load-sensitive test timing.

## Proposed fix
Make the regression test wait for the debounced write to appear within a bounded timeout instead of asserting immediately after one fixed sleep. Keep the production debouncer unchanged.

## Privacy/security
No private financial data, Plaid tokens, account IDs, transaction IDs, balances, prompts, or response bodies are involved. This is a test-only reliability fix for the WidgetKit/glance snapshot debounce gate.

ftchvs:
## Summary
- Replaces the fixed 80 ms assertion delay in the GlanceSnapshot debouncer test with a bounded wait for the async write to be observed.
- Keeps the production debouncer unchanged; this is a test reliability fix for full-suite scheduler load.

## Verification
- `git diff --check`
- `swift test --filter debouncerCoalescesBurstWritesToLatestSnapshot`
- `swift test --filter PlaidBarCoreTests`

Fixes #468
Linear: AND-477
Kanban: t_550b0fc6

### DIFF
--- a/Tests/PlaidBarCoreTests/GlanceSnapshotTests.swift
+++ b/Tests/PlaidBarCoreTests/GlanceSnapshotTests.swift
@@ -124,8 +124,7 @@ struct GlanceSnapshotTests {
             await recorder.record(snapshot)
         }
 
-        try await Task.sleep(for: .milliseconds(80))
-        let snapshots = await recorder.snapshots
+        let snapshots = try await recorder.waitForSnapshotCount(1, timeout: .milliseconds(500))
         #expect(snapshots.map(\.netWorth) == [300])
     }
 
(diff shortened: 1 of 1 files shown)


ftchvs:
Closing as a duplicate of #472, which is the more robust fix. Per adversarial review: #469 only widens the wait (80ms→500ms poll) and stays wall-clock-dependent, whereas #472 injects a DebounceScheduler seam so the coalescing/cancellation path is exercised with zero timing dependence (production behavior unchanged). #472 merged. (Issue #468.)

PR #473, opened 2026-06-17
fix: strip pending transaction ids from local cache

### LINKED ISSUE #435: Move posted↔pending reconciliation off the persisted raw pending_transaction_id (Codex P1 from #434)

ftchvs:
## Summary
- Strip Plaid `pending_transaction_id` from the local transaction cache before writing `transactions*.json`.
- Apply the same sanitization while remapping legacy scoped transaction caches during migration.
- Add a regression test proving the cache payload and loaded local snapshot do not retain the raw pending id.

Fixes #435
Linear: AND-476

## Verification
- Reproduced before the fix: `swift test --filter localTransactionCacheStripsPendingTransactionIdsBeforePersistence` failed with cache JSON containing `pendingTransactionId` and the pending id value.
- `swift test --filter localTransactionCacheStripsPendingTransactionIdsBeforePersistence` — 1 test passed.
- `swift test --filter transactionSyncReducerReconcilesPendingToPosted` — 1 test passed.
- `git diff --check` — passed (no output).
- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — passed.

## Privacy / security
- Server sync responses can still carry `pendingTransactionId` in memory for one-sync reconciliation.
- Local cache persistence now drops that field so the app does not durable-write the Plaid pending transaction identifier into `transactions*.json`.
- No credentials, balances, raw account ids, raw transaction ids, or payload bodies are included here.

### DIFF
--- a/Sources/PlaidBarCore/Utilities/LocalDataStore.swift
+++ b/Sources/PlaidBarCore/Utilities/LocalDataStore.swift
@@ -576,7 +576,7 @@ public enum LocalDataStore {
 
         let remappedCache = TransactionCache(
             context: remappedContext,
-            transactions: cache.transactions
+            transactions: transactionsForLocalPersistence(cache.transactions)
         )
         return (
             transactionCacheURL(in: currentDirectory, context: remappedContext),
@@ -696,7 +696,10 @@ public enum LocalDataStore {
         try ensurePrivateDirectory(directory, fileManager: fileManager)
 
         let url = transactionCacheURL(in: directory, context: context)
-        let cache = TransactionCache(context: context, transactions: transactions)
+        let cache = TransactionCache(
+            context: context,
+            transactions: transactionsForLocalPersistence(transactions)
+        )
         let data = try JSONEncoder().encode(cache)
         try writePrivateCacheFile(data, to: url, fileManager: fileManager)
         try setPrivateCacheFilePermissions(url, fileManager: fileManager)
@@ -757,6 +760,24 @@ public enum LocalDataStore {
 
         let key = "\(context.environment.rawValue)|\(context.storagePath)"
         return "transactions-\(context.environment.rawValue)-\(stableHashHex(key)).json"
+    }
+
+    private static func transactionsForLocalPersistence(_ transactions: [TransactionDTO]) -> [TransactionDTO] {
+        transactions.map { transaction in
+            TransactionDTO(
+                id: transaction.id,
+                itemId: transaction.itemId,
+                accountId: transaction.accountId,
+                amount: transaction.amount,
+                date: transaction.date,
+                name: transaction.name,
+                merchantName: transaction.merchantName,
+                category: transaction.category,
+                pending: transaction.pending,
+                pendingTransactionId: nil,
+                isoCurrencyCode: transaction.isoCurrencyCode
+            )
+        }
     }
 
     // Review metadata and merchant rules are scoped to the active cache context
--- a/Tests/PlaidBarCoreTests/PlaidBarCoreTests.swift
+++ b/Tests/PlaidBarCoreTests/PlaidBarCoreTests.swift
@@ -3869,6 +3869,32 @@ struct PlaidBarCoreTests {
         #expect(resolved == directory)
     }
 
+    @Test("Local transaction cache strips pending transaction ids before persistence")
+    func localTransactionCacheStripsPendingTransactionIdsBeforePersistence() throws {
+        let directory = URL(fileURLWithPath: NSTemporaryDirectory())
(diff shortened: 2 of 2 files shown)


ftchvs:
Closing — consistent with the by-design resolution of #435.

This strips `pendingTransactionId` from the persisted transaction cache (`LocalDataStore`) only, which has two problems:

1. **Introduces a cold-start regression.** `TransactionReview.evaluate` reads `transaction.pendingTransactionId` (TransactionReview.swift:196) to carry a user's pending-phase review state forward to the posted charge. At cold start the transactions are cache-loaded, so stripping the field there breaks pending→posted review carry-forward until the next live sync re-populates it — a charge the user already reviewed while pending re-appears as needs-review after a restart (and for the whole session if offline). The added test asserts only that the id is absent from the cache; it does not cover carry-forward, so the regression is uncaught.
2. **Zero net privacy gain.** The same id value is necessarily persisted as a `TransactionReviewMetadata.id` key (that's what makes carry-forward survive a restart; metadata is only pruned on item-removal, not normal sync). It's also the same sensitivity class as `TransactionDTO.id`, already pervasively persisted to the same 0600 cache, never surfaced to a View/log/network.

So this trades a real (if transient) correctness regression for no privacy improvement. See #435 for the full analysis; the only genuine path is a metadata re-keying redesign (deferred there). Please also close Linear AND-476. Reopen if you want the re-keying redesign despite the regression risk.

PR #479, opened 2026-06-17
fix: engage App Lock before first launch render

### LINKED ISSUE #478: Lock-on-launch can leak first dashboard frame before App Lock engages
## Finding

`AppState.loadInitialData()` currently calls `lockOnLaunchIfNeeded()` from the SwiftUI `.task` path. SwiftUI evaluates the first `MainPopover.body` before that task runs, so a user with App Lock + Lock on Launch enabled can get one initial frame of cached/demo dashboard content before `isContentLocked` flips true and the lock gate overlays it.

## Evidence

- `Sources/PlaidBar/Views/MainPopover.swift` renders `chromedPopover` and starts `await appState.loadInitialData()` in `.task`.
- `Sources/PlaidBar/App/AppState.swift` only engaged `lockOnLaunchIfNeeded()` inside `loadInitialData()`, after initial state creation and first render opportunity.
- App Lock's content gate depends on `isContentLocked`, which depends on `isAppLocked`; before the task runs, `isAppLocked` mirrors the persisted service state but Lock on Launch has not been applied.

## Privacy impact

First paint can reveal account/institution/activity labels from local cached data even though the user explicitly selected Lock on Launch. No Plaid credentials/tokens are involved; this is a local UI privacy leak.

## Acceptance criteria

- Lock-on-launch is applied synchronously during `AppState` initialization, before the state is placed in SwiftUI environment and before the first popover/detached dashboard render.
- The load task may re-assert the lock idempotently, but it is not the first point where launch locking occurs.
- Focused App Lock policy tests and strict-concurrency build pass.

ftchvs:
## Summary
- apply Lock on Launch synchronously during `AppState` initialization, before SwiftUI can render the first dashboard frame
- keep the existing load-task lock assertion as an idempotent defensive guard
- add Core policy predicates/tests for launch/background lock triggers

Fixes #478
Linear: AND-482
Kanban: t_5767e072

## Verification
- `swift test --filter AppLockPresentation` — 11 tests passed
- `git diff --check` — passed
- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — Build complete

## Privacy/security
Local UI privacy fix only. No Plaid credentials, provider tokens, account IDs, transaction IDs, balances, local SQLite data, prompts, or response bodies are moved into app/docs/tests.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -354,6 +354,12 @@ final class AppState {
         _ = try? LocalDataStore.migrateLegacyDefaultStorageIfNeeded()
         self.notificationService = notificationService ?? NotificationService.shared
         loadSettings()
+        // Engage launch App Lock synchronously before the first SwiftUI body is
+        // ever evaluated. Deferring this to `loadInitialData()` leaks one frame
+        // of cached/demo account names and activity because `.task` runs after
+        // the initial render; the lock-on-launch privacy contract needs to be
+        // true as soon as AppState enters the environment.
+        lockOnLaunchIfNeeded()
         isSetupComplete = storedSetupCompletion()
         isFirstRunSnapshotDismissed = storedFirstRunSnapshotDismissal()
         if isSetupComplete {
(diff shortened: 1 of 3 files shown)


ftchvs:
@codex please review

PR #481, opened 2026-06-17
fix: prefer an instruct model for Local AI insights (fixes "Local Error")

ftchvs:
## The report
"Local Error" on the Local Insight Receipt — even with Ollama running and models installed.

## Diagnosis (not a broken integration)
Ollama is healthy. Model auto-discovery preferred **`gpt-oss`** first (it was #1 in `preferredModelPrefixes`), so it picked **`gpt-oss:20b`** — a **reasoning model**. For the short summary prompt it spends its small token budget on chain-of-thought and returns an **empty `response`** (reproduced: `RESPONSE: ''`). VaultPeek correctly rejects empty/unusable output → deterministic fallback → "Local Error / returned invalid or unsafe output." A small instruct model (`llama3.2:1b`) returns clean usable output to the same prompts.

## Fix
- **Prefer instruct models:** reorder `preferredModelPrefixes` to llama3.2 / llama3.1 / qwen2.5 / gemma / mistral / phi first; drop the gpt-oss reasoning family from the preferred set.
- **Accurate probe:** a model that responds but returns empty/blank output is reported as `.invalidOutput` with an actionable hint, instead of being marked "available" and then failing on the real summary call.

## Verification (CI/Codex billing-blocked; app-target code isn't @testable-importable)
- `swift build -strict-concurrency=complete -warnings-as-errors` ✅ · `swift test` ✅ **853**.
- **Live Ollama:** simulated the new discovery order against the installed models → now selects **`llama3.2:1b`** (usable output) instead of `gpt-oss:20b` (empty). The Local Insight Receipt will render a real on-device summary.

Tip: users can still pin a specific model via `PLAIDBAR_LOCAL_AI_MODEL`.

### DIFF
--- a/Sources/PlaidBar/Services/LocalAIInsightsService.swift
+++ b/Sources/PlaidBar/Services/LocalAIInsightsService.swift
@@ -138,7 +138,21 @@ struct LocalAIInsightsService: Sendable {
         }
 
         do {
-            _ = try await boundedProbe(model: model)
+            let probeOutput = try await boundedProbe(model: model)
+            // A model that responds but emits nothing usable (e.g. a reasoning
+            // model that spends a short token budget on chain-of-thought and
+            // returns an empty `response`) must NOT be reported as available —
+            // the real summary call would hit the same empty output and be
+            // rejected. Surface it as invalid output with an actionable hint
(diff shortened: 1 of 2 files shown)


ftchvs:
@codex please review

PR #483, opened 2026-06-17
fix: suppress the system focus ring on the embedded Review Inbox

ftchvs:
## What you saw
A blue rounded outline around the **entire** Review Inbox column — looked like the whole panel was selected.

## Why
It's the macOS **system keyboard-focus ring**. `ReviewInboxView`'s scroll container is `.focusable()` and auto-grabs focus on appear (`onAppear { isFocused = true }`) to enable arrow-key row navigation (`.onMoveCommand`). Embedded as the full right-hand inspector column, the focus ring wraps the whole panel. The actually-selected *row* already has its own accent (leading bar + tinted wash), so the panel-wide ring is redundant and misreads as an error/selected state.

## Fix
Add `.focusEffectDisabled()` to the focusable container — keeps focusability and arrow-key navigation, drops the visual ring. (macOS 14+; the app targets macOS 15.)

## Verification
- `swift build -strict-concurrency=complete -warnings-as-errors` ✅
- App-target SwiftUI (not @testable-importable); no logic change, purely the focus *effect*.

### DIFF
--- a/Sources/PlaidBar/Views/ReviewInboxView.swift
+++ b/Sources/PlaidBar/Views/ReviewInboxView.swift
@@ -96,6 +96,11 @@ struct ReviewInboxView: View {
             .modifier(ConditionalRaisedSurface(embedded: embedded))
             .focusable()
             .focused($isFocused)
+            // Keep the container focusable for arrow-key row navigation, but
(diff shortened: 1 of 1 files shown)


ftchvs:
@codex review

PR #489, opened 2026-06-18
feat: pull live balances on the manual refresh (AND-495)

ftchvs:
## What & why
The in-app **manual refresh** (refresh button → `refreshDashboard(force:)`) used the cached `/accounts/get`. Genuinely-live balances via `/accounts/balance/get` were only reachable from the **widget's** `refreshBalances` glance command — so "am I about to overdraft?" at the moment of a manual refresh showed a 1–4×/day cached value.

## Change
- `refreshAccounts(live:)` gains a `live` flag → `live ? getBalances() : getAccounts()`.
- `refreshDashboard(force:)` passes `live: force`.

So a **user-initiated** refresh pulls a fresh institution balance; **automatic** throttled ticks (and refresh-on-open) stay on the cached endpoint and never add a billable live call — the cost/non-goal guardrail AND-495 calls out. `refreshBalances()` (widget glance command) is untouched.

## Verification
- `swift test --skip PlaidBarServerTests` — 699 app+core tests pass (server keychain suite skipped only because this branch predates the AND-481 fix; full suite runs in CI).

## Please review
- That auto-refresh truly stays cached: only `refreshDashboard(force:)` passes `live: force`; the auto-on-open path (`AppState` ~2145) calls `refreshAccounts()` (cached) directly, as its comment intends.
- Whether the two edge-case user refreshes (Settings empty-state, reconnect fallback) should also be `live` — left cached for now to keep the change minimal.

Closes AND-495.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -1479,6 +1479,50 @@ final class AppState {
         await startBundledServerIfAvailable()
     }
 
+    /// Minimum interval between billable live `/accounts/balance/get` refreshes,
+    /// so rapid clicks of the refresh button/control can't run up per-call Plaid
+    /// balance fees.
+    private static let liveBalanceRefreshCooldown: TimeInterval = 30
+    private var lastLiveBalanceRefreshAt: Date?
+
+    /// True when a live balance refresh is allowed (first time, or past the
+    /// cooldown window since the last live call).
+    private func liveBalanceRefreshAllowed() -> Bool {
+        guard let last = lastLiveBalanceRefreshAt else { return true }
+        return Date().timeIntervalSince(last) >= Self.liveBalanceRefreshCooldown
+    }
+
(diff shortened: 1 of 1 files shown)


ftchvs:
@codex review

ftchvs:
Addressed all four Codex findings in the latest commit:
- **P1 cooldown** — live `/accounts/balance/get` is now gated to once per 30s; rapid refresh-button / "Refresh balances" control clicks fall back to cached, bounding paid calls. Auto ticks stay cached.
- **P1 `current: null`** — `balancesPreservingCachedCurrent` backfills nil `current`/`limit` from the cached balance on live refreshes, so debt/utilization (derived from `current ?? 0`) can't be zeroed.
- **P2 metric** — live refreshes now record `.balancesRefresh`, not `.accountsRefresh`.
- **P2 glance** — the macOS "Refresh balances" control now uses live balances too (same cooldown).

Strict build clean; 699 Core/app tests pass.

PR #490, opened 2026-06-18
feat: add an Investments account filter (AND-496)

ftchvs:
## What & why
Investment / brokerage / retirement **balances already counted** toward net worth and the wealth-mix composition, but there was no way to view those accounts in the list — the filters were Cash/Credit/Savings/Debt only. This completes net worth by making investments a first-class filter.

## Change
- New `investment` case in `DashboardAccountFilterKind`, predicate `account.type == .investment`.
- **Balance line only** — no holdings, performance, or allocation (the explicit non-goal; a portfolio dashboard stays out of scope).
- No Plaid `investments` product needed: `/accounts/get` already returns investment-type accounts with their institution-reported value.
- Chip auto-renders from `allCases`, grouped with the balance filters before Status → **Status moves ⌘6 → ⌘7**. Persisted `@AppStorage` selections are unaffected (raw-value based).

## Verification
- Strict-concurrency build clean.
- `swift test --skip PlaidBarServerTests` — 700 app+core tests pass (server keychain suite skipped only because this branch predates AND-481; full suite runs in CI). Added an investment-matching test; updated ordinal/help-text assertions.

## Please review
- Filter **order**: investments before Status (shifts Status's ⌘ shortcut). Acceptable, or prefer investments last to keep ⌘6 = Status?
- 7 segments in the menu-bar segmented control — still legible at popover width?

Closes AND-496.

### DIFF
--- a/Sources/PlaidBar/Views/DashboardNavBand.swift
+++ b/Sources/PlaidBar/Views/DashboardNavBand.swift
@@ -5,8 +5,9 @@ import SwiftUI
 
 /// The dashboard account filter is the core `DashboardAccountFilterKind`
 /// reused directly. The persisted `@AppStorage("dashboard.accountFilter")`
-/// raw values ("All", "Cash", "Credit", "Savings", "Debt", "Status") are the
-/// core enum's raw values, so stored selections decode exactly as before.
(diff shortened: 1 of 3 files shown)


ftchvs:
@codex review

PR #492, opened 2026-06-18
feat: Real Merchants — Plaid category confidence + cached merchant logos (AND-494)

ftchvs:
Completes AND-494 — both halves of Real Merchants.

## Part 1 — category confidence → Review Inbox
Plaid PFCv2 returns a `confidence_level` per categorization, which the server **decoded then discarded**. Preserved on `TransactionDTO.categoryConfidence`; the Review Inbox now surfaces LOW/UNKNOWN-confidence categories (gated on no user override) and stops flagging confidently-categorized items. No new Plaid call — the field rides the existing `/transactions/sync`.

## Part 2 — cached merchant logos (server-side proxy)
Plaid enrichment provides a `logo_url` (Plaid CDN PNG) per transaction. The real brand mark now renders in the transaction list, keeping the app **strictly localhost-only**:
- `PlaidTransaction.logoUrl` → `TransactionDTO.logoURL` → transform.
- New authed `GET /api/merchant-logo`: **server-side** fetch + **on-disk cache** under the data dir, **host-allowlisted** to Plaid logo CDNs. The app never reaches a CDN directly; only public brand images are proxied (no user data).
- `MerchantLogoStore` (@MainActor @Observable) caches `NSImage` via the proxy; `MerchantLogoView` renders the logo with a tinted-dot fallback that preserves the income/outflow signal. Wired into `TransactionMiniRow` (account detail).

## Verification
- `swift test --skip PlaidBarServerTests` — **700** Core/app tests pass (new low-confidence Review-Inbox test; no regression from the logo plumbing).
- The logo proxy + render are covered by the build; **recommend a sandbox pass** for live image verification (demo has no logo URLs).

## Please review
- Part 1: the no-user-override gate (don't re-flag a user's own category).
- Part 2: the unauthenticated-vs-authed choice (this uses **authed** `/api/merchant-logo` + host allowlist); the leading-avatar size bump in `TransactionMiniRow`; whether the trust line belongs here or with AND-491's "Where Your Data Lives" panel.

Closes AND-494.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -367,6 +367,8 @@ final class AppState {
 
     // MARK: - Services
     private let serverClient = ServerClient()
+    /// Fetches + caches merchant logos via the local server's authed proxy.
+    let merchantLogoStore = MerchantLogoStore()
     private let localDataCache = LocalDataCacheService()
     private let appLockService = AppLockService()
     private let reviewStorageWriter = ReviewStorageWriter()
--- a/Sources/PlaidBar/Services/MerchantLogoService.swift
+++ b/Sources/PlaidBar/Services/MerchantLogoService.swift
@@ -1,0 +1,77 @@
+import AppKit
+import SwiftUI
+
+/// Fetches and caches merchant logos through the local server's authenticated
+/// proxy (`/api/merchant-logo`). The app never reaches a logo CDN directly: the
+/// server does the fetch and on-disk caching, keeping the UI localhost-only.
+@MainActor
+@Observable
+final class MerchantLogoStore {
+    private let serverClient: ServerClient
+    private var images: [String: NSImage] = [:]
+    private var failed: Set<String> = []
+    private var inFlight: Set<String> = []
+
+    init(serverClient: ServerClient = ServerClient()) {
+        self.serverClient = serverClient
+    }
+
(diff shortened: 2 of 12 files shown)


ftchvs:
@codex review

ftchvs:
Addressed the Codex findings in the latest commit:
- **P1 raw confidence** — Plaid's `confidence_level` enum is now mapped server-side to an app-owned `TransactionDTO.isLowConfidenceCategory` bool; the raw enum no longer reaches the app. Custom `init(from:)` keeps older cached `transactions.json` decodable.
- **P1 reset cache** — `resetLocalData` now clears the `logo-cache` directory, so a reset removes cached brand images.
- **P2 redirect** — the logo proxy re-validates the FINAL response host against the Plaid-CDN allowlist after following redirects.
- **P2 retry** — transport/5xx/server-starting errors no longer permanently mark a logo failed; only invalid image bytes are cached as failed, so logos retry once the server/network recovers.

**On the remaining P1 (raw Plaid logo URL in the DTO):** this is a real logo-pipeline redesign — an opaque token in the DTO plus a server-side token→URL index threaded through the sync transform. Worth noting the logo URL reveals only the merchant, which is already present as `merchantName`, so it adds no new user data beyond the existing transaction. I scoped it as a focused follow-up rather than rush the redesign in this batch — happy to implement the opaque-token version next if you'd like it in this PR.

Strict build clean; 700 Core/app tests pass.

PR #520, opened 2026-06-19
[macOS 26] Epic E — Control Center controls, Widgets, Spotlight

ftchvs:
## Summary

Epic E of the macOS 26 migration (AND-513), stacked on Epic D's App Group snapshot store. Adds pinnable **Control Center controls** (Refresh balances + a new Privacy-Mask toggle), gives the glance **widget** a macOS 26 glass pass that reads live values from the shared App Group snapshot and honors the privacy mask, and adds a display-only **Spotlight** account index. All within the Epic E ownership boundary — no edits to `Package.swift`, entitlements, or `PlaidBarCore`.

Implements: **AND-513**

Build: **green** — `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` passes; the widget extension target builds clean.

Tests: **green** for the suite, with one caveat — `LocalInsightModelRuntimeTests` "Runtime timeout returns without waiting for cancellation-ignoring models" is a pre-existing wall-clock-timing flake (asserts `elapsed < 0.15`). It fails only under full-suite CPU contention and passes in isolation (0.005s). It lives in `PlaidBarCore` and is untouched by this change. No new tests were added: the Epic E logic lives in the `PlaidBar` (`@main`) and widget-extension targets, neither of which is reachable from the existing `PlaidBarTests` suite (which can only `@testable import PlaidBarCore`), and adding a test target would require editing `Package.swift` (owned by Epic D).

### E1 — Control Center controls
- Kept the existing `PlaidBarRefreshControl` button (Refresh balances) unchanged.
- Added `PlaidBarPrivacyMaskControl` — a `ControlWidgetToggle` driven by `SetPrivacyMaskIntent` (`SetValueIntent`) with a `PrivacyMaskValueProvider` that reads `FinanceSnapshot.isMasked`. State is conveyed by glyph shape (`eye` / `eye.slash`), never color alone.
- The toggle deliberately does **not** open the app (`openAppWhenRun` false) — masking is silent. It writes a command file to the shared App Group via the new `WidgetControlCommandStore`, mirrored by an app-side `PrivacyMaskControlCommandReader` + an `AppState.applyPendingPrivacyMaskControlCommand()` that drives the same `privacyMaskEnabled` path as the in-app eye toggle.

### E2 — Widget glass pass
- The glance widget now reads the live `isMasked` flag from `AppGroupSnapshotStore` and, when masked, dots every figure (`••••`), drops the sparkline, and swaps in a value-free accessibility summary.
- Container background switched to a macOS 26 glass-friendly tinted gradient via `.containerBackground(for: .widget) { ... }`.

### E3 — Spotlight account index (display-only)
- `AccountSpotlightIndexer` indexes account **names + last-4 + institution** only. It never indexes balances, transactions, or Plaid `account_id`/`item_id`; the searchable item id is a salted, non-reversible FNV-1a digest of the display fields. Selecting a result opens `vaultpeek://dashboard`. The index is cleared while figures are masked.

## Files changed
- `Sources/PlaidBarWidgetExtension/PlaidBarWidgetBundle.swift` — privacy-mask control, refresh control retained, glance widget masking + glass background.
- `Sources/PlaidBarWidgetExtension/WidgetControlCommandStore.swift` (new) — App-Group command channel for the privacy toggle (writer side).
- `Sources/PlaidBar/Controls/PrivacyMaskControlCommandReader.swift` (new) — app-side command reader + `AppState` applier.
- `Sources/PlaidBar/Spotlight/AccountSpotlightIndexer.swift` (new) — display-only CoreSpotlight index.

## Follow-ups
- **TODO(AND-513)** Wire `appState.applyPendingPrivacyMaskControlCommand()` into the `didBecomeActive` handler in `PlaidBarApp` (alongside `consumePendingGlanceCommand()`). That file is owned by Epic D; add the one-liner when the stack merges. Until then the privacy command file holds the latest desired state and is applied on the next activation/refresh that calls the method.
- **TODO(AND-513)** Call `AccountSpotlightIndexer.refresh(accounts:isMasked:)` from `AppState`/`SyncService` after accounts load and whenever `shouldMaskFinancialValues` flips. Those call sites are owned by other epics; indexing is display-only so deferring leaves zero data exposed.
- Privacy toggle eventual-consistency: the `ControlValueProvider` re-reads `FinanceSnapshot.isMasked`, which the app rewrites only after it applies the command. `ControlCenter.reloadAllControls()` keeps the toggle responsive; full snapshot sync happens on next app activation.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -3241,6 +3241,11 @@ final class AppState {
             // empty path so Spotlight/Siri/Shortcuts stop reporting removed-account
             // figures (AND-512).
             try? AppGroupSnapshotStore.clear()
+            // No accounts left to surface in search — drop the display-only
+            // Spotlight account index too so removed-account names don't linger
+            // (AND-513). `writeFinanceSnapshot` is skipped on this empty path, so
+            // the index clear must happen here explicitly.
+            AccountSpotlightIndexer.clear()
             Task {
                 await clearGlanceSnapshot()
                 await MainActor.run {
@@ -3333,6 +3338,14 @@ final class AppState {
                 )
             }
         }
+        // Keep the display-only Spotlight account index in lockstep with the
+        // finance snapshot: this is the shared seam for account load/refresh
+        // (via `writeGlanceSnapshot`), demo data, and the Privacy Mask / App Lock
+        // transitions (`togglePrivacyMask` / `lockApp`). `refresh` clears the
+        // index when masked, so masking removes account names from system search
+        // immediately; the empty-accounts clear lives in `writeGlanceSnapshot` /
+        // `clearGlanceSnapshot` (AND-513).
+        AccountSpotlightIndexer.refresh(accounts: accounts, isMasked: shouldMaskFinancialValues)
     }
 
     private func clearGlanceSnapshot() async {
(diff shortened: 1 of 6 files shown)


ftchvs:
Addressed review (AND-513): wired `applyPendingPrivacyMaskControlCommand()` into `PlaidBarApp`'s `didBecomeActive` handler so the Control Center Privacy Mask toggle now flips the app's mask, and deleted the dead `WidgetControlCommandStore.consumePrivacyCommand` (extension is writer-only; app consumes via its own reader). AppState.swift untouched. Strict-concurrency build green.

PR #531, opened 2026-06-19
fix(privacy): redact glance-snapshot.json under Privacy Mask / App Lock (AND-517)

ftchvs:
## Summary

When **Privacy Mask** or **App Lock** is active, the App-Group `glance-snapshot.json` — the file that feeds WidgetKit widgets and the Control Center glance — was still written with the user's real `netWorth`, `todayChange`, and `sparkline`. Those figures leaked on surfaces *outside* the gated popover (Home Screen / Notification Center widget, Control Center), even while the in-app dashboard was masked or locked.

This PR makes the on-disk glance snapshot **value-free whenever the app is masked or locked**, and rewrites the existing file on the mask/lock state transition so no stale real snapshot is left behind. The redaction is a pure, `Sendable`, unit-tested transform in `PlaidBarCore`.

Three cases covered:
- **(a) Build while masked/locked** — `AppState.writeGlanceSnapshot` builds the snapshot with `isMasked: shouldMaskFinancialValues`, so every periodic/refresh write during a masked/locked session is redacted.
- **(b) On-state-change rewrite** — `togglePrivacyMask` and `lockApp` now call `writeGlanceSnapshot` (which also re-emits the sibling App Intents `FinanceSnapshot`), so *enabling* mask/lock immediately overwrites any prior real file; a successful `unlockApp` that clears the mask restores the real figures.
- **(c) Unmasked path unchanged** — the default `make(...)` argument (`isMasked: false`) leaves existing callers and real snapshots byte-for-byte identical.

## Linear

- **AND-517** (P0) — Redact glance-snapshot.json on Privacy Mask / App Lock.

## Architectural decisions

- **Redaction as a pure transform in `PlaidBarCore`.** `GlanceSnapshot` gains an `isRedacted` flag and a `redacted()` method that zeroes `netWorth`/`todayChange`, empties `sparkline`, sets the flag, and preserves only the non-sensitive `updatedAt`/`isDemo` metadata. `make(..., isMasked:)` applies it. Keeping it in Core makes it `Sendable`, reusable, and unit-testable at the file level — defense in depth beyond the widget's read-time dotting.
- **Single mask predicate.** Both the glance snapshot and the existing `FinanceSnapshot` are driven by `AppState.shouldMaskFinancialValues` (true for both `.masked` Privacy Mask and `.locked` App Lock), so the two App-Group files redact in lockstep with the Spotlight index.
- **State-change rewrite via `writeGlanceSnapshot`.** The previous code re-emitted only the `FinanceSnapshot` on toggle/lock. Routing through `writeGlanceSnapshot` redacts *both* files together — closing the gap where a real glance file lingered while masked.
- **Widget self-masks on either flag.** `GlanceTimelineProvider` now masks on `FinanceSnapshot.isMasked || GlanceSnapshot.isRedacted`, so it stays masked even if the sibling `FinanceSnapshot` is missing or stale; a redacted glance file already holds no figures regardless.

## Migration notes

- **Backward-compatible decode.** `isRedacted` is decoded with `decodeIfPresent(...) ?? false` via an explicit `init(from:)`, so a `glance-snapshot.json` written by a pre-AND-517 build (no `isRedacted` key) decodes as a normal, non-redacted snapshot. Upgrades never mistake a real snapshot for a masked one. New field defaults to `false` in the memberwise init, so all existing call sites are unaffected.
- No schema migration, no data-dir changes, no server changes.

## Testing notes

New tests in `Tests/PlaidBarCoreTests/GlanceSnapshotTests.swift`:
- `Masked build zeroes every financial value`
- `Unmasked build is unchanged and carries real values`
- `redacted() clears figures but keeps metadata`
- `Redacted snapshot JSON on disk carries no real figures`
- `Enabling mask redacts the persisted snapshot (transition)`
- `Legacy snapshot without isRedacted decodes as not redacted`

CI gate (strict concurrency) + full suite, sandbox off:

```
Build complete! (46.41s)

$ swift test
✔ Test run with 1135 tests passed after 0.713 seconds.
```

GlanceSnapshot suite in isolation:

```
$ swift test --filter "GlanceSnapshotTests"
✔ Test "redacted() clears figures but keeps metadata" passed
✔ Test "Legacy snapshot without isRedacted decodes as not redacted" passed
✔ Test "Redacted snapshot JSON on disk carries no real figures" passed
✔ Test "Unmasked build is unchanged and carries real values" passed
✔ Test "Masked build zeroes every financial value" passed
✔ Test "Enabling mask redacts the persisted snapshot (transition)" passed
✔ Suite "Glance snapshot" passed
```

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -397,10 +397,14 @@ final class AppState {
     func togglePrivacyMask() {
         guard !isContentLocked else { return }
         appLockPreferences.privacyMaskEnabled.toggle()
-        // Re-emit the App Intents snapshot so the masked/unmasked state takes
+        // Re-emit both App Group snapshots so the masked/unmasked state takes
         // effect on disk immediately: enabling the mask overwrites any prior real
-        // figures with a value-free snapshot; disabling restores the real values.
-        writeFinanceSnapshot()
+        // figures with value-free snapshots; disabling restores the real values.
+        // `writeGlanceSnapshot` also rewrites the App Intents `FinanceSnapshot`
+        // (via `writeFinanceSnapshot`), so the widget glance snapshot and the
+        // intents snapshot are redacted together — no stale real glance file is
+        // left on disk while the app is masked (AND-517).
+        writeGlanceSnapshot()
     }
 
     /// True only in full App Lock (`.locked`) — distinct from
@@ -690,10 +694,13 @@ final class AppState {
         // stale message from a prior cancelled/failed unlock attempt.
         if isAppLocked {
             lastUnlockMessage = nil
-            // Locking masks financial values, so immediately overwrite the App
-            // Intents snapshot with a value-free one — no figures leak past the
-            // lock via Spotlight / Siri / Shortcuts.
-            writeFinanceSnapshot()
+            // Locking masks financial values, so immediately overwrite both App
+            // Group snapshots with value-free ones — no figures leak past the lock
+            // via the widget / Control Center (glance snapshot) or Spotlight / Siri
+            // / Shortcuts (App Intents snapshot). `writeGlanceSnapshot` redacts the
+            // glance file and also re-emits the redacted `FinanceSnapshot` through
+            // `writeFinanceSnapshot` (AND-517).
+            writeGlanceSnapshot()
         }
     }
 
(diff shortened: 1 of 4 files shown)


ftchvs:
@codex please review — comprehensive review requested (privacy-sensitive).

PR #542, opened 2026-06-19
feat(intents): SetFocusFilterIntent auto privacy mask (AND-506)

ftchvs:
## Summary

Adds **Focus-aware Privacy** (AND-506): attaching VaultPeek to a Focus (Work, a shared Focus, Do Not Disturb…) in **System Settings → Focus → Focus Filters** now automatically enables **Privacy Mask** while that Focus is active, and restores the user's prior mask state when the Focus ends.

The new `FocusPrivacyFilterIntent` is a thin `SetFocusFilterIntent` shell. It does **not** add new `AppState` machinery — it routes the desired mask state through the *existing* privacy-mask command channel that the Control Center "Privacy Mask" toggle already uses:

```
FocusPrivacyFilterIntent.perform()
  → PrivacyMaskControlCommandReader.write(privacy-mask-command.json)
  → AppState.applyPendingPrivacyMaskControlCommand()  (consumed on next activation)
  → appLockPreferences.privacyMaskEnabled
```

`perform()` fires on both Focus activation (configured parameters) and deactivation (default parameters); `SetFocusFilterIntent.current` (resolves while active, throws once inactive) is the activate/deactivate signal. The pre-Focus mask state is remembered in App Group `UserDefaults` (a single bool — never a balance) so deactivation restores exactly what the user had.

The intent's conformance to `SetFocusFilterIntent` in the app target is what surfaces it under Focus Filters — no extra registration glue is required.

### Files
- `Sources/PlaidBar/Intents/FocusPrivacyFilterIntent.swift` — new `SetFocusFilterIntent` conformer + App-Group-backed `FocusPrivacyMaskMemory`.
- `Sources/PlaidBar/Controls/PrivacyMaskControlCommandReader.swift` — adds an app-side `write(_:)`, the twin of the extension's `WidgetControlCommandStore.savePrivacyCommand`, so the app can feed the command file it already consumes.
- `Sources/PlaidBarCore/Utilities/FocusPrivacyMaskDecision.swift` — pure, testable decision (focus active/inactive → desired mask + prior-state bookkeeping).
- `Tests/PlaidBarCoreTests/FocusPrivacyMaskDecisionTests.swift` — unit tests for the pure decision.

## Linear (AND-506)

Implements **AND-506 — Focus-aware privacy: auto-enable Privacy Mask when a Focus activates.** The mask plumbing already existed (the Control Center command path + `appLockPreferences.privacyMaskEnabled`); this wires a Focus filter into it.

## Testing notes

- New suite `Focus-aware Privacy Mask decision (AND-506)` — 7 tests, all pass:
  - `activationFromVisibleEnablesMaskAndRemembers`
  - `activationWhileAlreadyMaskedRemembersTrue`
  - `repeatedActivationKeepsOriginalPrior`
  - `deactivationRestoresVisible`
  - `deactivationRestoresMaskedPreference`
  - `deactivationWithoutMemoryReveals`
  - `inertFilterIsNoOp`
- Strict-concurrency CI gate **passes**: `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → `Build complete!`
- Full suite **passes**: `swift test` → `Test run with 1197 tests passed`. (The `✘` lines in output are pre-existing *skips* — "macOS Keychain accepts test writes" — in the headless env, not failures.)

Widget/Control/Intent types aren't headlessly unit-testable, so the only real logic (the mask decision, incl. respecting + restoring prior user state) lives in `PlaidBarCore` and is tested there.

## Architectural decisions

- **Reuse, not new state.** Drives the same `applyPendingPrivacyMaskControlCommand` apply path as the Control Center toggle, keeping a single privacy-mask source of truth. The app target can't reference the extension's `WidgetControlCommandStore`, so I added the symmetric `write(_:)` to the app's own copy of the command contract (`PrivacyMaskControlCommandReader`), byte-identical to the extension writer.
- **Restore prior state.** Because the Focus callback is stateless, the pre-Focus mask is captured on first activation and restored on deactivation. The decision only captures the prior value when none is remembered yet, so repeated activations don't overwrite it with the now-masked `true`.
- **Inert filter is a true no-op.** If the user unchecks "Hide balances," the filter never moves the mask in either direction and clears any stray remembered value.
- **Silent.** `openAppWhenRun` stays false (inherited default) — masking shouldn't yank the app forward; the app applies the pending command on its next natural activation, mirroring the Control Center toggle. Never conveys state by color alone (figures are dotted out; SF Symbols differ by shape).

## Migration notes

None. No persistence schema, migration, or server change. New bookkeeping is a single bool in the existing App Group `UserDefaults` suite (`group.com.ftchvs.PlaidBar`), absent until a Focus filter first runs and self-clearing on deactivation. The Focus filter appears in System Settings only after the user installs a build whose app metadata includes the new intent.

### DIFF
--- a/Sources/PlaidBar/Controls/PrivacyMaskControlCommandReader.swift
+++ b/Sources/PlaidBar/Controls/PrivacyMaskControlCommandReader.swift
@@ -37,6 +37,13 @@ enum PrivacyMaskControlCommandReader {
         return decoder
     }
 
+    private static var encoder: JSONEncoder {
+        let encoder = JSONEncoder()
+        encoder.dateEncodingStrategy = .iso8601
+        encoder.outputFormatting = [.sortedKeys]
+        return encoder
+    }
+
     /// Resolves the shared command directory, reusing the glance store's
     /// container-or-local-fallback logic so the reader and the control agree on a
     /// single App Group location.
@@ -59,6 +66,30 @@ enum PrivacyMaskControlCommandReader {
         let data = try Data(contentsOf: url)
         try fileManager.removeItem(at: url)
         return try decoder.decode(PrivacyMaskControlCommand.self, from: data)
+    }
+
+    /// Writes a pending privacy command into the shared container so it is applied
+    /// on the next `applyPendingPrivacyMaskControlCommand()` activation — the same
+    /// channel the Control Center toggle uses (`WidgetControlCommandStore`), but
+    /// from the **app** side. The Focus filter intent (AND-506) uses this so it
+    /// drives the existing apply path instead of new state machinery. Writes
+    /// atomically with owner-only permissions, byte-identical to the extension's
+    /// writer so the contract stays in sync.
+    static func write(
+        _ command: PrivacyMaskControlCommand,
+        directory: URL = directory(),
+        fileManager: FileManager = .default
+    ) throws {
+        let url = commandURL(directory: directory)
+        try fileManager.createDirectory(
+            at: url.deletingLastPathComponent(),
+            withIntermediateDirectories: true
+        )
+        let data = try encoder.encode(command)
+        try data.write(to: url, options: [.atomic])
+        #if os(macOS)
+        try? fileManager.setAttributes([.posixPermissions: 0o600], ofItemAtPath: url.path)
+        #endif
     }
 }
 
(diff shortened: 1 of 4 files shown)


ftchvs:
@codex please review — comprehensive review requested (Apple platform).

PR #551, opened 2026-06-19
feat(ui): inline "always categorize…" rule prompt at correction (AND-531)

ftchvs:
## Summary

When a user assigns a category to a transaction in the Review Inbox (the recategorize action), VaultPeek now surfaces an inline, dismissible prompt offering to turn that one-off correction into a durable rule: **"Always categorize \<merchant\> as \<category\>?"**. This is the Copilot-style "correct once, remember forever" affordance at the exact moment the correction happens — instead of the user having to dig into the row's Rule menu separately.

Accepting routes through the **existing** `AppState.createRule(from:category:)` path (no new rule plumbing), so the rule then flows into the override-aware spend math already on `main` like any other rule. The offer is deliberately **non-nagging**:

- one prompt at a time; a newer correction replaces an older offer
- dismissible (a small ✕), and dismissing leaves the applied correction in place
- suppressed automatically when the merchant is blank or an identical (merchant + category) rule already exists
- cleared when the inbox drains
- hidden entirely under Privacy Mask / App Lock (it names the merchant)

## Linear (AND-531)

AND-531 — *inline rule prompt at the moment a user recategorizes a review-inbox item* (Epic AND-519, Review Inbox parity). Related to the completed AND-398 and the override-aware spend-math foundation (AND-518) the created rule feeds.

## Spec alignment (§3 / §5; Option A)

- **§3 Target experience** lists "inline rule prompt" as a Review Inbox parity item — this delivers it.
- **§5** maps AND-531 under Epic AND-519. Option A (display-only rollups over the existing taxonomy, no schema migration): this change adds **no** schema, no new `AppState` state, and reuses `TransactionRule` / the existing add-rule path verbatim. The rule it creates is exactly the one the row's "Always categorize as" menu already produces, so it feeds the persisted-only override-aware resolver with no new source of truth.

## Testing notes

- New pure logic extracted to `PlaidBarCore/Models/InlineCategoryRulePrompt.swift`, unit-tested in `InlineCategoryRulePromptTests` (12 tests): the built rule matches the merchant and carries the category, the matcher mirrors `createRule` (trimmed effective merchant name), an identical merchant+category rule dedupes case-insensitively, a different category or a transfer-only rule still offers a prompt, and a blank merchant yields no rule / no prompt.
- Strict-concurrency CI gate green: `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors`.
- Full suite green: `swift test` — 1566 tests pass.

## Architectural decisions

- **Pure-model split:** the "should we prompt / what rule would Accept create / is this a duplicate" decisions live in a `Sendable` PlaidBarCore type, not the SwiftUI view (views aren't headlessly testable). The view is a thin renderer + wiring.
- **Reuse, don't fork:** Accept calls `appState.createRule(from:category:)` — the same path the Rule menu and undo stack already use — so the inline prompt can never diverge from how rules are otherwise created (matcher, merchant label, transfer/exclusion context, undo, persistence).
- **Captured row on accept:** the recategorization marks the row reviewed (it leaves the live snapshot), so the offered `TransactionReviewItem` is captured alongside the prompt to drive `createRule` even after the row is gone.
- **Accessibility:** the offer and its actions are always text + glyph labelled (`wand.and.stars`, "Create rule", "Dismiss"); meaning never rides on color alone. Creating a rule posts a VoiceOver announcement.

## Migration notes

None. No schema, no DTO, no `SpendingCategory` rawValue, and no new `AppState` state changes. Rules continue to live in the existing app-local `transaction-rules.json` (local-first), written through the existing persistence path.

### DIFF
--- a/Sources/PlaidBar/Views/ReviewInboxView.swift
+++ b/Sources/PlaidBar/Views/ReviewInboxView.swift
@@ -22,6 +22,18 @@ struct ReviewInboxView: View {
     /// scheduled for — a newer action (which bumps the token) is never wiped by
     /// an older timer.
     @State private var confirmationGeneration = 0
+    /// The inline "always categorize <merchant> as <category>?" offer staged the
+    /// moment the user recategorizes a row (AND-531). One at a time, dismissible;
+    /// a newer correction replaces an older offer. Accepting routes through the
+    /// existing `AppState.createRule` path. Nil when no offer is pending. The pure
+    /// `InlineCategoryRulePrompt` decides whether an offer is even worth showing
+    /// (suppressed for a blank merchant or a duplicate rule), so this stays
+    /// non-nagging.
+    @State private var rulePrompt: InlineCategoryRulePrompt?
+    /// The row the staged `rulePrompt` was offered for, captured so an Accept can
+    /// route through the existing `AppState.createRule(from:category:)` path even
+    /// after the recategorization removed the row from the live snapshot.
+    @State private var rulePromptItem: TransactionReviewItem?
 
     private var snapshot: TransactionReviewInboxSnapshot {
         appState.transactionReviewInboxSnapshot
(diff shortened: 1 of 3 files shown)


ftchvs:

PR #556, opened 2026-06-19
fix(security): harden Plaid token Keychain accessibility to ThisDeviceOnly (AND-572)

ftchvs:
## Summary

Hardens the Plaid access-token Keychain storage so tokens are pinned to the user's machine and never reach iCloud Keychain.

- **New `KeychainAccessPolicy` (PlaidBarCore)** — a pure, framework-free type that centralizes the data-protection decision: accessibility class `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` and `synchronizable = false`. Its raw values mirror Apple's documented `kSecAttrAccessible` string constants so the choice is unit-testable in CI with no live Keychain.
- **`PlaidTokenVault` re-asserts the hardened protection on every write.** Both `SecItemAdd` and `SecItemUpdate` now set the device-only accessibility class **and** `kSecAttrSynchronizable = false`, sourced from the policy. Items written by older builds are upgraded in place on the next save.
- **SECURITY.md** documents the hardened storage policy and the rationale for the class choice.

## Linear (AND-572)

Implements AND-572: audit then harden the Plaid access-token Keychain item's accessibility / data-protection.

Audit finding: the item was already using `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly` on add (introduced in #488), but the choice was an inline literal, was **not** re-asserted with an explicit `kSecAttrSynchronizable = false`, and had no CI-runnable test guarding it. This PR closes those gaps.

## Testing notes

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — green.
- `swift test` — green (1597 tests).
- New `KeychainAccessPolicyTests` (6 tests, PlaidBarCore) assert the chosen class is `afterFirstUnlockThisDeviceOnly` (raw `"cku"`), that it is device-only, that `synchronizable == false`, and that a `ThisDeviceOnly` class is never paired with `synchronizable = true`. These run in CI without a Keychain.
- The live Keychain round-trip tests (`TokenStorageSafetyTests`) stay environment-gated behind `PLAIDBAR_TEST_KEYCHAIN` and continue to pass when enabled; they exercise the real add/resolve/delete path that now writes the hardened attributes.

## Architectural decisions

- **`AfterFirstUnlockThisDeviceOnly`, not `WhenUnlockedThisDeviceOnly`.** `PlaidBarServer` is a headless background daemon that must resolve the token to refresh data even while the screen is locked (after the first post-boot unlock). `WhenUnlockedThisDeviceOnly` would fail those reads.
- **`ThisDeviceOnly` + `synchronizable = false`.** Apple forbids pairing a `ThisDeviceOnly` accessibility class with `kSecAttrSynchronizable = true`; setting it explicitly to `false` documents intent and keeps tokens off iCloud Keychain and paired devices. Lookups already match only non-synchronizable items by default, so reads/writes are consistent.
- **Decision lives in `PlaidBarCore`, mapping to CFStrings lives in the server.** Keeps the policy `Sendable`, CI-testable, and free of a Security-framework dependency, while `PlaidTokenVault` owns the concrete `kSecAttr*` constants.

## Migration notes

No schema or data migration. Existing Keychain items are upgraded in place: the next time an item's token is saved, `SecItemUpdate` re-asserts the device-only accessibility class and `synchronizable = false`. The SQLite `keychain:<item_id>` reference scheme is unchanged, and the `PlaidBar.PlaidAccessToken` service name is preserved, so existing references continue to resolve.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -53,6 +53,19 @@ Use GitHub private vulnerability reporting if available, or contact the reposito
   stores only a `keychain:<item_id>` reference. Build/test environments without
   Keychain support may fall back to local SQLite token storage, so release
   claims must distinguish runtime Keychain behavior from fallback builds.
+- Keychain access-token items are hardened to stay on the user's machine. Each
+  item is written with the `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`
+  data-protection class and `kSecAttrSynchronizable = false`, so the token is
+  **never** copied to iCloud Keychain or any paired device, yet remains readable
+  by the headless companion server after the first post-boot unlock (including
+  while the screen is locked, which is required for background refresh). The
+  weaker `WhenUnlockedThisDeviceOnly` class is intentionally avoided because it
+  would block background reads on a locked screen, and the default
+  iCloud-syncable classes (`WhenUnlocked`, `AfterFirstUnlock`) are avoided
+  because they would let tokens leave the device. The protection is re-asserted
+  on every write, so items created by older builds are upgraded in place. The
+  accessibility-class decision is centralized in `KeychainAccessPolicy`
+  (`PlaidBarCore`) and unit-tested in CI without requiring a live Keychain.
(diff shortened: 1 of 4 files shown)


ftchvs:

PR #558, opened 2026-06-19
fix(a11y): refresh audio-graph descriptor on Privacy Mask toggle to stop stale-amount leak (AND-569)

ftchvs:
## Summary

Fixes a **P1 Privacy Mask staleness leak** in the chart audio graphs — a follow-up to AND-569 (the `AXChartDescriptor` audio graphs merged in #557).

AND-569 masked the audio graph at **build time** (`makeChartDescriptor`), but `updateChartDescriptor(_:)` was a no-op. Per Apple's docs (verified via Context7), SwiftUI calls `makeChartDescriptor()` **once per view identity** and `updateChartDescriptor(_:)` whenever inputs/Environment change. So when inputs changed *without* a new identity — most importantly when **Privacy Mask / App Lock toggles ON while a chart is on screen**, or when data refreshes — the already-built descriptor stayed **stale**. A descriptor built while unmasked kept exposing exact amounts through VoiceOver's audio graph (point labels, summary, and the y-axis value descriptions) even after masking engaged.

## The fix

- Implement `updateChartDescriptor(_:)` so it re-derives `title`, `summary`, `xAxis`, `yAxis`, and `series` from the **current** `ChartAudioGraph.Descriptor` value model — which carries the current `isPrivacyMasked` state and current data — and assigns them onto the passed-in `AXChartDescriptor` in place (the documented lifecycle hook for input changes).
- Extract the AX translation into **one set of private builders** (`makeXAxis` / `makeYAxis` / `makeSeries`) that both `make` and `update` call, so build-time and update-time masking go through a **single path** and cannot diverge.
- Add `ChartAudioGraph.Descriptor.axSpokenStrings(currencyCode:)` to Core: the full set of strings the AX descriptor will speak (summary + every point label + the sampled y-axis value descriptions). This is the single masking surface the representable consumes from both `make` and `update`.

## Why the test lives in Core

`PlaidBar` is an `executableTarget`, so the SwiftUI representable can't be a SwiftPM test dependency (this is why AND-569 kept all logic in Core with a thin translator). The masking invariant is therefore asserted at the Core level on `axSpokenStrings()` — the exact surface both code paths feed into the `AXChartDescriptor`.

## Tests

New (all green):
- `masked donut's full spoken surface (summary + labels + value axis) leaks no amount`
- `unmasked donut's spoken surface still carries amounts (masking is conditional)`
- `toggling Privacy Mask on a live chart re-masks the whole spoken surface` — models the exact leak: unmasked descriptor on screen → mask engages → masked descriptor adopted, asserting no figure from the unmasked surface survives anywhere
- `masked heatmap's full spoken surface leaks no amount`

All 16 pre-existing `ChartAudioGraphTests` remain green.

## Verification

- `swift test` — **1617 passed**
- `swift test --filter ChartAudioGraphTests` — 20 passed

## Scope

Only the three owned files: `ChartAudioGraphDescriptor.swift`, `ChartAudioGraph.swift` (Core), and `ChartAudioGraphTests.swift`. No changes to `MainPopover`, `AppState`, or `ReviewInboxView`.

### DIFF
--- a/Sources/PlaidBar/Views/Charts/ChartAudioGraphDescriptor.swift
+++ b/Sources/PlaidBar/Views/Charts/ChartAudioGraphDescriptor.swift
@@ -20,9 +20,39 @@ struct ChartAudioGraphRepresentable: AXChartDescriptorRepresentable {
     let descriptor: ChartAudioGraph.Descriptor
 
     func makeChartDescriptor() -> AXChartDescriptor {
-        let model = descriptor
+        // Built once per view identity. `update` re-derives from the same model.
+        AXChartDescriptor(
+            title: descriptor.title,
+            summary: descriptor.summary,
+            xAxis: makeXAxis(from: descriptor),
+            yAxis: makeYAxis(from: descriptor),
+            additionalAxes: [],
+            series: [makeSeries(from: descriptor)]
+        )
+    }
 
-        let xAxis = AXNumericDataAxisDescriptor(
+    func updateChartDescriptor(_ axDescriptor: AXChartDescriptor) {
+        // CRITICAL (AND-569 follow-up): `make` runs only once per view identity,
+        // so when inputs change WITHOUT a new identity — most importantly when
+        // Privacy Mask / App Lock toggles ON while a chart is on screen, or when
+        // data refreshes — VoiceOver keeps reading the descriptor built here. If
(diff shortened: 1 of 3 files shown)


ftchvs:
@codex please review — privacy-sensitive (audio-graph staleness).

PR #559, opened 2026-06-19
feat(a11y): in-app text size control overriding macOS Dynamic Type (AND-570)

ftchvs:
## Summary

macOS does **not** honor the system Dynamic Type setting for third-party apps, so VaultPeek users had no way to enlarge the app's text. The typography already *responds* to a `DynamicTypeSize` environment value (`@ScaledMetric`, shipped in AND-515) — this PR adds the missing user control that *sets* that value, applied once at every scene/window root so every surface scales together.

- **Core model** — new `TextSizePreference` (`default` / `large` / `xLarge` / `accessibility`) with a SwiftUI-free `ForcedDynamicTypeSize` mapping, stable persisted `rawValue`, display `title`/`detail`, and a `--text-size` CLI-override resolver. Mirrors the existing `AppearancePreferences` pattern (`storageKey` / `defaultValue` / `Identifiable`). Cases are clamped to the `.large … .accessibility1` band that the hero/display typography already supports.
- **App-wide application** — a shared `.appliesAppTextSize()` modifier reads `@AppStorage(TextSizePreference.storageKey)` and sets `.dynamicTypeSize` at the popover root, the Settings root, and the three detached AppKit-hosted window roots (floating dashboard, Category Dashboard, Review Table). A `DynamicTypeSize(ForcedDynamicTypeSize)` initializer is the single SwiftUI/Core meeting point (parity with `ForcedColorScheme` → `ColorScheme`).
- **Settings** — a dedicated "Text Size" section in Appearance with a labeled, VoiceOver-hinted `Picker` and a privacy-safe live preview (synthetic `$12,345.67` net-worth row, never real data) that scales as you change the picker.

## Linear

[AND-570](https://linear.app/and/issue/AND-570) — In-app text size / display scaling control. Parent epic AND-562 (accessibility); related AND-365 (appearance/display comfort) and AND-515 (`@ScaledMetric` typography).

## Testing notes

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — green (CI gate).
- `swift test` — **1624 tests pass**, including 11 new `TextSizePreferenceTests`: case→`DynamicTypeSize` mapping, strict monotonicity, default (`.default` → `.large`), `--text-size` CLI-override precedence, `rawValue` round-trip + unknown-value → `nil`, `appearance.*` storage-key isolation, `Identifiable`/case-count, and non-empty labels.

## Architectural decisions

- **Pure logic in Core, SwiftUI bridge in app.** `TextSizePreference` keeps the case→size decision testable without importing SwiftUI; `ForcedDynamicTypeSize` (raw values match the SwiftUI case names) crosses into `SwiftUI.DynamicTypeSize` in one app-layer `init`. This is the same split AND-365 uses for `ForcedColorScheme`.
- **Apply at roots, not per-view.** Setting `.dynamicTypeSize` once at each scene/window root means every `@ScaledMetric`/`Text` below it scales — no view edits required, and the three detached windows stay in sync because the modifier reads `@AppStorage` (live updates).
- **Menu-bar label intentionally unscaled.** The status-item text lives in the system menu bar; enlarging it would clip/misalign the item, so the label root does not get `.appliesAppTextSize()`.
- **Case set clamped to the supported band.** Hero/display surfaces already cap at `.accessibility3`; keeping the preference at `.accessibility1` max guarantees no surface is pushed past a step it has decided to clamp.

## Migration notes

None. New persisted key `appearance.textSize` defaults to `.default` (raw value `"default"`), which maps to `DynamicTypeSize.large` — the current/standard rendering — so existing installs see no visual change until a user opts in. No data migration; unknown stored values fall back to the default.

### DIFF
--- a/Sources/PlaidBar/App/CategoryDashboardWindowController.swift
+++ b/Sources/PlaidBar/App/CategoryDashboardWindowController.swift
@@ -177,6 +177,10 @@ final class CategoryDashboardWindowController: NSObject, NSWindowDelegate {
             CategoryDashboardWindow()
                 .environment(appState)
                 .forcedCategoryDashboardColorScheme(forcedColorScheme)
+                // Apply the in-app text-size preference here too so the detached
+                // Category Dashboard scales with the same control as the popover
+                // (AND-570). Reads @AppStorage, so it tracks live changes.
+                .appliesAppTextSize()
                 .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .topLeading)
         )
     }
--- a/Sources/PlaidBar/App/DetachedDashboardWindowController.swift
+++ b/Sources/PlaidBar/App/DetachedDashboardWindowController.swift
@@ -417,6 +417,9 @@ final class DetachedDashboardWindowController: NSObject, NSWindowDelegate {
                 // window's live (NSApp-driven) appearance and follows Light/Dark
                 // changes made while the window is open.
                 .forcedDetachedColorScheme(forcedColorScheme)
+                // Apply the in-app text-size preference so the floating desktop
+                // dashboard scales with the same control as the popover (AND-570).
+                .appliesAppTextSize()
                 // The panel owns its own width via resize; let the dashboard fill
                 // it rather than imposing the popover's screen-anchored width math.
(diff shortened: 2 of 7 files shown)


ftchvs:

PR #564, opened 2026-06-19
feat(ai): detect Foundation Models availability + prefer it in the local-AI tier resolver (AND-563)

ftchvs:
## Summary

Foundation-only slice of **AND-563**: detect Apple **Foundation Models** (Apple Intelligence) and slot it at the **top** of the on-device AI tier order when available. This is detection + tier representation + preference ordering + graceful fallback **only** — it does not build FM-backed insight generation (AND-564) or FM categorization (AND-565), both of which are held.

What landed:

- **Pure tier resolver in `PlaidBarCore`** — new `LocalAITierResolver.resolvePreferredTier(facts:)` plus `LocalAIRuntimeTier` (ordered enum), `LocalAITierFacts`, and `LocalAIFoundationModelsTierState`. The resolver returns `.foundationModels` **only** when the FM probe reports `.available`; for every other FM state it returns **exactly** the pre-FM order: `ollama → naturalLanguage → heuristic`. That equivalence is the regression guard.
- **App-side probe** — `FoundationModelsAvailabilityProbe`, gated behind `#available(macOS 26, *)` + `canImport(FoundationModels)`, maps `SystemLanguageModel.default.availability` (`.available` / `.deviceNotEligible` / `.appleIntelligenceNotEnabled` / `.modelNotReady` / other) into the OS-independent Core tier state. **Detection only** — it never constructs a model session or routes any transaction-derived prompt through Foundation Models. Returns `.unsupported` when FM can't be probed, leaving the legacy order untouched.
- **Surfacing** — `AppState` exposes `localAIPreferredTier`, `foundationModelsAvailability`, and `refreshFoundationModelsAvailability()`; the Settings "Local AI" section shows a new **Preferred tier** row next to the existing Availability/Runtime rows, with FM unavailability cause copy (device not eligible / Apple Intelligence off / model preparing).

Insight generation stays on the existing path; a follow-up wires generation.

## Linear (AND-563)

Implements AND-563 scoped to the foundation only (detection + tier representation + preference ordering + graceful fallback). AND-564 (FM-backed insight generation) and AND-565 (FM categorization) remain held and are intentionally out of scope here.

## Testing notes

- New `Tests/PlaidBarCoreTests/LocalAITierResolverTests.swift` (10 tests): FM preferred when available; FM outranks every lower rung; **FM-unavailable-for-any-reason resolves identically to the pre-FM order** (regression guard across all unavailable states); each fallback rung (Ollama → NL → heuristic); rank ordering; tier-state availability; display/cause metadata. The probe result is injected as a plain Core value, so the suite runs on any OS without Apple Intelligence.
- Strict-concurrency CI build clean: `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → Build complete.
- Full suite green: `swift test` → **1638 tests passed, 0 failures**.

## Architectural decisions

- **Pure decision in Core, probe in app.** The tier-ordering decision is a pure `Sendable` function in `PlaidBarCore`; the only code that touches the FoundationModels framework is the app-target probe, behind `#available` + `canImport`. This keeps the decision unit-testable on any machine and confines all OS-version/SDK gating to one file.
- **FM state modeled as an OS-independent Core enum.** `SystemLanguageModel.Availability` is mapped into `LocalAIFoundationModelsTierState` so neither Core nor its tests import FoundationModels; unknown/future unavailability reasons collapse to `.unavailableOther` (never silently "available").
- **Preference order, not a privacy boundary.** The new resolver is purely about *which on-device tier is the best available producer*. The opt-in / localhost privacy boundary stays in the existing `LocalAIRuntimeResolution` and is unchanged.
- **`#available(macOS 26, *)` kept even though the deployment target is macOS 26.** Makes the FM access path explicit and keeps the code defensive/reversible if the floor ever lowers; `canImport(FoundationModels)` covers SDKs/toolchains without the framework.

## Migration notes

None — additive, availability-gated. On older OS, machines without Apple Intelligence, or builds without the FoundationModels SDK, the probe returns `.unsupported` and the tier order is byte-for-byte what it was before this change. No data migration, no settings migration, no behavior change on non-capable machines.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -515,6 +515,12 @@ final class AppState {
     private var localAIModelNamePreference: String?
     private var localAIProbeAvailability: LocalAIAvailability?
     private var localAIProbeGeneration = 0
+    /// Detection-only Foundation Models (Apple Intelligence) tier state (AND-563).
+    /// Probed cheaply via `FoundationModelsAvailabilityProbe`; `.unsupported` on
+    /// any OS/build without Foundation Models, so the tier order is unchanged
+    /// there. No insight generation is routed through Foundation Models yet.
+    private let foundationModelsProbe = FoundationModelsAvailabilityProbe()
+    private var foundationModelsTierState: LocalAIFoundationModelsTierState = .unsupported
     private var isLoadingLocalAISettings = false
     private var isLoadingAppLockPreferences = false
     private var isUpgradingManagedServer = false
@@ -527,6 +533,10 @@ final class AppState {
         _ = try? LocalDataStore.migrateLegacyDefaultStorageIfNeeded()
         self.notificationService = notificationService ?? NotificationService.shared
         loadSettings()
+        // Detection-only: record whether Apple Foundation Models is available so
+        // the tier resolver can prefer it. No model session is created and no
+        // prompt is routed through it (AND-563). Cheap + synchronous.
+        foundationModelsTierState = foundationModelsProbe.currentState()
         // Engage launch App Lock synchronously before the first SwiftUI body is
         // ever evaluated. Deferring this to `loadInitialData()` leaks one frame
         // of cached/demo account names and activity because `.task` runs after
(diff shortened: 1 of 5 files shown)


ftchvs:

PR #565, opened 2026-06-19
feat(perf): energy-aware background refresh scheduling (AND-568)

ftchvs:
## Summary

Makes the resident menu-bar app's **automatic** background refresh respect the device's energy and thermal conditions. When `ProcessInfo.isLowPowerModeEnabled` is `true`, or `ProcessInfo.thermalState` is `.serious`/`.critical`, the periodic Plaid data fetch backs off:

- The cheap **connectivity re-probe, notification evaluation, and category-budget recompute keep running** every tick.
- Only the **network-and-CPU-heavy Plaid round-trip is deferred** while constrained.
- The background loop **sleeps longer** (base interval × `4`) so a throttled / battery-saving machine wakes the app less often, instead of spinning a tight wall-clock timer.
- **Manual refresh always runs** — energy state never suppresses a user-initiated refresh.
- Behavior is **unchanged when power and thermal are normal.**

The app subscribes to `NSProcessInfoPowerStateDidChange` and `ProcessInfo.thermalStateDidChangeNotification` and restarts the loop on a transition, so the cadence re-evaluates the moment conditions change (e.g. plug in → Low Power Mode off → loop drops back to the normal interval without waiting out a stale long sleep).

## Linear (AND-568)

Implements **AND-568** — energy-aware background refresh for the resident menu-bar app. Parent **AND-561**; related to **AND-422** (latency/cost) and **AND-425** (avoid full-cache rewrites), both Done.

## Testing notes

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — green.
- `swift test` — **1641 tests pass**, including **13 new** `EnergyAwareRefreshPolicyTests` covering: low-power defers, serious/critical thermal defers, normal proceeds, manual always allowed, not-yet-due never runs, the backoff multiplier (>1, lengthens the delay), and degenerate-base-interval clamping.
- The OS observers/timer live in the app; the pure decision (`shouldRunAutomaticRefresh` / `nextTickDelay`) lives in `PlaidBarCore` and is fully deterministic in CI (no live device thermal state).

## Architectural decisions

- **Pure decider in Core, observers in the app.** `EnergyAwareRefreshPolicy` (Sendable, framework-light) owns the gate + backoff math and a mirror of `ProcessInfo.ThermalState`, mirroring the existing `KeychainAccessPolicy` extraction pattern. `AppState` maps live `ProcessInfo` readings onto it. This keeps every branch unit-testable without a real device.
- **Gate placed inside `refreshDashboard(force:)`**, around the Plaid-fetch block, so the connectivity re-probe and budget recompute are never skipped — only the Plaid round-trip backs off. `force` (manual) bypasses the gate entirely.
- **Backoff multiplier co-located with the policy** rather than in `Constants.swift`, since it is policy-specific rather than a global tunable (the `refreshInterval` base remains the user-facing tunable).
- **`isolated deinit` (SE-0371)** removes the two block-based notification observers on the main actor, avoiding a leak without weakening actor isolation.

## Migration notes

None — additive, availability-gated. The power/thermal APIs are long-standing Foundation APIs (no macOS-26-only frameworks), so older OSes and non-Apple-Intelligence machines are unaffected and existing behavior is preserved when power/thermal are normal.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -503,6 +503,10 @@ final class AppState {
     private let notificationService: any NotificationServiceProtocol
     private var refreshTask: Task<Void, Never>?
     private var localAISummaryRefreshTask: Task<Void, Never>?
+    /// Observer tokens for the OS power-state / thermal-state change
+    /// notifications that drive energy-aware background refresh (AND-568).
+    /// Retained so they can be removed in `deinit`.
+    private var energyStateObservers: [NSObjectProtocol] = []
     private let glanceSnapshotWriteDebouncer = GlanceSnapshotWriteDebouncer()
     private var glanceSnapshotWriteGeneration = 0
     private var reviewUndoStack: [(metadata: [TransactionReviewMetadata], rules: [TransactionRule])] = []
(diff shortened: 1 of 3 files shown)


ftchvs:

PR #566, opened 2026-06-19
feat(ai): Foundation Models guided-generation categorization tier above NaturalLanguage (AND-565)

ftchvs:
## Summary

Adds an on-device **Apple Foundation Models (Apple Intelligence) guided-generation categorization tier** that slots **ABOVE NaturalLanguage** in the merchant-categorization suggestion order (parent AND-560; depends on AND-563, relates to AND-507).

When Apple Intelligence is available, a merchant/transaction is mapped to a `SpendingCategory` via **guided generation constrained to the 16-case enum**, so the model can only ever emit a valid case — there is no free-text path. The result is treated as a **suggestion with provenance** (exactly like the existing NL "Suggested" path): it is display-only and **never auto-applies** or bypasses the review/override flow. `EffectiveCategoryResolver` remains the source of truth for persisted categories.

When Foundation Models is **not** the resolved/available tier (older OS, no SDK, device ineligible, Apple Intelligence off, model not ready), behavior is **byte-identical to today's NaturalLanguage/heuristic path** — additive and fully reversible.

### What's in the diff
- **`PlaidBarCore` (pure, tested):**
  - `FMCategorizationTierDecision` — attempt FM only when the probe reports `.available`; otherwise reproduce the pre-FM path.
  - `FMSpendingCategoryMapper` — the single, unit-tested place that maps a constrained string output back to `SpendingCategory` (raw value or display name, case/whitespace-insensitive; `nil` on anything malformed → NL fallback).
  - `FMMerchantCategorizer` — composes the existing `NLMerchantCategorizer`; emits a `MerchantCategorySuggestion` carrying tier provenance. FM preferred when available, else NL. The live FM call is **injected** (`FMMerchantCategorizing` seam), so Core stays framework-free.
  - New `.appleFoundationModels` case on `LocalAICategoryResolutionSource`, surfaced under the existing **"Suggested"** badge.
- **App target:** `FoundationModelsMerchantCategorizer` — the only place touching the FoundationModels framework, behind `#if canImport(FoundationModels)` + `#available(macOS 26, *)`. A `@Generable` enum mirrors the 16 categories; the model receives **only a sanitized merchant string** (redaction-safe); greedy sampling for determinism; all errors swallowed to `nil` → NL fallback.
- **`AppState`:** an additive `merchantCategorizer` seam that wires the FM tier **only when available**; no existing review/budget/insight flow is changed.

## Linear (AND-565)

Implements **AND-565** — "Foundation Models guided-generation categorization tier above NaturalLanguage." Parent **AND-560**; depends on **AND-563** (FoundationModels availability detection + `LocalAITierResolver`, done); relates to **AND-507** (zero-setup NL categorization, done).

## Testing notes

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — **green** (the FM-gated code compiled against the real SDK on macOS 26.5.1).
- `swift test` — **1666 tests green**, including 15 new tests.
- New tests (`Tests/PlaidBarCoreTests/FMMerchantCategorizerTests.swift`), with the live FM call replaced by a deterministic injected stub so they run on any OS:
  - FM-available prefers the FM suggestion over NL.
  - An FM result maps to a valid `SpendingCategory` for **all 16 cases**.
  - FM receives **only** the redaction-safe merchant string — never the transaction/account id (asserted).
  - FM miss / unwired / unavailable falls back to NL.
  - **Regression guard:** for every non-`.available` FM state, the output is byte-identical to the bare `NLMerchantCategorizer` baseline, and the FM stub is **never consulted**.
  - The suggestion **never** becomes a persisted budget category (`EffectiveCategoryResolver.resolve` still returns `nil`).
  - `FMSpendingCategoryMapper` round-trips all raw values + display names and returns `nil` on malformed input.

## Architectural decisions

- **Privacy seam reuse.** FM goes through the same on-device, redaction-safe contract as the existing engines: only the merchant label the user already sees is sent to the model — no ids, tokens, amounts, dates, or Plaid payloads. The merchant string is sanitized (whitespace collapsed, length-capped) before it enters the prompt, and the constrained-enum output guarantees a valid category regardless of any injection attempt.
- **Suggestion, not application.** The tier produces a `MerchantCategorySuggestion` with provenance, mirroring the NL "Suggested" path. `EffectiveCategoryResolver` is left **untouched** and stays the source of truth for persisted categories; unapproved suggestions remain display-only (consistent with the Copilot review-inbox spec).
- **Pure decision in Core, gated call in app.** Tier-selection and the constrained-output→category mapping are pure and unit-tested in `PlaidBarCore`; only the gated, on-device call lives in the app target behind `#available`. The injected `FMMerchantCategorizing` seam keeps Core framework-free and the live call testable.
- **Reversible by construction.** The FM categorizer is wired only when the probe reports `.available`; otherwise the categorizer is the always-on NL path verbatim.

## Migration notes

None — additive, availability-gated. No schema, storage, or persisted-format change; no migration. On devices without Apple Intelligence the behavior is unchanged.

## On-device QA note (needs eyes-on a real Apple-Intelligence Mac)

The actual Foundation Models guided generation **cannot be exercised in CI** (no Apple Intelligence in the test runner) — tests cover the pure logic + the no-FM regression path with an injected stub. On a real Apple-Intelligence-eligible Mac (macOS 26, Apple Intelligence enabled), please verify:
- `FoundationModelsMerchantCategorizer.suggestCategory(merchant:)` returns a sensible `SpendingCategory` for representative merchants (e.g. "Blue Bottle" → Food & Drink, "Uber" → Transportation, "Netflix" → Entertainment).
- The constrained generation always yields one of the 16 cases (no free text / no throw leaking across the boundary).
- The model is only ever handed the merchant string (confirm no ids/amounts in any FoundationModels logging).
- Latency is acceptable for a per-merchant call and failures degrade quietly to the NL suggestion.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -1530,6 +1530,28 @@ final class AppState {
         #endif
     }
 
+    /// The merchant categorizer for the current device, with the Foundation
+    /// Models guided-generation tier (AND-565) slotted ABOVE NaturalLanguage when
+    /// Apple Intelligence is available, and falling back to NaturalLanguage
+    /// otherwise.
+    ///
+    /// Additive and reversible: the FM seam is wired ONLY when the probe reports
+    /// `.available`; in every other state this returns a categorizer whose
+    /// behavior is byte-identical to the always-on NaturalLanguage path, so the
+    /// no-FM device is unchanged. The categorizer produces *suggestions* with
+    /// provenance — it never auto-applies a category or bypasses the review /
+    /// override flow (`EffectiveCategoryResolver` stays the source of truth for
+    /// persisted categories).
+    var merchantCategorizer: FMMerchantCategorizer {
+        FMMerchantCategorizer(
+            foundationModelsState: foundationModelsTierState,
+            nlCategorizer: NLMerchantCategorizer(),
+            fmCategorizer: foundationModelsTierState.isAvailable
+                ? FoundationModelsMerchantCategorizer()
+                : nil
+        )
+    }
+
     /// Re-probe Apple Foundation Models availability (e.g. after the user enables
     /// Apple Intelligence in System Settings). Detection only.
     func refreshFoundationModelsAvailability() {
--- a/Sources/PlaidBar/Services/FoundationModelsMerchantCategorizer.swift
+++ b/Sources/PlaidBar/Services/FoundationModelsMerchantCategorizer.swift
@@ -1,0 +1,140 @@
+import Foundation
+import PlaidBarCore
+
+#if canImport(FoundationModels)
+import FoundationModels
+#endif
+
+/// On-device Apple Foundation Models (Apple Intelligence) merchant categorizer
+/// (AND-565). Maps a merchant string to a `SpendingCategory` via guided
+/// generation *constrained to the 16-case enum*, so the model can only ever emit
+/// a valid category — there is no free-text path.
+///
+/// This is the ONLY place in the categorization tier that touches the
+/// FoundationModels framework, and it is fully additive and reversible:
+///   - When the SDK is absent (`canImport` false) or the OS is older than
+///     macOS 26, the type compiles to a no-op that always returns `nil`, so the
+///     `FMMerchantCategorizer` tier falls back byte-identically to NaturalLanguage.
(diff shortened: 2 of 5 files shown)


ftchvs:
@codex please review — comprehensive review requested (on-device AI, privacy-sensitive).

PR #568, opened 2026-06-19
feat(perf): SwiftData disposable read-model cache for instant cold render (AND-566)

ftchvs:
## Summary

Adds a disposable SwiftData **read-model cache** so the dashboard popover can paint its first frame instantly on a cold start — and serve offline/widget reads — before the HTTP refresh returns. Today the app keeps decoded DTOs in memory plus a JSON balance ledger and UserDefaults caches; a cold popover open waits on the HTTP refresh. This change writes the last-known dashboard read-model to a dedicated, private, on-disk SwiftData store after each successful refresh and hydrates the first render from it on the next launch.

The change is **additive, disposable, and fallback-safe**: the existing in-memory + JSON/UserDefaults path stays the source of truth. The SwiftData store is a rebuildable cache, never authoritative. If SwiftData init/read/write fails or is disabled, the app behaves exactly as before (no regression), proven by a dedicated test.

## Linear (AND-566)

Implements **AND-566** (parent **AND-561**): a SwiftData disposable read-model cache for instant cold render + offline/widget reads. There was no SwiftData in the repo before this PR (`@Model`/`@Query` count was 0).

## Architectural decisions

- **Where the `@Model` lives.** The pure, `Sendable` read-model (`DashboardReadModel`) and the DTO↔read-model mapping (`DashboardReadModelMapper`) live in **`PlaidBarCore`** (widget-reachable, fully unit-testable). The SwiftData layer — `@Model CachedDashboardReadModel` + `@ModelActor ReadModelCacheStore` — lives in a **new app-only library target `PlaidBarCache`** (depends on `PlaidBarCore`). Rationale: `PlaidBarCore` is shared with the lean `PlaidBarServer` (Hummingbird), `PlaidBarCLI`, and `PlaidBarWidgetExtension`; importing SwiftData into Core would drag a macOS-26 GUI framework into those targets. A separate target keeps SwiftData out of them while remaining `@testable import`-able (the app target itself is an `@main` executable and cannot be imported by tests).
- **Disposable-store config.** A single trivial `@Model` (a JSON blob of the read-model + two scalar columns) — no relationships, no per-field schema to migrate. `ModelConfiguration("DashboardReadModelCache", schema:url:allowsSave:true, cloudKitDatabase: .none)` at `~/.vaultpeek/dashboard-read-model-cache-v1.store` (versioned filename). A `currentSchemaVersion` bump makes any older row a cache miss — the file can always be deleted and rebuilt from the authoritative refresh.
- **Isolation model.** `@ModelActor` makes the store a `Sendable` actor owning its non-`Sendable` `ModelContext`. `@Model` rows never escape the actor; only the `Sendable` `DashboardReadModel` value crosses the boundary, so the strict-concurrency gate stays clean. (`#Predicate` was avoided because the generated `KeyPath` capture is non-`Sendable` under Swift 6 mode; the cache holds a handful of rows, so it fetches-and-filters instead.)
- **Fallback contract.** Store open, read, and write are all `throws` and wrapped in `try?`/detached tasks at the AppState seams. Any failure (SwiftData unavailable, store unopenable, decode error, schema mismatch, key mismatch, empty row) yields "no hydration" → the app stays on its existing JSON/UserDefaults cold path. A feature flag (`readModelCacheEnabled`, default true) can disable it entirely.

## Testing notes

- `PlaidBarCoreTests/DashboardReadModelMapperTests` (8 tests): environment-scoped cache key, recent-transaction cap (newest-first), summary reuses the authoritative `MenuBarSummary` aggregators, hydration guards (key mismatch / older schema / empty all refuse), Codable round-trip.
- `PlaidBarCacheTests/ReadModelCacheStoreTests` (10 tests, `.serialized`): in-memory write→read round-trip, cold miss, upsert, per-environment isolation, stale-schema purge, clear/clearAll, **on-disk round-trip + `0o600` private-permission assertion**, reopen-same-directory (persistence survives restart), and an **empty/disabled-cache no-regression guard** modeling the disabled cold-start path.
- Strict-concurrency build green: `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors`.
- Full suite: **1696 tests pass**. The cache suite is `.serialized` because initializing multiple in-memory SwiftData containers in parallel crashes the SwiftData runtime.

## Migration notes

None. This is an additive cache. It introduces no migration to the authoritative data, no server change, and no schema migration on the real SQLite store. The cache's own store is disposable and versioned: a future shape change bumps `currentSchemaVersion` and the old file is treated as a miss / deleted.

## On-device QA note

Needs eyes-on: (1) the **cold-render speedup** — open the popover on a fresh launch for a previously-synced user and confirm accounts/recent transactions appear before the refresh spinner resolves; (2) **offline read** — launch with the local server down and confirm last-known data renders. Verify the store file exists at `~/.vaultpeek/dashboard-read-model-cache-v1.store` with `0o600` perms and that a local reset / removing the last institution clears it.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -33,6 +33,7 @@ let package = Package(
         .executable(name: "PlaidBarWidgetExtension", targets: ["PlaidBarWidgetExtension"]),
         .executable(name: "plaidbar-cli", targets: ["PlaidBarCLI"]),
         .library(name: "PlaidBarCore", targets: ["PlaidBarCore"]),
+        .library(name: "PlaidBarCache", targets: ["PlaidBarCache"]),
     ],
     dependencies: [
         .package(url: "https://github.com/sparkle-project/Sparkle", from: "2.8.0"),
@@ -55,6 +56,7 @@ let package = Package(
             name: "PlaidBar",
             dependencies: [
                 "PlaidBarCore",
+                "PlaidBarCache",
                 .product(name: "Sparkle", package: "Sparkle"),
                 .product(name: "MenuBarExtraAccess", package: "MenuBarExtraAccess"),
                 .product(name: "KeychainAccess", package: "KeychainAccess"),
@@ -114,6 +116,17 @@ let package = Package(
             swiftSettings: [.swiftLanguageMode(.v5)]
         ),
 
+        // MARK: - PlaidBarCache (Disposable SwiftData read-model cache, AND-566)
+        // App-only library so the SwiftData dependency never reaches the lean
+        // server / CLI / widget targets. Holds the @Model + @ModelActor store;
+        // the pure read-model + mapper live in PlaidBarCore.
+        .target(
+            name: "PlaidBarCache",
+            dependencies: ["PlaidBarCore"],
+            path: "Sources/PlaidBarCache",
+            swiftSettings: [.swiftLanguageMode(.v5)]
+        ),
+
         // MARK: - Tests
         .testTarget(
             name: "PlaidBarTests",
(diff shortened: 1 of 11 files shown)


ftchvs:
@codex please review — comprehensive review requested (architectural: new SwiftData layer).

PR #569, opened 2026-06-19
feat(perf): large-history list virtualization + paged fetch + SwiftData #Unique upsert (AND-567)

ftchvs:
## Summary

Handles large transaction histories efficiently — list virtualization, paged fetch, and a SwiftData `#Unique` upsert. Builds on AND-566's `PlaidBarCache`. Everything is **additive and fallback-safe**: the existing in-memory transaction list stays the source of truth and the default, and the list behaves exactly as today when the cache is disabled or unavailable (regression-guarded).

Three parts:

1. **`#Unique` upsert** — a new per-transaction `@Model` (`CachedTransaction`) in `PlaidBarCache`, keyed `@Attribute(.unique)` on `"<cacheKey>|<transactionId>"`, so re-syncing a transaction **updates in place** instead of duplicating. Populated off-main from the authoritative decoded DTOs through a `@ModelActor` (`TransactionCacheStore`). Disposable/versioned like AND-566's read-model cache.
2. **Paged fetch** — cached transactions are read in **pages** (newest-first, `offset`/`limit` from pure page-window math), not the whole history at once. Only the requested page's JSON payloads are decoded.
3. **List virtualization** — a `LazyVStack`-based `PagedTransactionListView` renders lazily and loads the next page on demand (infinite scroll). Wired into `AccountDetailFlyout`'s recent-activity list, which was a non-lazy `ForEach` capped at 6 rows.

**Fallback:** if `PlaidBarCache`/paging is unavailable or disabled (`readModelCacheEnabled == false`, demo mode, or the store fails to open), the list falls back to today's in-memory rendering, **unchanged**.

## Linear (AND-567)

AND-567 — the last item in epic AND-561. Builds directly on AND-566's `PlaidBarCache` (reuses its `cacheKey` scoping, lazy per-directory store-open pattern, and the disposable/versioned-store convention).

## Architectural decisions

- **`CachedTransaction` `@Model` + `#Unique` upsert.** A composite `@Attribute(.unique)` key `"<cacheKey>|<transactionId>"` makes a re-sync replace the row in place (no duplicates) while two Plaid environments can never collide on a bare transaction id. The authoritative `TransactionDTO` is stored as a JSON `payload` blob beside flat queryable scalars (`cacheKey`, `transactionId`, `sortDate`) — the same "flat primitives + one blob, disposable, bump-the-filename-not-a-migration" shape as `CachedDashboardReadModel`.
- **Paging design.** `TransactionPageWindow` (pure, in Core) owns the `offset`/`limit`/`hasMore` arithmetic; `PagedTransactionFeed` (pure) owns the infinite-scroll accumulator (append-page, dedup-by-id across pages, stop-at-end, grow/reset on a changed total). The store sorts + slices in Swift over the lightweight scalar columns and decodes **only the page's payloads**, so a multi-thousand-row history decodes one page at a time.
- **Strict-concurrency workaround.** Both the sort and the cacheKey scope are done in Swift rather than via `FetchDescriptor(sortBy:)` / `#Predicate`, because a SwiftData `SortDescriptor(\.keyPath)` (and `#Predicate`) captures a non-`Sendable` `KeyPath` — an error under the repo's `-strict-concurrency=complete -warnings-as-errors` gate. This is the same class of constraint AND-566 documented for `#Predicate`. `sortDate` is `YYYY-MM-DD`, so a plain descending string compare equals chronological newest-first.
- **Fallback contract.** `PagedTransactionSource` (`@Observable`) only flips to the paged path after a non-empty cache page actually loads; otherwise it renders `fallbackTransactions` (today's in-memory array). The which-rows decision is a pure `PagedTransactionRendering` helper so the fallback guarantee is pinned by a unit test, not just the view. The integration point is gated behind the existing `readModelCacheEnabled` kill switch (off / demo → byte-for-byte today's `ForEach`).
- **Reused seams.** Persist hooks into the single post-refresh `writeGlanceSnapshot()` seam (alongside `persistReadModelCache()`), using `replaceAll` so a refresh that removed transactions doesn't leave stale rows. Clear hooks into both local-reset paths next to `clearReadModelCache()`.

## Testing notes

Strict-concurrency build (`-strict-concurrency=complete -warnings-as-errors`) and the full suite are green: **1735 tests pass**, including 4 new AND-567 suites.

- `TransactionPageWindowTests` — offset/limit/hasMore, last partial page, past-end, empty history, invalid page size, `pageCount`, `next` walk, shrunk-total.
- `CachedTransactionUpsertTests` — all-new inserts, in-store updates, last-write-wins dedup, stable order, pending→posted same-id replace.
- `PagedTransactionFeedTests` — walks the whole history once, dedup across overlapping pages, grow/reset, and the **no-regression** tests proving the fallback path renders the in-memory array byte-for-byte.
- `TransactionCacheStoreTests` (`.serialized`, in-memory `ModelContainer`) — `#Unique` upsert round-trip, full paging walk (no overlap/gaps), `replaceAll` drops removed rows, cacheKey scoping, clear/empty.

## Migration notes

None — additive disposable cache. New SwiftData store file (`transaction-cache-v1.store`) in the local private data dir (`~/.vaultpeek/`, `0o700`/`0o600`); deletable at any time and rebuilt on the next refresh. No server, schema-migration, spend-math, or review/budget/dashboard changes.

## On-device QA note

Large-history scroll performance needs eyes-on: a real multi-thousand-transaction account on device, scrolling the virtualized list, to confirm page-on-demand stays smooth and memory stays bounded (the headless suite can't measure scroll/FPS).

### DIFF
--- a/Sources/PlaidBar/App/AppState+TransactionCache.swift
+++ b/Sources/PlaidBar/App/AppState+TransactionCache.swift
@@ -1,0 +1,102 @@
+import Foundation
+import OSLog
+import PlaidBarCache
+import PlaidBarCore
+
+/// AppState wiring for the disposable per-transaction SwiftData cache that backs
+/// the virtualized large-history list (AND-567).
+///
+/// Built on the AND-566 seams: it reuses ``readModelCacheKey()`` for environment
+/// scoping and the same lazy, per-directory store-open pattern. Two seams only:
+///   1. **Post-refresh persist** — `persistTransactionCache()` runs from the same
+///      `writeGlanceSnapshot()` seam as the read-model cache, mirroring the live
+///      transactions into the cache (upsert, no duplicates) so the paged list is
+///      ready on the next open.
+///   2. **List data source** — `makePagedTransactionSource(fallback:)` hands the
+///      virtualized list a ``PagedTransactionSource`` over the cache, falling back
+///      to the supplied in-memory array when the cache is unavailable/disabled.
+///
+/// Everything is best-effort and fallback-safe: any failure leaves the list on
+/// today's in-memory rendering (no regression).
+extension AppState {
+    nonisolated static let transactionCacheLogger = Logger(
+        subsystem: "com.ftchvs.PlaidBar",
+        category: "TransactionCache"
+    )
+
+    /// Lazily opens (or re-opens) the on-disk per-transaction store for the active
+    /// data directory. Returns `nil` — and stays disabled for this call — when the
+    /// cache is feature-disabled, in demo mode, or when SwiftData fails to open the
+    /// store. Reopens when the active storage directory changes so the store always
+    /// matches the environment whose key it is asked about.
+    func transactionCacheStoreIfAvailable() -> TransactionCacheStore? {
+        guard readModelCacheEnabled, !isDemoMode else { return nil }
+
+        let directory = activeStorageDirectoryURL
+        let directoryPath = directory.standardizedFileURL.path
+
+        if let existing = transactionCacheStore,
+           transactionCacheStoreDirectoryPath == directoryPath {
+            return existing
+        }
+
+        do {
+            let store = try TransactionCacheStore(onDiskIn: directory)
+            transactionCacheStore = store
+            transactionCacheStoreDirectoryPath = directoryPath
+            return store
+        } catch {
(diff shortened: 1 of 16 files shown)


ftchvs:
@codex please review — comprehensive review requested (perf + SwiftData paging).

PR #571, opened 2026-06-19
feat(ux): animated shared-element reflow on dashboard filter switch (matchedGeometryEffect) (AND-577)

ftchvs:
## Summary

Switching the dashboard filter segment (Cash / Credit / Savings / Debt / Status …) now **glides** the selection indicator between segments instead of cutting instantly. A single glass "pill" is carried from the old segment's frame to the new one with `matchedGeometryEffect`, so it slides — quiet and premium, not flashy.

`DashboardFilterBar` previously used `Picker(.segmented)` (an AppKit-bridged `NSSegmentedControl`). That bridge does **not** expose its selection indicator to SwiftUI's geometry system, so a matched-geometry glide is impossible on it. This PR replaces it with a custom segmented control that re-earns every affordance the bridge gave for free:

- per-segment VoiceOver `accessibilityLabel` / `accessibilityValue`
- the container rollup label ("Account filters. Cash: 3 of 8 accounts; …")
- ⌘1–⌘N keyboard shortcuts (now routed through the same glide path)
- the adjacent Status attention glyph (shape carries meaning; color is redundant)
- light/dark, and RTL (the `HStack` flips with layout direction and the pill glides along with it)
- `.isSelected` / `.isButton` traits on each segment

Pure policy lives in a new `Sendable` `PlaidBarCore` presenter, `DashboardFilterReflowModel` (stable matched-geometry id per filter kind, the Reduce-Motion gate, and the selection predicate), so it is unit-testable without SwiftUI and identical across the popover and the detached window.

## Linear (AND-577)

Implements **AND-577** — animated shared-element reflow when switching dashboard filter segments, via `matchedGeometryEffect`. Parent **AND-575**; related to **AND-295** (quiet filter bar, Done) and the established `MotionTokens` Reduce-Motion gating (**AND-339 / AND-378**).

## Testing notes

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — green
- `swift build -c release` — green (the CI gate)
- `swift test` — **1739 tests pass**, including 4 new `DashboardFilterReflowModelTests` (geometry-id distinctness/stability/namespacing, the Reduce-Motion glide gate, and the selection predicate)
- No redundant `try? #require` introduced (AND-574 strict test gate).

The pill glide is view-layer animation and not headlessly unit-testable; the pure decision logic it depends on is fully tested, and the rest needs eyes-on (see below).

## Architectural decisions

- **Custom control over native picker.** Chosen only because `NSSegmentedControl`'s indicator is opaque to `matchedGeometryEffect`. Every a11y/RTL/shortcut affordance the native control supplied is reconstructed explicitly so this is a no-regression swap on behavior.
- **Pure logic in Core.** `geometryID(for:)`, `shouldAnimateGlide(reduceMotion:)`, and `isSelected(_:selected:)` are `Sendable` and SwiftUI-free in `PlaidBarCore`, matching the repo convention and keeping the view thin.
- **One Reduce-Motion gate.** The glide flows through the existing `MotionTokens.standard` animation, applied only when `shouldAnimateGlide(reduceMotion:)` is true. **Additive + reversible:** with Reduce Motion on, the selection is assigned with no animation, so the pill snaps instantly — the exact end state the native control produced before this change. The selection value itself is never altered by the animation path.
- **Scoped `@Namespace`.** The pill's namespace is local to `DashboardFilterBar`, so the matched-geometry pair never collides with the popover's existing glass namespace (AND-511).

## On-device QA note

These two need eyes-on (they are not unit-testable):

1. **Glide:** Open the popover, click between Cash / Credit / Savings / Debt / Status (and try ⌘1–⌘N). The pill should *slide* smoothly between segments, settling without overshoot — not jump or cross-fade.
2. **Reduce-Motion fallback:** Enable System Settings → Accessibility → Display → **Reduce Motion**, then switch filters. The pill must **snap instantly** with no slide — identical to today's behavior. Also confirm VoiceOver still announces each segment's label/value and the container rollup, and that the Status attention glyph still appears when an item is degraded.

### DIFF
--- a/Sources/PlaidBar/Views/DashboardNavBand.swift
+++ b/Sources/PlaidBar/Views/DashboardNavBand.swift
@@ -20,17 +20,33 @@ extension DashboardAccountFilterKind {
 
 // MARK: - Filter Bar
 
-/// Native segmented control for the dashboard's primary filter. Native
-/// before novel: the system control brings focus rings, vibrancy, light
-/// mode, and RTL for free. ⌘1–7 shortcuts are preserved via hidden
-/// equivalents; the per-filter counts live in the rows themselves and in
-/// each segment's help/accessibility text rather than in segment labels.
+/// Segmented control for the dashboard's primary filter with an animated
+/// shared-element reflow (AND-577): a single glass selection pill glides between
+/// segments via `matchedGeometryEffect` rather than the selection cutting
+/// instantly from one to the next. Quiet and premium, not flashy.
+///
+/// This is a custom control rather than `Picker(.segmented)` because the
+/// AppKit-bridged `NSSegmentedControl` does not expose its selection indicator
+/// to SwiftUI's geometry system, so a matched-geometry glide is impossible on
+/// it. The custom control re-earns the affordances the bridge gave for free:
+/// per-segment VoiceOver label/value, a container rollup label, ⌘1–⌘N
+/// shortcuts, the Status attention glyph, light/dark, and RTL (the `HStack`
+/// flips with layout direction, and the pill glides along with it).
+///
+/// Additive + reversible: with Reduce Motion on, the pill snaps instantly with
+/// no geometry animation — the exact end state the native control produced
+/// before this change. The selection itself is never altered here.
 struct DashboardFilterBar: View {
     @Environment(AppState.self) private var appState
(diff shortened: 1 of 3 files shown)


ftchvs:
@codex please review.

PR #574, opened 2026-06-19
fix(core): clamp energy back-off product to a finite ceiling

ftchvs:
## Summary

Fixes a LOW-severity audit finding in `EnergyAwareRefreshPolicy.nextTickDelay`.

The constrained branch sanitized the **base** interval (`isFinite && > 0`) but multiplied the sanitized base by `constrainedBackoffMultiplier` **without sanitizing the product**. A finite-but-very-large base (e.g. near `Double.greatestFiniteMagnitude`) overflows to `.infinity` once multiplied, which silently violates the type's documented "the loop can never busy-spin or sleep forever" guarantee.

## Fix

After multiplying, guard the product: if it is not finite/positive, fall back to the (finite) sanitized base. Pure `PlaidBarCore` change, no behavior change for normal inputs.

```swift
let constrained = sanitizedBase * constrainedBackoffMultiplier
return (constrained.isFinite && constrained > 0) ? constrained : sanitizedBase
```

## Testing

- `swift test` — 1781 tests pass
- Added two Swift Testing cases:
  - **overflow**: huge finite base under constrained conditions returns a finite, positive delay (falls back to the base) instead of `.infinity`
  - **normal**: ordinary finite base still multiplies by the backoff multiplier

### DIFF
--- a/Sources/PlaidBarCore/Utilities/EnergyAwareRefreshPolicy.swift
+++ b/Sources/PlaidBarCore/Utilities/EnergyAwareRefreshPolicy.swift
@@ -126,6 +126,11 @@ public enum EnergyAwareRefreshPolicy: Sendable {
             ? baseInterval
             : PlaidBarConstants.backgroundRefreshInterval
         guard conditions.isConstrained else { return sanitizedBase }
-        return sanitizedBase * constrainedBackoffMultiplier
+        // Sanitize the *product*, not just the base: a finite-but-very-large base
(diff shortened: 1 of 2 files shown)


ftchvs:
@codex please review: this clamps the constrained back-off product in `EnergyAwareRefreshPolicy.nextTickDelay` so a finite-but-huge base interval can't overflow to .infinity (the base was sanitized but the post-multiply product wasn't). Confirm the fallback-to-base behavior and the new overflow/normal tests cover the guarantee.

PR #576, opened 2026-06-19
fix(cache): clear-wins gate stops cold-start resurrection of removed-institution balances

ftchvs:
## Audit finding
**[MEDIUM concurrency/privacy] readmodel-race** — Cold-start read-model cache can resurrect removed-institution balances.

A detached `save()` in `persistReadModelCache()` races the empty-accounts `clearReadModelCache()` issued from `writeGlanceSnapshot()`. When the user removes their last institution the cache is cleared, but an in-flight detached persist of the prior (non-empty) read-model can land **after** the clear on the shared store actor — so the next cold start paints stale balances for an account the user already removed.

## Fix
Confined to `Sources/PlaidBar/App/AppState+ReadModelCache.swift`.

Add `ReadModelCacheClearGate`, a `@MainActor` monotonic clear-epoch:
- `clearReadModelCache()` bumps the epoch **synchronously, before** its off-main `clearAll()` work — so any persist already scheduled in program order observes the clear.
- `persistReadModelCache()` captures the epoch when its write is scheduled and **re-checks it on the main actor immediately before committing**; if a clear has been requested since, the stale write is dropped instead of resurrecting rows.

Because the epoch bump and the capture/recheck both run synchronously on the main actor, a clear requested after a persist in program order is always observed by that persist. As defense in depth, even if a stale write slips through the recheck, the clear's `clearAll()` is serialized on the store actor **after** the bump and wipes it — so an empty-state clear can never be overwritten by a stale write.

No new `AppState` stored property was needed (extensions can't add one); the gate is a small main-actor singleton local to the wiring file.

## Tests
New `Tests/PlaidBarCacheTests/ReadModelCacheClearRaceTests.swift` (Swift Testing, serialized like the sibling store suite):
- documents the raw hazard (stale write after clear resurrects balances),
- pins the fix's guarantee (clear sequenced after a stale write -> cold-start `load` returns nil),
- fuzzes concurrent persists followed by a terminal clear -> always an empty cache.

## Verification
- `swift test` — 1782 tests pass (3 new).

## Note
The reusable epoch primitive + a direct unit test of it ideally live in `PlaidBarCore`/`PlaidBarCache` (the `PlaidBar` app target has no test target). That move requires editing `Package.swift` / `AppState.swift`, owned by other streams — deferred. The new store-level test fully covers the behavioral invariant the fix produces.

### DIFF
--- a/Sources/PlaidBar/App/AppState+ReadModelCache.swift
+++ b/Sources/PlaidBar/App/AppState+ReadModelCache.swift
@@ -17,6 +17,47 @@ import PlaidBarCore
 /// Everything is best-effort: any failure leaves the app on its existing
 /// JSON/UserDefaults cold path. The cache is never read as a source of truth and
 /// never blocks a render.
+///
+/// ## Persist/clear ordering (clear-wins)
+/// `persistReadModelCache()` schedules its write off the main actor so it never
+/// delays the render, and the empty-accounts path schedules a clear of the same
+/// store. Those two land on the store actor in an undefined order, so without a
+/// guard an in-flight persist of the *previous* (non-empty) read-model could
+/// commit *after* the clear that fires when the user removes their last
+/// institution — resurrecting removed-account balances on the next cold start.
+///
+/// `ReadModelCacheClearGate` closes that race: a monotonic clear epoch is bumped
+/// synchronously on the main actor the instant a clear is requested, *before* the
+/// clear's own off-main work begins. Every scheduled write captures the epoch it
+/// was scheduled under and re-checks it (still on the main actor) immediately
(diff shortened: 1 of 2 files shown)


ftchvs:
@codex please review this clear-wins concurrency fix for the disposable read-model cache. Key questions: (1) Is the main-actor epoch capture/recheck in persistReadModelCache() sufficient to guarantee the empty-state clear wins, given persist uses Task.detached and the empty path fires clearReadModelCache() inside a Task? (2) Is the defense-in-depth reliance on clearAll() being serialized on the store actor after the epoch bump sound? (3) Any residual window where a stale non-empty write could survive the remove-last-institution sequence?

PR #577, opened 2026-06-19
fix(ai): single FM boundary parser + bounded FM categorization (audit fm-core)

ftchvs:
## Summary

Fixes three Foundation-Models robustness findings from the code-review audit (stream `fm-core`). Scoped to the FM categorization/insight seam; does **not** touch `AppState.swift` (owned by another stream).

### (A) Dead `FMSpendingCategoryMapper` → made the live, single boundary parser
`FMSpendingCategoryMapper` (PlaidBarCore) was dead: the FM seam returned `SpendingCategory` directly, so the constrained string never crossed the boundary the mapper parses.

- `FMMerchantCategorizing.suggestCategory` now returns the **constrained `SpendingCategory.rawValue` String**.
- `FMMerchantCategorizer` resolves it via `FMSpendingCategoryMapper.category(from:)` — now the single, unit-tested string→enum boundary.
- PlaidBarCore stays free of any FoundationModels dependency.
- An unparseable/unknown value degrades to the NL fallback (never a wrong guess).

### (C) `FoundationModelsMerchantCategorizer.generate()` had no timeout/cancellation
The doc claimed timeouts were swallowed, but the `await` could block unbounded.

- Generation now races a structured `withTaskGroup` timeout child against `FMGenerationLimits.generationTimeout` (8s, defined in PlaidBarCore so it is shared/pure/testable).
- On timeout **or** outer-task cancellation it returns `nil`, so the NL fallback engages promptly and a superseded categorization stops.
- Doc comment aligned with reality.

### (B) Superseded-refresh cancellation in `FoundationModelsInsightModel` — documented follow-up
The root cause is upstream in `LocalInsightModelRuntime.withTimeout`, which runs the model in an unstructured `Task {}` that does **not** inherit outer (superseded-refresh) cancellation — so `Task.checkCancellation()` in the stream loop never fires for a superseded refresh and it runs to the 8s deadline. That helper file is owned by another stream, so I corrected the misleading "stops promptly" comment to state the limitation and the structural fix (run the operation as a `withThrowingTaskGroup` child). See follow-up.

## Testing
- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — passes.
- `swift test` — **1781 tests pass**.
- Added: `fmRawStringResolvedThroughMapper`, `fmUnparseableStringFallsBackToNL`; updated the FM stub to the new `String` contract.

## Follow-up
- (B) Fix `LocalInsightModelRuntime.withTimeout` to run its operation as a `withThrowingTaskGroup` child so outer cancellation propagates, making superseded FM insight refreshes stop without waiting out the 8s deadline.

### DIFF
--- a/Sources/PlaidBar/Services/FoundationModelsInsightModel.swift
+++ b/Sources/PlaidBar/Services/FoundationModelsInsightModel.swift
@@ -82,9 +82,17 @@ extension FoundationModelsInsightModel {
 
             // Single-pass consumption: each snapshot carries the cumulative
             // partially-generated content, so the LAST snapshot holds the final
-            // headline. Tracking it as we stream keeps cancellation responsive (a
-            // superseded refresh stops generation promptly) without depending on a
+            // headline. Tracking it as we stream lets generation stop at the next
+            // snapshot boundary once this task is cancelled, without depending on a
             // post-iteration `collect()`.
+            //
+            // NOTE: cancellation here is only responsive when *this* task is
+            // cancelled. Today `LocalInsightModelRuntime.withTimeout` runs the
+            // model in an unstructured `Task {}` that does not inherit the outer
+            // (superseded-refresh) cancellation, so a superseded refresh still runs
+            // to the runtime's 8s deadline. The structural fix lives in that shared
+            // runtime helper (run the operation as a `withThrowingTaskGroup` child
+            // so parent cancellation propagates); see PR follow-up.
             var latestHeadline: String?
             for try await snapshot in stream {
                 try Task.checkCancellation()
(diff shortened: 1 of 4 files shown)


ftchvs:
@codex please review this FM robustness fix. Key things to check: (1) the protocol boundary now returns the constrained SpendingCategory.rawValue String and FMMerchantCategorizer resolves via FMSpendingCategoryMapper — confirm the parser is genuinely the single source of truth and the unparseable path falls back to NL; (2) the withTaskGroup deadline in FoundationModelsMerchantCategorizer.generate() correctly returns nil on timeout AND propagates outer-task cancellation; (3) sanity-check the documented (B) limitation — outer cancellation cannot propagate through the unstructured Task{} in the unowned LocalInsightModelRuntime.withTimeout, so the real fix belongs there.

PR #578, opened 2026-06-19
fix(perf): bound large-history paging cost and stop dropping rows after re-sync

ftchvs:
## Summary

Fixes two confirmed code-review findings on the large-history paged transaction path (AND-567), in `TransactionCacheStore` and `PagedTransactionSource`.

### (A) Unbounded paging fetch + misleading docs — MEDIUM perf/doc
`TransactionCacheStore.page()`, `count()`, and `existingTransactionIds()` each ran an unbounded `fetch(FetchDescriptor<CachedTransaction>())` — materializing **every** row including the heavy `payload` JSON blob, then filtering/sorting/slicing in Swift. So every page loaded the whole history and re-sorted `O(N log N)`, defeating the documented bounded-paging goal (the doc comment falsely claimed "the DB does the heavy lifting … only up to `window.limit` rows are materialized").

The `FetchDescriptor(sortBy:)` / `#Predicate` KeyPath-non-`Sendable` constraint under Swift 6 strict concurrency is real (verified by a standalone typecheck probe — both fail; `propertiesToFetch`, which takes `PartialKeyPath`, compiles). The fix bounds the cost with the levers that *are* available:

- **Scalar-only ordering fetch** — `propertiesToFetch = [\.cacheKey, \.sortDate, \.transactionId]` excludes the `payload` blob, so JSON stays faulted out while the order is computed. Only the page slice then faults in and decodes its payloads (bounded to `window.limit`).
- **Memoized ordering per (data generation, cacheKey)** — a monotonic generation counter bumped by every write (`upsert`/`replaceAll`/`clearAll`) invalidates the cache, so a multi-page scroll sorts once instead of once per page. Keyed by `cacheKey` too, since one store file can briefly hold two environments.
- Corrected the misleading doc comments to describe the actual behavior.

Correctness preserved (newest-first, no gaps/overlap); all existing `TransactionCacheStoreTests` stay green.

### (B) Stale captured total drops a transaction after re-sync — LOW bug
`PagedTransactionSource.loadNextPageIfNeeded` used the `total` captured when the first page loaded and never re-read it. A concurrent re-sync that *grew* the history left the window math believing paging was done, silently dropping the newly added newest rows. It now re-reads `count()` and `updateTotal()` before computing the next window; the feed's per-id dedup keeps a shifted row from rendering twice.

## Testing
- `swift test` — 1785 tests pass
- New tests:
  - `TransactionCacheStoreTests`: cached ordering stable across repeated page reads; invalidated on insert/update/growth; scoped per `cacheKey` (no cross-key bleed within a generation).
  - `PagedTransactionFeedTests`: regression pinning the stale-total drop scenario + a stable re-read no-op. (`PagedTransactionSource` is in the `@main` app target and can't be `@testable import`ed, so the regression is pinned at the pure-Core feed layer the source's new control flow relies on.)

### DIFF
--- a/Sources/PlaidBar/Services/PagedTransactionSource.swift
+++ b/Sources/PlaidBar/Services/PagedTransactionSource.swift
@@ -117,11 +117,23 @@ final class PagedTransactionSource {
 
     /// Loads the next page when the list scrolls near the end. No-op outside the
     /// paged path, when a fetch is already running, or when there is nothing more.
+    ///
+    /// Before computing the next window it re-reads the cache `count()` and feeds
+    /// it back via `updateTotal`. The feed's `total` is captured when the first
+    /// page loads, so without this a concurrent re-sync that *grows* the history
+    /// would leave the window math believing there is nothing more to load and
+    /// silently drop the newly added newest transactions. Re-reading keeps the
+    /// window bound to the current size; the feed's per-id dedup keeps a row that
+    /// shifted between pages from rendering twice.
     func loadNextPageIfNeeded() async {
         guard mode == .paged, let store, let cacheKey, !isLoadingPage else { return }
-        guard let window = feed.nextWindow else { return }
         isLoadingPage = true
         defer { isLoadingPage = false }
+
+        if let currentTotal = try? await store.count(cacheKey: cacheKey) {
+            feed.updateTotal(currentTotal)
+        }
+        guard let window = feed.nextWindow else { return }
         guard let page = try? await store.page(cacheKey: cacheKey, window: window), !page.isEmpty else {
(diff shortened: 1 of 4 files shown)


ftchvs:
@codex please review this paged-history fix. Two things to scrutinize: (1) the new per-(generation,cacheKey) ordering cache in TransactionCacheStore — confirm invalidation on every write path (upsert/replaceAll via upsert, clearAll) is complete and that propertiesToFetch excluding payload still lets the page slice fault its blobs in correctly; (2) the loadNextPageIfNeeded re-read of count()+updateTotal — confirm it can't double-render or skip a row when total shrinks (vs grows) under the loaded window.

PR #579, opened 2026-06-19
fix(app): wire FM availability re-probe, live FM categorization, and gate energy-loop restarts

ftchvs:
## Summary

Three AppState runtime-wiring findings from the **appstate-runtime** audit stream. All fixes are additive, reversible, and unchanged on a non-FM / unconstrained-energy device.

### (A) FM availability never re-probed — `refreshFoundationModelsAvailability()` was dead code
Foundation Models availability was probed once in `AppState.init` and never again, and `refreshFoundationModelsAvailability()` had **zero callers**. Transient states (`.modelNotReady` while the model downloads, `.appleIntelligenceNotEnabled`) never recovered within a session — the user had to relaunch.

**Fix:** wired `refreshFoundationModelsAvailability()` to `NSApplication.didBecomeActiveNotification` (registered alongside the existing energy observers, removed in `deinit`). The function already guards `previous == current`, so the common no-change reactivation is a cheap no-op.

### (B) FM merchant categorizer wired but `suggest(for:)` never called — tier dormant
`merchantCategorizer` was built and wired but its `suggest(for:)` API was never invoked at runtime, so the FM categorization tier produced nothing.

**Fix:** added `refreshFoundationModelsCategorySuggestions()` (a no-op unless the probe reports `.available`) that calls the **existing** categorizer API for uncategorized Review Inbox rows and caches the `.foundationModels` results as **display-only** suggestions. `ReviewInboxView` triggers it via `.task` and surfaces the result through the existing "Suggested" badge + category pill. Suggestions never persist, never feed budget/export totals, and never bypass the review/override flow (`EffectiveCategoryResolver` stays the source of truth). The categorizer's own source files were **not** touched.

### (C) Energy observer restarted the loop + HTTP-probed on every notification
`handleEnergyStateChange()` restarted the whole refresh loop (which immediately issues a server connectivity probe) on **every** power/thermal notification, even transitions that don't cross the constrained boundary (e.g. `.nominal` → `.fair`).

**Fix:** cache the last `EnergyConditions.isConstrained` verdict and restart only when it actually flips.

## Testing
- `swift test` — **1783 tests pass**

The `PlaidBar` app target is an executable and cannot be `@testable`-imported (the test target depends only on `PlaidBarCore`, matching the existing "mirror AppState logic" pattern). New Swift Testing cases pin the pure decisions the wiring depends on:
- constrained-verdict boundary semantics + flip-count gating (mirrors `handleEnergyStateChange`)
- FM categorizer producing a `.foundationModels` suggestion when available, and skipping the model entirely when not

## Follow-ups
- Extract the verdict-flip gating decision (C) into a pure `EnergyAwareRefreshPolicy` Core helper so the gating itself (not just the underlying `isConstrained`) is directly unit-testable. Not done here because `EnergyAwareRefreshPolicy.swift` is owned by another stream.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -529,6 +529,19 @@ final class AppState {
     /// notifications that drive energy-aware background refresh (AND-568).
     /// Retained so they can be removed in `deinit`.
     private var energyStateObservers: [NSObjectProtocol] = []
+    /// The last energy-constrained verdict the power/thermal observers acted on.
+    /// Cached so a power/thermal notification only restarts the refresh loop (and
+    /// issues its connectivity probe) when the constrained verdict actually flips
+    /// — not on every transition that stays on the same side of the boundary
+    /// (e.g. `.nominal` → `.fair`, both unconstrained). `nil` until the first
+    /// notification establishes a baseline.
+    private var lastEnergyConstrainedVerdict: Bool?
+    /// Observer token for `NSApplication.didBecomeActiveNotification`, used to
+    /// re-probe Apple Foundation Models availability when the app reactivates so a
+    /// transient FM state (model still downloading, Apple Intelligence just turned
+    /// on in System Settings) recovers within the session instead of requiring a
+    /// relaunch (AND-563/564). Retained so it can be removed in `deinit`.
+    private var appActivationObserver: NSObjectProtocol?
     private let glanceSnapshotWriteDebouncer = GlanceSnapshotWriteDebouncer()
     private var glanceSnapshotWriteGeneration = 0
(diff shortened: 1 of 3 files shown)


ftchvs:
@codex please review this PR. Focus areas: (1) the NSApplication.didBecomeActiveNotification observer lifecycle (registration alongside energy observers, removal in isolated deinit) and MainActor.assumeIsolated usage; (2) the FM categorization seam — confirm suggestions stay display-only and never persist or bypass the review/override flow, and that the .task(id: snapshot.totalCount) trigger has no redundant model calls; (3) the energy verdict-flip caching in handleEnergyStateChange, including the nil-baseline first-notification behavior.

PR #588, opened 2026-06-20
feat(nav): Route deep-linking into the window-first window + glance-chip routing (AND-597)

ftchvs:
## Summary

Final sub of **Epic 2** (ADR-001 window-first migration). Makes the window-first shell **deep-linkable**: any surface can route a typed `Route` (destination + selection) into the primary `Window`, and the menu-bar **glance attention chips** now hand off into the window at the *relevant* destination (IA §3.6 / §6) instead of just raising the dashboard.

With this, **Epic 2 is functionally complete**: sidebar (AND-595) + ⌘K palette (AND-596) + deep-linking (AND-597).

### Routing API — the reusable entry point (also the App Intents Epic-8 path)
- **`AppState.route(to: Route, openWindow:)`** — a **pending-route handoff**. The primary scene is a declarative `Window` (not a `WindowGroup` with a presented value), so a route can't be threaded through the open call. Instead `route(to:)` stages the route on `AppState.pendingRoute`, then asks the injected `openWindow` to bring the `Window` forward.
- **`AppShellView`** consumes it via `onAppear` + `onChange(of: pendingRoute)` → `AppState.consumePendingRoute(into:)`, which applies it to the window's `NavigationModel` (destination + selection). `consumePendingRoute` clears the slot, so it's **single-shot** — the two triggers never double-apply. (`onAppear` covers "this route opened the window"; `onChange` covers "the window was already open".)
- App Intents (Epic 8) call this same method, passing the intent's `openWindow` — one definition for the whole deep-link path.

### Glance-chip routing (flag-gated)
- Pure **`Route.from(attentionRow:)`** in `PlaidBarCore`: financial low-cash + high-utilization → `accounts()`, unusual-spend → `transactions()`, degraded-item rows → `accounts(itemID:)`; infrastructure/generic rows return `nil` (they keep their existing in-place action). Pure + testable without the app target.
- New **`\.openRoute`** environment hook (mirrors `openCategoryDashboard` / `openReviewTable`), installed with a real handler **only when the flag is ON**. `AttentionQueueView.perform` routes when ON + a route exists, else falls back to today's action.

### Selection restoration (AND-594 extension)
- `NavigationModel` now persists/restores the last **destination** under a new `navigation.destination` key (absent ⇒ Dashboard), alongside the existing filter/selection/heatmap. `apply(_:)` and `go(to:)` persist it. Account selection already round-tripped; verified + covered.

## Flag-OFF note
Byte-identical with `WindowFirstFeatureFlag` OFF (the shipping default): nothing opens the `Window`, so `pendingRoute` is never read; `\.openRoute` stays its no-op default; and the chip route branch is additionally guarded by `WindowFirstFeatureFlag.resolved()` — belt-and-suspenders. The popover behaves exactly as today.

## Routing API signature
\`\`\`swift
// The reusable deep-link entry point (glance chips today, App Intents Epic 8):
@MainActor func route(to route: Route, openWindow: @MainActor () -> Void = {})
@MainActor func consumePendingRoute(into navigationModel: NavigationModel)
// Pure mapping (PlaidBarCore):
static func Route.from(attentionRow: AttentionQueueRow) -> Route?
\`\`\`

## Testing
- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — **passes**.
- `swift test` — **1885 passed** (+9 new over the ~1876 baseline; one pre-existing keychain test is *skipped* in this environment, not failed).
- New pure-layer tests: glance→route mapping (incl. driving the real `AttentionQueue.evaluate` so ids/`targetItemId` aren't invented), `apply(_:)` sets the destination for every route, `apply(.accounts(itemID: nil))` doesn't clobber an existing selection, destination restoration parity + Dashboard fallback + full route round-trip.

Refs: **AND-597**, **AND-580** (Epic 2), **ADR-001**.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -129,6 +129,18 @@ final class AppState {
     /// unchanged.
     let navigationModel: NavigationModel
 
+    /// Pending deep-link route awaiting the primary window (ADR-001 / AND-597).
+    ///
+    /// The window-first primary scene is a declarative `Window`, not a
+    /// `WindowGroup` with a presented value, so a deep-link cannot be passed *into*
+    /// the open call. Instead the opener sets this slot, then calls
+    /// `openWindow(id: "main")`; `AppShellView` consumes it on appear / change and
+    /// applies it through its window's `NavigationModel`. It is set only by
+    /// ``route(to:)`` and cleared by ``consumePendingRoute()``, so the round-trip
+    /// is single-shot. Inert with the flag OFF: nothing opens the window then, so
+    /// the slot is never read — flag-OFF behavior is unchanged.
+    var pendingRoute: Route?
+
     /// The dashboard's account filter, persisted under `dashboard.accountFilter`.
     /// Setting a *new* filter clears the account selection, exactly as the
     /// popover's `.onChange(of:)` did (AND-373/375).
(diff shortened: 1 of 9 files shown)


ftchvs:
@codex please review this PR (AND-597, final sub of Epic 2). Focus areas: (1) the pending-route handoff correctness — `AppState.route(to:openWindow:)` stages `pendingRoute` then opens the Window, and `AppShellView` consumes via onAppear + onChange(of: pendingRoute) → `consumePendingRoute(into:)` which clears the slot (single-shot; verify no double-apply and no missed hand-off in either window-already-open vs window-opened-by-route case); (2) the flag-OFF invariant — confirm the popover is byte-identical with WindowFirstFeatureFlag OFF (no-op `\.openRoute` default + the `isWindowFirstEnabled` guard in AttentionQueueView + nothing opens the Window so pendingRoute is never read); (3) the pure `Route.from(attentionRow:)` mapping — are the destination choices right per IA §3.6/§6, and is keying off row.id stable; (4) the new `navigation.destination` persistence key in NavigationModel (restore + Dashboard fallback). The routing API (`route(to:)`) is the reusable entry point App Intents (Epic 8) will reuse.

PR #595, opened 2026-06-20
feat(ux): chrome-only Liquid Glass + cross-window App-Lock/Privacy-Mask a11y parity (AND-588)

ftchvs:
## Summary

Epic 10 of the window-first migration (**ADR-001**, **AND-588**): Liquid Glass polish on **chrome only** plus cross-window accessibility finalization for the window-first shell. The flagship of this PR is the **App Lock security parity** fix — the window-first window previously could show data while App Lock was engaged; it now withholds content identically to the popover.

All changes live in the **window-first surface** (behind `WindowFirstFeatureFlag`, default **OFF**) or are behavior-preserving relocations, so the menu-bar popover is **byte-identical when the flag is OFF**.

## Chrome-only Liquid Glass (R-08)

- Replaced the unconditional scene-level `.containerBackground(.ultraThinMaterial, for: .window)` with `WindowChromeBackgroundModifier`, which **self-manages its Reduce Transparency degradation** rather than relying on the framework to thin the material (per the Epic 10 contract: custom translucency requires explicit a11y handling):
  - Resolves glass vs solid via the new pure, unit-tested `WindowChromeGlass.chromeBackground(reduceTransparency:)` in `PlaidBarCore`.
  - Glass = `.ultraThinMaterial`; **solid fallback** = opaque `windowBackgroundColor` (legible in light + dark).
  - `reduceTransparency` is `true` when the **system** Reduce Transparency setting is on **or** the user's reduced decorative-effects preference applies — the system setting always wins, mirroring the popover's `ResolvedDecorativeEffects.allowsTexture` gate so the two surfaces degrade identically.
- **Glass is confined to the navigation layer.** Audit of the shell + destinations:
  - Glass surfaces: the window container background and the sidebar footer `DataModeChip` (`.glassSurface(.inset)`) — both chrome. Destination section/card containers use `.glassSurface(.raised)` as the established card-chrome pattern.
  - **No glass on data:** the Transactions table, Review table, and chart **plot areas** carry no material (charts render on a `.clear`-fill card; marks are solid). Verified `TransactionsTable` / `TransactionsDestinationView` have zero `glass*` usage.
  - Encoded the policy as the pure `WindowChromeGlass.allowsGlass(on:)` (`.chrome` → true, `.data` → false).

## App Lock parity (security — the important parity work)

- Extracted `AppLockedGateView` into its own file (`Views/AppLockedGateView.swift`) so the popover, its detached host, and the window-first shell all use **one** gate. The popover's call site is unchanged (pure relocation).
- Mounted the gate on `AppShellView` as the **topmost** overlay: when content is **LOCKED** (not merely masked), the entire window is obscured — no balances, account/institution names, or sidebar badges leak (AND-462). Verified the window-first window **cannot show data while App Lock is engaged**.
- Gate transition gates on **Reduce Motion**; the ⌘K palette is force-dismissed while locked so unlocking never reveals a stale surface.

## Privacy Mask, text size, motion

- **Privacy Mask:** all real destinations (Transactions, Budgets, Planning, Insights, Review, Alerts) already mask via `appState.shouldMaskFinancialValues`; the glance/widgets already redact. Pinned the cross-window parity invariant in tests.
- **Text size (AND-570):** the shell already applies `.appliesAppTextSize()` at the scene root — confirmed; no change needed.
- **Reduce Motion:** existing shell springs/matched-geometry (AND-577) already gate on Reduce Motion; the new App-Lock transition does too.

## Audio-graph coverage gap closed

The task asked to confirm no shell chart lacks an audio graph. `ProjectedBalanceChart` (Planning) was the one chart with a spoken `accessibilityLabel` but **no scrubbable audio graph** (the net-worth trend and spend donut both had one). Added `ChartAudioGraph.projection(_:)` (a continuous line series, projected-low day called out, privacy-aware via the shared value-description path) and wired `.audioGraph(...)` onto the chart.

## Testing

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — **passes**.
- `swift test` — **1986 tests pass** (~1971 baseline + 15 new).
- New tests: `WindowChromeGlassTests` (glass-vs-solid decision + chrome-only policy + system-setting-wins), `WindowFirstAppLockParityTests` (locked withholds / masked dots / normal shows, on every surface), and `ChartAudioGraph.projection` builder tests.

## Parity gaps found (for the Epic 9 sign-off checklist)

- **[fixed in this PR]** Window-first window did not gate on App Lock — content was visible while locked. Now gated.
- **[fixed in this PR]** `ProjectedBalanceChart` lacked an audio graph. Now has one.
- **[note]** `.glassSurface(.raised)` is applied to destination *card containers* that hold dense data (e.g. Budgets `treeSection`, Insights/Budgets donut cards). This is the established card-as-chrome pattern from the popover (the chart/list marks themselves stay solid), consistent with how R-08 is interpreted in already-merged destinations. Flagging for explicit sign-off rather than silently changing merged patterns.

Refs: **AND-588**, **ADR-001**, risk **R-08**, AND-462, AND-570, AND-569, AND-577.

### DIFF
--- a/Sources/PlaidBar/App/PlaidBarApp.swift
+++ b/Sources/PlaidBar/App/PlaidBarApp.swift
@@ -478,11 +478,15 @@ struct PlaidBarApp: App {
                 // the window-first flag), so flag-OFF behavior is unchanged.
                 .onAppear { windowActivationPolicy.onWindowAppear() }
                 .onDisappear { windowActivationPolicy.onWindowDisappear() }
-                // Liquid Glass on chrome only (ADR-001): the window background is
-                // the ultra-thin material; data surfaces inside stay opaque. This
-                // is a content modifier (it walks up to the window), so it lives on
-                // the scene's root view, not on the `Window` scene itself.
-                .containerBackground(.ultraThinMaterial, for: .window)
+                // Liquid Glass on chrome only (ADR-001, R-08): the window
+                // background is the ultra-thin material; data surfaces inside stay
+                // opaque. Under Reduce Transparency (system setting or the reduced
+                // decorative-effects preference) it falls back to a fully solid
+                // window background — custom translucency self-manages its a11y
+                // degradation rather than relying on the framework to thin the
+                // material (AND-588). The glass-vs-solid decision is the pure,
+                // unit-tested `WindowChromeGlass.chromeBackground(reduceTransparency:)`.
+                .appliesWindowChromeBackground()
         }
         // Do not steal launch/activation: the window only appears when opened.
         .defaultLaunchBehavior(.suppressed)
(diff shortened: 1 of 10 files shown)


ftchvs:
@codex please review, focus on glass-on-data avoidance + App-Lock/Privacy-Mask parity:

- **Glass-on-data (R-08):** confirm Liquid Glass touches chrome only. The scene chrome background (WindowChromeBackgroundModifier) and sidebar DataModeChip are chrome; the Transactions/Review tables and chart plot areas stay solid. Flag any `.glassSurface`/`.glassEffect`/material that lands on a list row, table, or chart mark. Note the deliberate card-as-chrome pattern (`.glassSurface(.raised)` on destination *cards* that contain data) — call out if you think any of those cross the line.
- **App Lock parity:** verify AppShellView's `.overlay { if appState.isContentLocked { AppLockedGateView(...) } }` is the topmost overlay and truly withholds all content (no leak of account/institution names or sidebar badges) while locked, matching MainPopover. Confirm the palette force-dismiss on lock and that nothing renders the window data path before the gate.
- **Reduce Transparency fallback:** confirm WindowChromeGlass.chromeBackground returns .solid under system Reduce Transparency AND the reduced decorative-effects preference, and that the solid fallback (windowBackgroundColor) is legible in light + dark.
- **Reduce Motion:** the App-Lock gate transition gates on reduceMotion via MotionTokens.animation.
- **Flag-OFF inertness:** confirm the popover is byte-identical — the only popover-touching change is the AppLockedGateView relocation (same view, same call site).
- **Audio graph:** sanity-check ChartAudioGraph.projection (continuous series, projected-low call-out, privacy-aware value descriptions).

PR #596, opened 2026-06-20
fix(nav): move Budgets/Alerts selection into per-window NavigationModel (AND-621, R-10)

ftchvs:
## Summary

Today `BudgetsDestinationView` and `AlertsDestinationView` share their content↔inspector selection through process-wide singletons (`BudgetsSelectionModel.shared`, `AlertsSelectionModel.shared`). Two split-view columns can't share `@State`, so the singletons were the bridge — but that means two open windows share one selection, violating **R-10** (per-window navigation state, never a singleton).

This moves that selection onto the per-window `NavigationState`/`NavigationModel`:

- **`NavigationState`** (PlaidBarCore, pure/testable) gains:
  - `budgetCategorySelection: SpendingCategory?` — `nil` content-gates the Budgets inspector prompt.
  - `alertSelection: String` — `""` sentinel, matching the existing `selectedTransactionID` style.
  - `acknowledgedAlertIDs: Set<String>` — the alerts acknowledgement set (previously on the Alerts singleton; also per-window so two windows acknowledge independently).
  - Pure transitions: `selectBudgetCategory`/`deselectBudgetCategory`, `selectAlert`/`deselectAlert`, `acknowledgeAlert`/`unacknowledgeAlert`/`acknowledgeAllAlerts`, and `pruneAlerts(toRowsIn:)` (self-heal, reusing `AlertsInbox.pruneAcknowledgedIDs`).
- **`NavigationModel`** exposes a thin façade. Selection/acknowledgement is **not persisted** — it is ephemeral session-scoped per-window state, exactly as the retired singletons were.
- **Both destination views** (content + inspector columns) read/write `appState.navigationModel` (the established pattern used by the Transactions destination) instead of the singletons.
- **Deleted** `BudgetsSelectionModel.swift` + `AlertsSelectionModel.swift`.

## Issue / ADR
- AND-621 (R-10: move selection singletons into per-window NavigationModel)
- ADR-001 window-first migration

## Flag-off note
These destination views mount only when `AppShellView` is built behind `WindowFirstFeatureFlag` (default **OFF**); with the flag off they are never instantiated. The `NavigationModel` changes are purely **additive** (new fields/methods, no existing field restructured) and **non-persisted**, so the shipping popover surface is **byte-identical**. Field names use the exact `budgetCategorySelection` / `alertSelection` agreed for cross-stream mergeability (no `goalSelection` added — owned by the other stream).

## Testing
- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — passes.
- `swift test` — **1992 passed** (1986 baseline + 6 new).
- New R-10 tests in `RouteNavigationTests` cover: independent Budgets selection, independent Alert selection, independent acknowledgement across windows, `pruneAlerts` self-heal (drops stale selection + resolved acked ids, keeps live), and `acknowledgeAllAlerts`.


ftchvs:
@codex please review this PR. Focus areas: (1) R-10 correctness — Budgets/Alerts selection + acknowledgement now live on the per-window NavigationState/NavigationModel (not singletons), confirm two windows are truly independent; (2) flag-off byte-identical guarantee — NavigationModel changes are additive + non-persisted and the two destination views mount only behind WindowFirstFeatureFlag (default OFF); (3) parity with the deleted singletons — selection/acknowledge/acknowledge-all and the pruneAlerts self-heal (stale selection + resolved acked ids) behave identically; (4) the "" sentinel for alertSelection matching the selectedTransactionID convention.

PR #602, opened 2026-06-20
feat(tooling): headless window-first render harness (AND-624)

ftchvs:
## What

Adds a **headless render harness** — `swift run PlaidBar --demo --render-window-first <dir>` — so the window-first surfaces can be screenshotted without an on-screen window. The live window-first `Window` scene won't present from automated sessions and there's no Screen Recording permission, so this rasterizes everything off-screen.

For each of the **9 in-shell destinations** (Dashboard, Review, Transactions, Budgets, Planning, Goals, Insights, Alerts, Accounts — `Settings` is the native scene, excluded), the harness:
1. routes the per-window `NavigationModel` to that destination,
2. hosts the routed `DestinationContentView` (plus `DestinationInspectorView` for 3-column destinations) in a far-off-screen `NSWindow` via `NSHostingView`,
3. lets SwiftUI run the full presentation/layout cycle, then
4. `cacheDisplay(in:to:)` → `<dir>/window-<destination>.png`.

It also emits `window-shell.png` (full `AppShellView`). Exits non-zero on any capture failure, mirroring `SnapshotRenderer`.

## Layout strategy (documented in the file header)

Two approaches were tried and rejected against the real views:
- **`ImageRenderer` (window-less):** renders 3-column (Table/List) destinations but leaves the 2-column destinations (Dashboard/Planning/Insights — `GeometryReader` + `ScrollView` roots) **blank** (GeometryReader resolves to zero outside a presented hierarchy).
- **A never-ordered off-screen `NSWindow`:** same blank result.

**Working path:** order the window front at an origin far beyond any physical display (`-50000,-50000`) and let the run loop process. SwiftUI then runs the full cycle (charts, scrolled and lazy content all draw) while never appearing on screen and needing no Screen Recording permission.

**Fallback note:** a headless `NavigationSplitView` lays out its content/detail columns but does **not** populate its sidebar `List`, so `window-shell.png` is a best-effort whole-shell reference; the per-destination images (which never go through the split view) are the contract.

## Tests
- `swift test` — **2035 pass** (baseline maintained; the GUI E2E is opt-in/skipped)
- New fast unit test in `PlaidBarCoreTests`: pins `isRenderingWindowFirst` flag detection + the 9-destinations-plus-shell coverage
- New opt-in E2E in `PlaidBarTests` (`VAULTPEEK_RENDER_HARNESS_E2E=1`): runs the **prebuilt** binary and asserts all 10 PNGs are written (invoking the binary directly — nesting `swift run` inside `swift test` deadlocks on the SwiftPM build lock). Verified locally: passes, 10 PNGs.

## Scope
Owns the new `WindowFirstSnapshotRenderer.swift` + `CommandLineOptions` + tests. The only edits outside that are two one-line wiring touches (`PlaidBarApp` calls `renderIfRequested`; `AppState` ORs the new flag into the existing detached-popover neutralization). **No** destination view, `AppShellView`, `DesignTokens`, or `DestinationRouter` edits — those are read-only here.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -766,7 +766,12 @@ final class AppState {
             : nil
         isDashboardDetached = DetachedDashboardPreferences.resolvedDetachedIntent(
             storedValue: storedDetached,
+            // Both headless renderers must run deterministically: a stale
+            // `dashboard.detached = true` would spawn the floating window and
+            // intercept either harness. The window-first harness builds its own
+            // off-screen windows, so it likewise wants no detached popover.
             isRenderingSnapshot: CommandLineOptions.isRenderingSnapshot()
+                || CommandLineOptions.isRenderingWindowFirst()
         )
     }
 
--- a/Sources/PlaidBar/App/PlaidBarApp.swift
+++ b/Sources/PlaidBar/App/PlaidBarApp.swift
@@ -89,6 +89,7 @@ struct PlaidBarApp: App {
             }
         }
         _ = SnapshotRenderer.renderIfRequested(appState: state)
+        _ = WindowFirstSnapshotRenderer.renderIfRequested(appState: state)
         // Debug aid: register as a regular app (Dock icon, app switcher) so
         // automation/permission systems that skip accessory apps can see it.
         if CommandLine.arguments.contains("--regular-activation") {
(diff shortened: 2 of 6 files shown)


ftchvs:
@codex please review this headless window-first render harness (AND-624). Focus areas: (1) the off-screen NSWindow + cacheDisplay capture path in WindowFirstSnapshotRenderer.render — correctness of the orderFrontRegardless/Task.sleep settle loop and the isReleasedWhenClosed=false teardown (an earlier loop crashed in the autorelease-pool drain); (2) whether routing the shared AppState.navigationModel per destination during capture is safe; (3) the documented NavigationSplitView sidebar-blank fallback for window-shell.png; (4) the opt-in E2E test invoking the prebuilt binary directly to avoid the nested-swift-run build-lock deadlock.

PR #629, opened 2026-06-21
fix: 2 reachable bugs from bug-hunt round 5 (transactions search perf, FM cache eviction)

ftchvs:
## Bug-hunt round 5 — 2 reachable fixes

The cross-cutting (round 5) adversarial bug-hunt confirmed 3 real defects; the 2 **reachable** ones are fixed here. The 3rd (a webhook concurrent-delivery race) is on a currently-dormant code path and is tracked in AND-634 alongside the other webhook ordering defects, to be fixed when a real Plaid webhook signature verifier is wired.

### 1. Transactions search input-lag (P2) — `TransactionsDestinationView.swift`
Per keystroke, the uncached `rows` computed property ran the full **build → filter → sort** pipeline 3–4× per body pass (it was read by the filter bar's result count, the selection-reconcile `onChange`, `content`, and `table`), and the O(n) override-aware build (`EffectiveCategoryResolver` + metadata dictionary) re-ran on every character even though search doesn't change it.
- **Part 1:** `body` binds `let resolved = rows` once and threads it to every reader (`content`/`table` converted to functions taking the rows). Eliminates the 3–4× redundant pipeline runs.
- **Part 2:** a small `@MainActor @Observable BuiltRowCache` (in `@State`, mirroring `AppState._cachedCategoryDashboardPresentation`) memoizes `TransactionWorkspace.rows(transactions:metadata:rules:)` keyed by value-equality on `(transactions, reviewMetadata, rules)` — rebuilds only when those change, **not** on `searchText`. Per keystroke now runs only the cheap `filtered(...)` scan + `sort.sorted(...)`.

Behavior-identical (same final rows). `TransactionWorkspace` was already split into `rows`/`filtered`/`Sort.sorted`, so no Core change was needed.

### 2. Foundation Models suggestion cache grows unbounded (P3) — `AppState.swift`
`_foundationModelsCategorySuggestions` was insert-only, so it tracked session-cumulative review throughput instead of live inbox size — slow memory creep over a long session. `refreshFoundationModelsCategorySuggestions()` now prunes to the live Review Inbox before computing pending:
```swift
let liveIDs = Set(transactionReviewInboxSnapshot.items.map(\.id))
_foundationModelsCategorySuggestions = _foundationModelsCategorySuggestions.filter { liveIDs.contains($0.key) }
```
Display-behavior-preserving (suggestions are only consulted for items currently in the inbox). +2 regression tests pinning the eviction expression.

### Verify (combined merge of both fixes)
- Strict-concurrency build: ✅ `Build complete! (71.63s)`
- `swift test`: ✅ **2091 passed**

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -1927,6 +1927,14 @@ final class AppState {
     func refreshFoundationModelsCategorySuggestions() async {
         guard foundationModelsTierState.isAvailable else { return }
         let categorizer = merchantCategorizer
+        // Evict cache entries whose transactions have left the Review Inbox
+        // (categorized/approved/ignored/dropped). The cache is otherwise
+        // insert-only, so without this prune it grows with session-cumulative
+        // throughput rather than live inbox size (unbounded memory creep).
+        // `foundationModelsCategorySuggestion(for:)` only reads ids currently
+        // in the inbox, so dropping absent ids removes nothing on display.
+        let liveIDs = Set(transactionReviewInboxSnapshot.items.map(\.id))
+        _foundationModelsCategorySuggestions = _foundationModelsCategorySuggestions.filter { liveIDs.contains($0.key) }
         // Only items still flagged as needing a category benefit from a
         // suggestion; a row the user already categorized must not be second-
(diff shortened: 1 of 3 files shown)


ftchvs:
@codex review

PR #639, opened 2026-06-22
feat(ai): smarter on-device expense + income categories with restorable Plaid fallback

ftchvs:
Post-1.0 priority #5. On-device AI smarter expense + income categorization, Plaid kept as the auditable/visible/restorable fallback.

- **Plaid auditable fallback (Core, no AI):** `Resolution` gains additive `plaidCategory` + `isOverridingPlaid` *display* signals — NOT in the budget precedence chain. Budget-precedence invariant unchanged (override→rule→confident Plaid→uncategorized; NL/FM stay review-only).
- **Restorable in UI:** `AppState.clearReviewCategory` (undoable); inspector shows 'Plaid classified as X' + 'Restore Plaid category'; 'Uncategorized' clears the override. Text+icon, never color-alone.
- **Smarter prompts:** `CategorySuggestionContext` — injection-safe, identifier-free (Plaid hint + recurring + inflow/outflow); threaded via additive protocol overload.
- **Income taxonomy (Core):** `IncomeCategory` + deterministic `IncomeMerchantClassifier` (lexicon + heuristic).
- **Income FM tier:** `FMIncomeCategorizing` + constrained `@Generable` enum, gated on the FM probe; degrades to the heuristic floor. On-device only.
- **Income in app:** AppState income-suggestion cache (mirrors+prunes expense cache); inspector renders subtype + Suggested badge.

16 files (8 new). Strict build ✅, `swift test` ✅ 2143; the 3 budget-math suites pass; on-device/identifier-free guarantees test-covered.

PR #641, opened 2026-06-22
fix(ai): strip quote/delimiters from merchant FM prompt (injection symmetry with #639)

ftchvs:
Follow-up flagged during #639 review. The bare-merchant FM categorizer embedded the merchant in a quoted prompt (`"Categorize this merchant: \"…\""`) but only collapsed whitespace/control chars — a merchant containing `"`/`;`/`:` could break out and inject same-line instructions (milder than the #639 case since output is enum-constrained). Now strips those delimiters, mirroring `CategorySuggestionContext.sanitizedLine`. App-target seam (not unit-testable); strict build + full suite green.


ftchvs:
@codex review

PR #655, opened 2026-06-25
fix(ai): gate Foundation Models routing on FoundationModelsMacros availability (AND-656)

ftchvs:
## What & why

Follow-up to the Codex review of #649. The Foundation Models **availability probe** is the single source of truth for the FM tier state, but it only gated on `#if canImport(FoundationModels)`. On a **CLT-only macOS 26** environment the framework imports, but the `@Generable`/`@Guide` macros (`FoundationModelsMacros`) do **not** — so `SystemLanguageModel.default.availability` could report `.available` for a configuration that cannot compile or run the macro-backed generation path. AppState would then wire the FM model/categorizers on a build where generation can't actually work.

## Change

- `FoundationModelsAvailabilityProbe.swift`: gate the import block, `currentState()`, and the `map(_:)` extension on `#if canImport(FoundationModels) && canImport(FoundationModelsMacros)` — matching the merchant/income categorizers, which already dual-gate. When the macros are absent the probe returns `.unsupported`, `FMCategorizationTierDecision.shouldAttemptFoundationModels` returns `false`, and routing falls back to the existing NaturalLanguage/heuristic path byte-identically (existing regression guards in `FMMerchantCategorizerTests` hold).
- `PlaidBarTests.swift`: new deterministic, source-invariant regression test asserting every `#if canImport(FoundationModels)` in the probe also requires `FoundationModelsMacros`, and that the categorizers stay in lockstep.

## Testing

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → Build complete.
- `swift test --filter foundationModelsProbeGatesOnMacroAvailability` → passed.

## Links

Closes AND-656. Addresses Codex review findings on #649.

### DIFF
--- a/Sources/PlaidBar/Services/FoundationModelsAvailabilityProbe.swift
+++ b/Sources/PlaidBar/Services/FoundationModelsAvailabilityProbe.swift
@@ -1,15 +1,7 @@
 import Foundation
 import PlaidBarCore
 
-// Gate on BOTH the framework and its macro module. On a CLT-only macOS 26 you
-// can `import FoundationModels` but not `FoundationModelsMacros`, so the
-// `@Generable`/`@Guide` macros the categorizers rely on are unavailable even
-// though `SystemLanguageModel.default.availability` may report `.available`.
-// The merchant/income categorizers already dual-gate on the same condition, so
-// matching it here keeps the probe (the single source of truth for the FM tier
-// state) from reporting `.available` for a configuration that cannot compile or
-// run the macro-backed generation path (AND-656).
-#if canImport(FoundationModels) && canImport(FoundationModelsMacros)
+#if canImport(FoundationModels)
 import FoundationModels
 #endif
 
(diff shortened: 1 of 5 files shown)


ftchvs (commented):
Reviewed and pushed a fix in `e0a4177`.

Finding fixed: the availability gate was checking `FoundationModelsMacros` instead of the public `FoundationModels` framework path, which could false-negative on current SDKs and keep local Foundation Models routing disabled even when `SystemLanguageModel.default.availability` is available.

Changes made:
- Removed the `FoundationModelsMacros` compile gate from the availability probe and Foundation Models categorizer/model paths.
- Kept runtime routing behind `SystemLanguageModel.default.availability`.
- Added a source-level regression test so this does not drift back to the macro package gate.

Validation:
- `git diff --check`
- `swift build --target PlaidBar --skip-update -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors`
- `swift test --filter foundationModelsProbeUsesPublicFrameworkGate`

Docs checked: Sosumi/Apple Foundation Models availability docs and Context7 Swift strict-concurrency guidance. No merge performed.

PR #656, opened 2026-06-25
docs(privacy): note cache files follow PLAIDBAR_DATA_DIR and can sync off-device (AND-657)

ftchvs:
## What & why

The docs frame VaultPeek data as "local-only" / "never synced," but the disposable JSON read-model cache (target `PlaidBarCache`, AND-566) is written under the directory chosen by `PLAIDBAR_DATA_DIR` (default `~/.vaultpeek/`, which is local). If a user points `PLAIDBAR_DATA_DIR` at a cloud-synced folder (iCloud Drive, Dropbox, OneDrive, a synced network share, etc.), that sync client can copy the cache files off-device despite the "local-only" framing. The cache holds **derived** financial read-model data (balances, merchant names, amounts, Plaid identifiers) but **never** Plaid secrets, access tokens, or any Keychain material — those stay device-only regardless (`…ThisDeviceOnly`, non-synchronizable).

This is a docs-only clarification: add an accurate caveat plus a recommendation to point `PLAIDBAR_DATA_DIR` at a local, non-synced directory. No code, no behavior change.

## Change (files)

- `SECURITY.md` — added a caveat bullet right after the read-model-cache description, clarifying that the existing "never synced to iCloud" guarantee is about iCloud Keychain and the App Group container, not the filesystem location of the data dir itself; and that overriding `PLAIDBAR_DATA_DIR` to a cloud-synced folder can copy the derived cache off-device.
- `README.md` — added a blockquote callout under the `PLAIDBAR_DATA_DIR` override note ("Keep `PLAIDBAR_DATA_DIR` local"), matching the existing callout style.
- `CLAUDE.md` — extended the inline "Data modes" data-dir note with the same caveat and recommendation, pointing to `SECURITY.md`.

## Testing

Docs-only change — no Swift build or test run is applicable. Validation gate run:

```
git diff --check        # → no whitespace/conflict errors (clean)
```

The strict-concurrency build and `swift test` are N/A here (no source files touched). The three edited files are `.md` docs only.

## Links

Closes AND-657.

For UI changes, screenshots can be regenerated via `./Scripts/screenshots.sh` — not applicable to this docs-only change.


ftchvs:
Reviewed and pushed a docs-only fix in `2f74b47`.

What changed:
- Tightened the `PLAIDBAR_DATA_DIR` guidance so it no longer claims every persisted artifact lives there.
- Clarified production Keychain behavior versus fallback/test SQLite token storage.
- Fixed the cache clearer wording so `clearReadModelCache()` and `clearTransactionCache()` are described as separate store clearers.
- Noted that widget/App Intent App Group glance snapshots are separate from the private data dir and can contain derived finance values after redaction rules run.

Validation:
- `git diff --check`

Docs context checked: Context7 Swift strict-concurrency guidance and Sosumi Apple documentation for App Intents / local-device authentication. I did not merge the PR.

PR #658, opened 2026-06-25
fix(ux): window-scale shared sub-components + unified toolbar search (AND-625)

ftchvs:
## What & why

Shared presentational sub-components that are re-hosted **verbatim** in both the compact menu-bar popover and the desk-distance window workspace were rendering at popover caption-scale everywhere. Inside a window card their figures looked shrunken next to the window's own `.body`-scale rows. This workstream (W9 / AND-625) fixes the type-role mismatch and extends the window's unified toolbar search to a second destination.

Two parts; the type-role fix (part 1) is the higher-value / lower-risk half and is the core of this PR.

### Part 1 — window-scale shared sub-components
- New `ComponentScale` (`.popover` / `.window`) hint in `WindowMetrics.swift`, defaulting to `.popover` so every existing glance/popover call site stays **byte-for-byte** (flag-OFF parity). It carries no color and no financial meaning — only which type ramp to read at — so it is safe to thread through `Sendable` views.
- `RecurringObligationsSection` rows and `BalanceTimeMachineView` ("what the bank said" balance rows + restatement line) now switch, in their `.window` branch, to the window type ramp introduced in AND-628 — `WindowDataText` / `WindowFigureCaption` / `WindowBodyText` / `WindowCardTitle` — with tabular figures, matching how `DashboardDestinationView` / `AccountsDestinationView` read at desk distance.
- Window callers pass `scale: .window` (`DashboardRecurringCard`, `PlanningDestinationView`, the `DashboardOverviewColumn` "bank said" footer); popover hosts (`WealthSummaryFlyout`, `MainPopover`) keep the default, so the glance is unchanged.
- Privacy Mask / App Lock behavior is untouched — amounts still route through `PrivacyMaskPresentation`; only the font role changes.

### Part 2 — unified toolbar search propagation
- The window shell already owns ONE unified `.searchable` toolbar field (AND-624); its per-destination adoption explicitly anticipated a "propagation pass." This extends that pass to **Accounts**, which filters its account list by the shared query exactly like the Dashboard: case-insensitive display-name match (never a masked amount, so search stays safe under Privacy Mask), with a contextual `ContentUnavailableView.search` no-results state.
- Renamed the now-misnamed env key `dashboardSearchQuery` → `shellSearchQuery` (the field is shell-owned and shared by two destinations).
- **Follow-up:** `TransactionsDestinationView` keeps its own inline filter bar (`TransactionsFilterBar`) — migrating that to the shell field is a larger, more opinionated change (it carries date/category/amount facets, not just a name match) and is intentionally left out of this PR to keep the diff reviewable. Tracked as the next step of the same propagation pass.

## Change (files)
- `Sources/PlaidBar/Theme/WindowMetrics.swift` — add `ComponentScale` enum.
- `Sources/PlaidBar/Views/RecurringObligationsSection.swift` — `scale` param; scale-aware title/total/merchant/amount/detail roles.
- `Sources/PlaidBar/Views/BalanceTimeMachineView.swift` — `scale` param; scale-aware title/name/balance/date roles.
- `Sources/PlaidBar/Views/Destinations/DashboardRecurringCard.swift`, `.../PlanningDestinationView.swift`, `.../DashboardOverviewColumn.swift` — pass `scale: .window`.
- `Sources/PlaidBar/Views/Destinations/DashboardSearchQuery.swift` — env key `dashboardSearchQuery` → `shellSearchQuery` + docs.
- `Sources/PlaidBar/Views/AppShellView.swift` — Accounts adopts the unified field; prompt + injection updated.
- `Sources/PlaidBar/Views/Destinations/AccountsDestinationView.swift` — read `\.shellSearchQuery`, filter by display name, no-results state.
- `Tests/PlaidBarTests/PlaidBarTests.swift` — two source-invariant tests.

## Testing
Run with the SwiftPM sandbox disabled (SwiftPM spawns its own sandbox).

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete!** (strict-concurrency gate, the CI gate, passes).
- `swift test --filter sharedSubComponentsScaleToWindowRoles` → **1 test passed**.
- `swift test --filter accountsAdoptsUnifiedShellSearch` → **1 test passed**.
- `swift test --filter PlaidBarTests` → **53 tests passed** (no regression in the app suite).
- `git diff --check` → clean.

The two new tests are source-invariant (the app is an `@main` executable that can't be `@testable import`ed): they assert the shared sub-components carry a `ComponentScale` defaulting to `.popover` and reference the window type roles, that window call sites pass `.window` while the popover host does not, and that Accounts wires the unified `\.shellSearchQuery`.

UI changes are type-scale + a toolbar search field; screenshots can be regenerated via `./Scripts/screenshots.sh` (not run here — no fixture/data change).

## Links
Closes AND-625

### DIFF
--- a/Sources/PlaidBar/Theme/WindowMetrics.swift
+++ b/Sources/PlaidBar/Theme/WindowMetrics.swift
@@ -187,6 +187,31 @@ struct WindowFigureCaption: ViewModifier {
     }
 }
 
+// MARK: - Component scale (popover vs window)
+
+/// The surface a *shared* sub-component is rendering on, so one view can pick the
+/// right type ramp for its host (AND-625).
+///
+/// A handful of presentational sub-components (e.g. ``RecurringObligationsSection``,
+/// ``BalanceTimeMachineView``) are re-hosted verbatim in both the compact menu-bar
+/// popover and the desk-distance window workspace. Without this hint they read at
+/// popover caption-scale (`.caption2`/`.microText`) everywhere, so inside a window
+/// card their figures look shrunken next to the window's own `.body`-scale rows.
+///
+/// `.popover` is the default so every existing popover/glance call site stays
+/// byte-for-byte unchanged (flag-OFF parity); window call sites pass `.window`,
+/// which swaps the figure/caption roles to their window counterparts
+/// (``WindowDataText`` / ``WindowFigureCaption``). It carries no color and no
+/// financial meaning — only which type ramp to read at — so it is safe to thread
+/// through `Sendable` views.
+enum ComponentScale: Sendable {
+    /// Compact, arm's-length menu-bar popover (the default). Caption-scale figures.
+    case popover
+    /// Desk-distance window workspace. `.body`-scale tabular figures and `.caption2`
+    /// sub-labels, matching the window's own rows.
+    case window
+}
+
 extension View {
     /// Window page identity (`largeTitle`, bold).
     func windowLargeTitle() -> some View { modifier(WindowLargeTitle()) }
--- a/Sources/PlaidBar/Views/AppShellView.swift
+++ b/Sources/PlaidBar/Views/AppShellView.swift
@@ -104,6 +104,7 @@ struct AppShellView: View {
         )
 
         let destination = appState.navigationModel.destination
+        let isSearchEnabled = destinationSupportsSearch(destination) && !appState.isContentLocked
 
         // Column policy per destination (driven by the pure
         // `RouteDestination.prefersThreeColumnLayout` in PlaidBarCore). 3-column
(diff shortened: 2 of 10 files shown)


ftchvs (commented):
Reviewed and pushed a fix in `c9495f2`.

Findings fixed:
- Shell search stayed enabled while App Lock was active, so locked finance views could still receive and retain search input.
- The Accounts inspector/selection path resolved against all accounts instead of the currently filtered visible accounts, which could expose stale or hidden selection context after filtering.

Changes made:
- Disabled shell search when content is locked and cleared active search/palette state on lock.
- Resolved account selection and inspector state against the filtered visible accounts.
- Extended the AND-625 source-level test to cover lock-gated search and filtered selection.

Validation:
- `git diff --check`
- `swift test --filter accountsAdoptsUnifiedShellSearch`

No merge performed.

PR #660, opened 2026-06-25
feat(onboarding): window-first orientation moment for menu-bar users (AND-640)

ftchvs:
## What & why

VaultPeek is now window-first by default (AND-616): a fresh install lands in the primary `Window` workspace with the menu bar reduced to a glance. Users arriving from the menu-bar era have no signpost for the new model. This adds a **one-time first-launch orientation moment** — a sheet on the primary `Window` — that explains:

- the menu bar is the **calm glance** (quick status, a few key numbers one click away),
- the window is the **deeper workspace** (dashboard, transactions, budgets, planning, review),
- and that **App Lock / Privacy Mask apply to BOTH** surfaces.

It shows once on first window open, is dismissible, and is never shown again.

## Change (files)

- `Sources/PlaidBarCore/Models/WindowFirstOrientationCopy.swift` (new) — pure, `Sendable`, headless copy model (title, subtitle, three points, dismiss strings + pre-composed VoiceOver labels). Carries **only orientation copy, no financial values**, so it is safe under Privacy Mask / App Lock. `.standard` holds the shipped wording.
- `Sources/PlaidBar/Views/WindowFirstOrientationView.swift` (new) — thin SwiftUI renderer of the copy model. Decorative glyphs hidden from VoiceOver; the container announces one coherent phrase (`accessibilitySummary`); reduce-motion-gated fade-in via `MotionTokens.animation(_:reduceMotion:)`; default-action dismiss button. No theme flash (plain SwiftUI inheriting the window's already-applied appearance).
- `Sources/PlaidBar/App/AppState.swift` — adds `isWindowFirstOrientationDismissed`, mirroring `isFirstRunSnapshotDismissed`'s **context-aware UserDefaults key** (per Plaid environment + data directory) so it never re-shows once dismissed. Loaded synchronously at init, refreshed on context switch, with `dismissWindowFirstOrientation()` and the pure `shouldShowWindowFirstOrientation` gate (flag ON + not dismissed + not demo).
- `Sources/PlaidBar/App/PlaidBarApp.swift` — wires a `WindowFirstOrientationSheet` modifier onto the `Window("VaultPeek", id: "main")` scene, gated on `WindowFirstFeatureFlag.resolved()` + the dismissal flag.
- `Tests/PlaidBarCoreTests/WindowFirstOrientationCopyTests.swift` (new) — 8 tests for the pure copy model.

## Testing

```
swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors   # Build complete (PASS)
swift test --filter WindowFirstOrientationCopyTests                              # 8 tests passed (PASS)
swift test --filter WindowFirstFeatureFlagTests                                  # 7 tests passed — no regression
git diff --check                                                                 # clean
```

The new Core suite asserts: shipped copy shape (title/subtitle/3 points/dismiss), the three orientation beats (`menuBar`/`window`/`privacy`), unique point ids, the privacy point naming both controls and both surfaces, **no digits anywhere** (privacy-safe value-surface guard), per-point and whole-sheet VoiceOver label composition, and value-`Equatable` semantics.

UI is window-only; screenshots can be regenerated via `./Scripts/screenshots.sh`.

## Links

Closes AND-640

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -55,6 +55,7 @@ final class AppState {
         static let setupCompletedOnce = "setup.completedOnce"
         static let setupCompletedContextPrefix = "setup.completedOnce.context"
         static let firstRunSnapshotDismissedContextPrefix = "firstRunSnapshot.dismissed.context"
+        static let windowFirstOrientationDismissedContextPrefix = "windowFirstOrientation.dismissed.context"
         static let lastTransactionCacheContext = "cache.lastTransactionCacheContext"
         static let categoryBudgetCache = "cache.categoryBudgets"
         static let dashboardDetached = DetachedDashboardPreferences.detachedStorageKey
@@ -191,6 +192,15 @@ final class AppState {
         didSet {
             guard oldValue != isFirstRunSnapshotDismissed else { return }
             persistFirstRunSnapshotDismissal(isFirstRunSnapshotDismissed)
+        }
+    }
+    /// Whether the one-time window-first orientation moment (AND-640) has been
+    /// dismissed. Persisted per environment + data directory (mirroring the
+    /// first-run snapshot flag) so it never re-shows once the user has seen it.
+    var isWindowFirstOrientationDismissed = false {
+        didSet {
+            guard oldValue != isWindowFirstOrientationDismissed else { return }
+            persistWindowFirstOrientationDismissal(isWindowFirstOrientationDismissed)
         }
     }
     var serverConnected = false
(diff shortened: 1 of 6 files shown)


ftchvs (commented):
Reviewed and pushed a fix in `7c458eb`.

Finding fixed: the window-first orientation sheet could present while App Lock was active, and a programmatic dismissal caused by locking/unlocking could be treated as a user dismissal. That risked showing onboarding over a locked finance surface and marking the orientation as dismissed without the user actually dismissing it.

Changes made:
- Added `!isContentLocked` to the pure orientation presentation gate.
- Split user dismissals from programmatic dismissals so lock-driven hides do not persist dismissal state.
- Added a source-level regression test for the App Lock gate and dismissal behavior.

Validation:
- `git diff --check`
- `swift test --filter windowFirstOrientationRequiresUnlockedContent`

Note: the focused test build emitted pre-existing strict-concurrency warnings in files outside this fix’s changed logic, including `WindowActivationPolicy`. I kept this patch scoped to the App Lock/orientation bug. No merge performed.

PR #661, opened 2026-06-25
fix(transactions): surface in-session sync head rows in paged list (AND-632)

ftchvs:
## What & why

In the window-first Transactions list in `.paged` mode, the view renders the loaded cache pages, **not** the live `appState.transactions`. So an in-session sync that prepended newer (head) transactions left them hidden until the view was recreated — `onChange(of: appState.transactions)` only refreshed the render-dead in-memory fallback, never the paged rows.

A first fix (PR #624) re-seeded from the cache (re-read page 0) and was **pulled** after Codex flagged two regressions:
1. **Cache-timing**: re-reading page 0 could read the *old* cache before the detached `persistTransactionCache()` committed — resurrecting stale rows and re-hiding the fresh head.
2. **Filter drain**: resetting the feed to page 0 while a filter was active dropped later loaded pages, so active-filter matches on those pages vanished behind the filtered-empty state (the drain task is keyed only on `transactionFilter`, which hadn't changed).

This implements the correct shape: **merge the live array's new head into the already-loaded feed** — no cache re-read, no page dropped.

## Change (files)

- `Sources/PlaidBarCore/Utilities/PagedTransactionFeed.swift` — new pure `mergeHead(from:)`: walks the newest-first live array from the front, collecting rows not yet loaded and **stopping at the first id that is** (the boundary where the live head meets the paged window); prepends that prefix (deduped, newest-first) and grows `total`. A **lineage guard** (the loaded head must still appear in the live array) refuses a wholesale replace — that is `reset(total:)`'s job, never a blind prepend onto stale rows.
- `Sources/PlaidBar/Services/PagedTransactionSource.swift` — new `resyncHead(_:)`: refreshes the fallback and, **in paged mode only**, calls `mergeHead`. No cache read → no race with the detached persist; no page dropped → an active filter's later-page matches survive.
- `Sources/PlaidBar/Views/Destinations/TransactionsDestinationView.swift` — `onChange(of: appState.transactions)` now calls `resyncHead(latest)` instead of `updateFallback(latest)`.
- `Tests/PlaidBarCoreTests/PagedTransactionFeedTests.swift` — 7 new deterministic tests.

**Why it's safe under every filter (Cash/Credit/search):** the cache mirrors `appState.transactions` verbatim (`replaceAll`), so prepending the live head can never hide older history; untouched later pages keep their drained matches, so no filter re-key/re-drain is needed. Neither Codex regression can recur — there is no cache read and no page reset on the sync path. Fallback / demo / cache-unavailable paths are byte-identical (mergeHead no-ops outside `.paged`).

## Testing

```
# → Build complete! (strict-concurrency, warnings-as-errors: PASS)

swift test --filter PagedTransactionFeedTests
# → Test run with 20 tests passed (7 new)

git diff --check
# → clean
```

New tests: `mergeHead` prepend (AND-632) · idempotent no-op · boundary stop (never pulls later-page rows) · pre-paging no-op · **wholesale-replace refusal** (lineage guard) · and an **end-to-end repro**: a mid-session sync's new head row is visible under an active `searchText` filter via `TransactionWorkspace.filtered` (asserts the row is absent before the sync and present after `mergeHead`, with no other row dropped).

## Links

Closes AND-632.

Gate/risk note: pure-Core merge keyed on the in-memory array only — no cache read, no feed reset, no page drop on the sync path, so neither of the two regressions that pulled #624 can recur. Outside paged mode (fallback/demo/cache-unavailable) `resyncHead` is equivalent to the old `updateFallback`.

### DIFF
--- a/Sources/PlaidBar/Services/PagedTransactionSource.swift
+++ b/Sources/PlaidBar/Services/PagedTransactionSource.swift
@@ -165,6 +165,17 @@ final class PagedTransactionSource {
     func resyncHead(_ transactions: [TransactionDTO]) {
         fallbackTransactions = transactions
         guard mode == .paged else { return }
-        feed.mergeHead(from: transactions)
+        feed.mergeHead(from: Self.newestFirstPreservingSameDateOrder(transactions))
+    }
+
+    private static func newestFirstPreservingSameDateOrder(_ transactions: [TransactionDTO]) -> [TransactionDTO] {
+        transactions.enumerated()
+            .sorted { lhs, rhs in
+                if lhs.element.date != rhs.element.date {
+                    return lhs.element.date > rhs.element.date
+                }
+                return lhs.offset < rhs.offset
+            }
+            .map(\.element)
     }
 }
--- a/Sources/PlaidBar/Services/PagedTransactionSource.swift
+++ b/Sources/PlaidBar/Services/PagedTransactionSource.swift
@@ -147,4 +147,24 @@ final class PagedTransactionSource {
     func updateFallback(_ transactions: [TransactionDTO]) {
         fallbackTransactions = transactions
     }
+
+    /// Reconciles the rendered rows with a freshly-synced in-memory history
+    /// (AND-632).
+    ///
+    /// Always refreshes the fallback so the regression-safe path stays current. In
+    /// **paged** mode it additionally merges the live array's *new head* into the
+    /// loaded pages via the pure ``PagedTransactionFeed/mergeHead(from:)`` — so an
+    /// in-session sync's newest transactions appear immediately instead of staying
+    /// hidden behind the cache pages until the view is recreated.
+    ///
+    /// Crucially this never re-reads the cache and never drops a loaded page:
+    /// - No cache read → no race against the not-yet-committed detached persist
+    ///   (the regression that pulled the first attempt, PR #624).
+    /// - Loaded later pages are untouched → an active filter that drained later
+    ///   pages keeps its matches; nothing needs re-draining or re-keying.
+    func resyncHead(_ transactions: [TransactionDTO]) {
+        fallbackTransactions = transactions
+        guard mode == .paged else { return }
+        feed.mergeHead(from: transactions)
+    }
 }
--- a/Sources/PlaidBar/Views/Destinations/TransactionsDestinationView.swift
+++ b/Sources/PlaidBar/Views/Destinations/TransactionsDestinationView.swift
@@ -119,10 +119,15 @@ struct TransactionsDestinationView: View {
             pagedSource = appState.makePagedTransactionSource(fallback: appState.transactions)
             await pagedSource.loadFirstPageIfNeeded()
         }
-        // Keep the fallback rows fresh when a refresh replaces the in-memory array,
-        // so the list reflects new data even before/without a cache page.
+        // Reconcile the rendered rows when an in-session sync replaces the in-memory
+        // array: refresh the fallback AND, in paged mode, merge the new head rows
+        // into the loaded pages so a mid-session sync's newest transactions surface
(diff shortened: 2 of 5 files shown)


ftchvs:

**Verdict: LGTM (minor nits)** — the fix is correctly shaped and neither of the two regressions that pulled PR #624 can recur. Verified locally: `swift test --filter PagedTransactionFeedTests` → 20/20 pass; strict-concurrency `-warnings-as-errors` build → clean.

### Why the two pulled-#624 regressions can't recur (confirmed)
- **(a) Stale-cache resurrection** — `mergeHead` reads nothing from the cache; it only prepends rows already in hand from the live `appState.transactions`. No detached-persist race. ✓
- **(b) Filter drain on later pages** — `resyncHead` never resets the feed or drops a loaded page; `loaded.insert(at: 0)` leaves drained later-page matches untouched, so no re-key/re-drain is needed. ✓
- **(c) Duplicates** — `mergeHead` skips any id in `loadedIds` and breaks at the first already-loaded id, so a row can't render twice. The `Set(loaded).count == loaded.count` assertion backs this. ✓
- Prepend is correctly gated to `mode == .paged` only (fallback/demo/cache-unavailable stay byte-identical to the old `updateFallback`), and the lineage guard refuses a wholesale replace. ✓

### Observations (non-blocking)
- **Offset drift after prepend is safe, but subtle and untested.** After `mergeHead` grows `loaded`/`total` by N, the next `loadNextPageIfNeeded` still computes its window from `lastLoadedPageIndex+1` × `pageSize` against the now-+N-shifted cache. That window *re-reads N already-loaded rows at its leading edge* (cache offsets shifted down by N), which `appendPage`'s id-dedup silently absorbs — so there's overlap, never a gap, and no row is skipped or lost. Correct, but it relies on (cache is monotonic newest-first) ∧ (dedup-by-id). Worth a one-line unit test asserting "after `mergeHead`, walking the remaining pages still yields the full history with no gap/dup," since this cross-method interaction (`mergeHead` + later `appendPage`) is the exact seam #624 broke, and no test currently exercises it. `PagedTransactionSource.swift:128`.
- **Lineage guard is a linear `contains` scan of the live array on every sync.** Fine for now, but on a multi-thousand-row history every `onChange(of: appState.transactions)` does an O(n) scan plus the O(prefix) walk. Acceptable; just noting it isn't free. `PagedTransactionFeed.swift:~107`.
- **Head-removal staleness is unchanged (pre-existing, out of scope).** If a sync *removes* the current loaded-head row, the lineage guard correctly no-ops (right call — `reset` owns replacement), but the removed row stays rendered until view recreation. Not introduced here; flagging only so it's a known gap, not a re-review surprise.

### Scope / safety
- Change stays in declared scope (one pure Core method, one source method, one `onChange` line, tests). No Privacy Mask surface touched — `mergeHead` reorders DTOs and never formats/exposes a financial value, so masking is unaffected. No color-only signaling introduced. Strict-concurrency clean (`@MainActor`-isolated source, pure mutating func on a `Sendable` value type).

**Must-fix: none.** Recommend (optional) adding the `mergeHead`-then-continue-paging gap/dup test before merge, given this is the second attempt at this fix.

ftchvs (commented):
Reviewed and pushed a fix in `7136537`.

Finding fixed: `resyncHead` merged the live `appState.transactions` array without normalizing newest-first order. Because the reducer appends unseen incoming rows, the loaded head could remain ahead of truly newer live rows and `mergeHead` would prepend nothing, leaving stale transaction pages visible after sync.

Changes made:
- Normalize live rows newest-first before merging the paged feed head.
- Preserve original relative order for same-day transactions.
- Added a source-level regression test for the normalization call.

Validation:
- `git diff --check`
- `swift test --filter pagedTransactionResyncSortsLiveRowsNewestFirstBeforeMerge`

No merge performed.

PR #662, opened 2026-06-25
fix(cache): treat incompatible store as disposable miss + non-blocking open (AND-656)

ftchvs:
## What & why

Closes the two remaining Codex findings on **AND-656** (PR #649) that PR #655 (FoundationModelsMacros gate) did not cover. Both are about the disposable read-model and per-transaction cache *open/load* path in `PlaidBarCache`.

- **Finding 2 — incompatible store permanently disabled the cache.** After the SwiftData→JSON cache migration, an existing SwiftData `.store` file (or any corrupt file) decode-fails as JSON. The old throwing `init` meant the store could never reopen or replace itself until the user manually deleted the file — the cache stayed dead. It now self-heals: a decode failure on first read is a **disposable cache miss**, the unreadable file is discarded, and the next refresh rebuilds it. No permanent disable, no real-data loss (live in-memory data is always authoritative).
- **Finding 3 — synchronous full-history decode on the MainActor open path.** `transactionCacheStoreIfAvailable()` is `@MainActor`, and an actor's `init` runs in the caller's executor — so the old `init` decoded the entire JSON history synchronously on the UI thread, defeating paging and risking stalls. Opening now does **zero disk I/O**; the read+decode is deferred to the actor's executor on the first operation.

Also fixes the **AND-657 finding #2** doc inaccuracy: `clearReadModelCache()` wipes only the read-model store; the per-transaction wipe is `clearTransactionCache()` — they are no longer conflated in the doc-comment.

Stays in the store open/load lane only — no changes to view paging or the clear-gate (AND-632 / AND-633).

## Change (files)

- `Sources/PlaidBarCache/ReadModelCacheStore.swift` — non-throwing, no-I/O `init`; lazy `loadedRows()` that discards an undecodable file and starts empty; `clearAll` resets to empty without decoding.
- `Sources/PlaidBarCache/TransactionCacheStore.swift` — same lazy/self-heal treatment; `clearAll`/`replaceAll` reset without decoding; `existingTransactionIds(in:cacheKey:)` reads the already-loaded snapshot.
- `Sources/PlaidBarCache/{ReadModelCacheStore,TransactionCacheStore}+Factory.swift` — `onDiskIn:` / `inMemory:` inits are now non-throwing (no eager open).
- `Sources/PlaidBar/App/AppState+TransactionCache.swift`, `Sources/PlaidBar/App/AppState+ReadModelCache.swift` — open helpers drop the now-dead `do/catch`; corrected `clearReadModelCache()` doc-comment (AND-657 #2).
- `Tests/PlaidBarCacheTests/CacheStoreOpenRobustnessTests.swift` — **new suite** (6 tests).
- `Tests/PlaidBarCacheTests/{ReadModelCacheStoreTests,ReadModelCacheClearRaceTests,TransactionCacheStoreTests,TransactionCacheClearRaceTests}.swift` — mechanical `try` removal at the now-non-throwing init call sites only.

## Testing

```
swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors   # Build complete!
swift test --filter CacheStoreOpenRobustnessTests                                 # 6/6 passed
swift test --filter PlaidBarCacheTests                                            # 34/34 passed (28 pre-existing + 6 new)
git diff --check                                                                  # clean
```

New tests assert: (a) an incompatible/corrupt file reads as a miss and self-heals on the next write for **both** stores, with a reopened store round-tripping the rebuilt rows (no data loss); (b) `clearAll`/`replaceAll` win over an incompatible file without a decode failure; (c) construction performs **no** disk I/O (counted via a `FileManager` spy), with the first operation triggering the deferred load.

## Links

Closes AND-656 (remaining cache-robustness findings 2 & 3); also addresses AND-657 finding #2 doc inaccuracy.

## Gate / risk note

Behavior-preserving for the happy path (valid store round-trips unchanged). The store inits became non-throwing — all call sites (AppState helpers + existing tests) are updated in this PR. Disposable-cache contract is unchanged: any open/read/write failure still degrades to the existing cold path.

### DIFF
--- a/Sources/PlaidBar/App/AppState+ReadModelCache.swift
+++ b/Sources/PlaidBar/App/AppState+ReadModelCache.swift
@@ -66,10 +66,17 @@ extension AppState {
 
     /// Lazily opens (or re-opens) the on-disk disposable store for the active
     /// data directory. Returns `nil` — and stays disabled for this call — when
-    /// the cache is feature-disabled, when in demo mode (demo data is never
-    /// cached), or when the file-backed store fails to open. Reopens when the
-    /// active storage directory changes (e.g. sandbox↔production switch) so the
-    /// store always matches the environment whose key it is asked about.
+    /// the cache is feature-disabled or when in demo mode (demo data is never
+    /// cached). Reopens when the active storage directory changes (e.g.
+    /// sandbox↔production switch) so the store always matches the environment whose
+    /// key it is asked about.
+    ///
+    /// Opening does **no** disk I/O and cannot fail: the backing file is read and
+    /// decoded lazily on the store actor's executor, so this never blocks the
+    /// MainActor with a full decode (AND-656 finding 3). An incompatible/corrupt
+    /// file (e.g. a pre-JSON SwiftData `.store`) is self-healed into a disposable
+    /// miss on that first off-main read rather than disabling the cache (AND-656
+    /// finding 2).
     func readModelCacheStoreIfAvailable() -> ReadModelCacheStore? {
         guard readModelCacheEnabled, !isDemoMode else { return nil }
 
@@ -81,21 +88,10 @@ extension AppState {
             return existing
         }
 
-        do {
-            let store = try ReadModelCacheStore(onDiskIn: directory)
-            readModelCacheStore = store
-            readModelCacheStoreDirectoryPath = directoryPath
-            return store
-        } catch {
-            // Cache unavailable / store unopenable: behave exactly as before
-            // the cache existed. Logged without any financial material.
-            AppState.readModelCacheLogger.error(
-                "Read-model cache unavailable: \(String(describing: error), privacy: .public)"
-            )
-            readModelCacheStore = nil
-            readModelCacheStoreDirectoryPath = nil
-            return nil
-        }
+        let store = ReadModelCacheStore(onDiskIn: directory)
+        readModelCacheStore = store
+        readModelCacheStoreDirectoryPath = directoryPath
+        return store
     }
 
     /// The environment+directory-scoped key for the active context, or `nil`
@@ -168,8 +164,11 @@ extension AppState {
         }
     }
 
-    /// Wipes the disposable cache alongside the JSON/SQLite caches on local reset
-    /// and on the empty-accounts path (last institution removed).
+    /// Wipes **only the read-model store** alongside the JSON/SQLite caches on local
+    /// reset and on the empty-accounts path (last institution removed). The
+    /// per-transaction store is wiped separately by ``clearTransactionCache()``
+    /// (AND-657) — both route through the same ``ReadModelCacheClearGate`` epoch so
+    /// they clear in the same program order.
     ///
     /// Bumps the clear epoch synchronously *first* so any persist already
     /// scheduled in program order observes the clear and drops its write — the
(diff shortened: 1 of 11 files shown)


ftchvs:

**Verdict: LGTM (minor nits only).** The two AND-656 findings are correctly and narrowly addressed. An incompatible/corrupt store now self-heals into a disposable miss instead of permanently disabling the cache, and the open path does zero disk I/O so the full-history decode no longer runs synchronously on the MainActor. No real-data-loss risk found (live in-memory data stays authoritative; the on-disk store is rebuilt on the next write).

**What I verified**
- `loadedRows()` in both `ReadModelCacheStore.swift` and `TransactionCacheStore.swift` correctly: returns the memoized `[:]` for a missing file, decodes a valid snapshot once, and on a snapshot-decode failure calls `discardIncompatibleStoreFile()` (best-effort `removeItem`) + starts empty. The next `save`/`upsert` atomically overwrites the bad file → self-heal confirmed by the round-trip + reopen assertions.
- `init` no longer touches disk (`rowsByKey = nil`); the lazy-open test counts `fileExists` via a `FileManager` spy and `loadedRows()` is the only `fileExists` caller, so `== 0` at construction genuinely proves finding 3. Test is real, not a tautology.
- `clearAll`/`replaceAll` set the dict to `[:]` then `persist()` (atomic write) — they correctly win over a leftover incompatible file without decoding it, and the `replaceAll`/`clearAll`-over-garbage tests cover this.
- `existingTransactionIds(in:cacheKey:)` made `nonisolated` over a passed-in value-type snapshot is concurrency-safe; `dataGeneration`/`orderCache` invalidation is unchanged and still bumps after every write.

**Minor nits (non-blocking)**
- `ReadModelCacheStore.load(...)` doc-comment still says *"A decode failure throws so the caller's `try?` can drop back to today's cold path."* That is now only true for a corrupt **per-row payload** (still `try decoder.decode(...)`), not for a whole-store decode failure (now self-healed silently). The statement is technically still accurate but reads as if all decode failures throw — a one-line clarification would prevent a future reader misunderstanding the new split (snapshot-decode = self-heal, payload-decode = throw). No new test covers the payload-decode-throws path, but it's pre-existing behavior, so not required here.
- `CountingFileManager` is `@unchecked Sendable` with a non-atomic `fileExistsCallCount += 1`. Fine because the store is an actor and the suite is `.serialized`, so probes are serialized — but worth a comment if this spy ever gets reused outside a serialized suite.
- `discardIncompatibleStoreFile()` swallows the `removeItem` error (`try?`). Correct for the disposable contract, but if the delete fails *and* the subsequent `persist()` also fails (disk error), the in-memory `[:]` masks a still-broken file until the next process launch. Acceptable under the "cache is disposable, cold path always available" contract — just noting it's a silent degradation, not a hard failure.

**Scope:** stays in the open/load lane as claimed — no changes to paging, the clear-gate (AND-632/633), or the Keychain/server boundary. The `try` removals at the now-non-throwing init call sites are mechanical and complete (AppState helpers + all four existing test files updated). Privacy Mask / color-only-accessibility are not implicated (no financial value rendered or logged in this diff; the removed error log was already `privacy: .public` on a non-financial string).

**Must-fix: none.** Recommend a human glance at the `load(...)` doc-comment nit, then merge.

ftchvs (commented):
Reviewed this PR for the Plaid/local-first, cache-open robustness, and strict-concurrency risk areas. I did not find a serious issue requiring a code change in the reviewed scope.

Validation performed by review:
- Checked the cache-open fallback behavior against the local cache/data-dir contract.
- Checked for Plaid token/raw payload movement into app-facing surfaces.
- Checked for obvious strict-concurrency or privacy regressions in the changed path.

No merge performed.

PR #664, opened 2026-06-25
fix(cache): harden clear-gate against persist-after-clear resurrection (AND-633)

ftchvs:
## What & why

`ReadModelCacheClearGate` (the main-actor epoch gate from #576) narrowed the
persist-after-clear race for the disposable read-model and transaction caches, but
left a **two-hop residual** (AND-633, from the #624 codex review):

`persistReadModelCache()` / `persistTransactionCache()` recheck `mayCommit()` on the
**main actor**, then enqueue the store write as a **separate `await`**. A clear can
bump the epoch and enqueue `clearAll()` in the window between `mayCommit()` returning
`true` and the write reaching the store actor — so a stale write could still commit
*after* the clear, resurrecting removed-institution balances/transactions on the next
cold start. The window is ~1ms but inherent to the epoch-capture → recheck → commit
pattern, because the recheck and the commit live on different actors.

This closes the window at the **store actor**, where the check and the write are
FIFO-ordered against `clearAll()` and cannot interleave.

## Change (files)

- `Sources/PlaidBarCache/ReadModelCacheStore.swift` — adds a monotonic
  `clearGeneration` (bumped by `clearAll()`), a `currentClearGeneration()` accessor,
  and an atomic clear-gated `save(_:ifNotClearedSince:) -> Bool` that re-validates the
  captured generation as its first action on the actor and drops the write (returns
  `false`) if a clear has landed since.
- `Sources/PlaidBarCache/TransactionCacheStore.swift` — same pattern: a `clearGeneration`
  distinct from the existing order-cache `dataGeneration` (only `clearAll()` bumps it),
  `currentClearGeneration()`, and an atomic
  `replaceAll(cacheKey:transactions:ifNotClearedSince:) -> ClearGatedWriteResult`. Adds
  the `Sendable` `ClearGatedWriteResult` enum (`.committed(plan)` / `.dropped`).
- `Sources/PlaidBar/App/AppState+ReadModelCache.swift` /
  `Sources/PlaidBar/App/AppState+TransactionCache.swift` — the persist seams now capture
  `currentClearGeneration()` and pass it into the atomic store write. The main-actor
  epoch gate is retained as a cheap fast-drop (defense in depth). Doc-comments document
  the two-layer model.
- `Tests/PlaidBarCacheTests/ReadModelCacheClearRaceTests.swift` /
  `Tests/PlaidBarCacheTests/TransactionCacheClearRaceTests.swift` — 6 new concurrency
  regression tests covering the two-hop window (clear-bumped-generation drop, happy-path
  commit, non-clear write does **not** drop the persist, interleaved clear-gated writes
  never survive a terminal clear).

No changes to the server, Plaid client, or Keychain boundary. Every type stays
`Sendable`; both stores remain actor-isolated. The cache stays disposable and
fallback-safe.

## Testing (exact commands + results)

```
# Build complete! (104.11s) — strict-concurrency green, warnings-as-errors

swift test --filter "ReadModelCacheClearRaceTests|TransactionCacheClearRaceTests"
# Test run with 13 tests passed after 0.073 seconds (7 pre-existing + 6 new AND-633)

git diff --check
# clean
```

## Risk / residual

Window is **closed**, not merely bounded: the generation check and the row write are a
single store-actor hop, FIFO-ordered against `clearAll()`, so a clear that lands after
the capture always bumps the generation before the write commits and the write drops
itself. The only remaining "residual" is intentional design — the main-actor epoch gate
is kept purely as a fast-drop for the common case; correctness no longer depends on it.

## Links

Closes AND-633


ftchvs:

**Verdict: minor nits.** The store-actor `clearGeneration` guard is concurrency-correct, stays in scope, keeps everything `Sendable`/actor-isolated, and does narrow the two-hop persist-after-clear window with no deadlock, priority inversion, or dropping of legitimate writes. My concerns are about *honesty of the claim* and *test fidelity*, not a correctness regression.

**Findings**

- **The window is narrowed, not "closed" as the PR body claims.** The persist seam (`AppState+ReadModelCache.swift` ~L182-185, `AppState+TransactionCache.swift` ~L118-127) does two *separate* awaits: `let g = await store.currentClearGeneration()` then `await store.save(model, ifNotClearedSince: g)`. The store-actor guard only protects the gap *between those two store-actor hops*. A clear can still land before `currentClearGeneration()` is even called (after the main-actor `mayCommit`), and program-order "clear-requested-after-persist must win" is still enforced by the **main-actor epoch**, not the generation. So correctness still depends on the epoch gate — the PR's "correctness no longer depends on it" / "Window is closed, not merely bounded" overstates it. The generation guard is a *defense-in-depth narrowing of the residual*, which is a fine thing to ship — just label it as such. The doc-comments in the store files are actually honest ("second guard", "fast-drop"); the PR description is the part that oversells.

- **Tests don't exercise a real two-hop interleaving — they stage it sequentially.** `saveDroppedWhenClearRacedAfterGenerationCapture` and the transaction analog call `save` → `clearAll` → `save(ifNotClearedSince: stale)` strictly sequentially; they validate the generation-comparison *logic*, but never reproduce a clear landing *between* `currentClearGeneration()` and `save` under concurrency. The "interleaved…never survive a terminal clear" stress tests fan out saves that each capture-and-save in one closure with **no clear interleaved during the fan-out**, then a terminal clear — so they exercise the pre-existing store-actor FIFO backstop (which #576 already had), not the new between-hops window. Net: the new guard's actual race is asserted only by construction, not by a failing-without-the-fix concurrency test. Consider a test that holds the actor between the two hops (or drives `currentClearGeneration` and `clearAll` from competing tasks) to prove the guard fires under a genuine race.

- **Minor: `ifNotClearedSince` is captured fresh inside the detached task, not at schedule time.** This is correct for the residual-narrowing purpose, but means the generation token carries *no* program-order information — worth a one-line comment so a future reader doesn't assume it does the epoch's job.

**Non-issues verified:** `Sendable` `ClearGatedWriteResult` is fine; the two `replaceAll` overloads are unambiguous and both `@discardableResult`; `clearGeneration` is correctly distinct from `dataGeneration` (only `clearAll()` bumps it — the `ordinaryWriteDoesNotDropClearGatedPersist` test pins this well); `&+=` wrap is harmless (UInt64, equality compare). No Privacy Mask / color-only / financial-value surfaces touched. Scope is clean (cache only; server/Plaid/Keychain untouched).

**Must-fix:** none. **Should-fix before merge:** soften the "window closed / correctness no longer depends on the epoch" language in the PR body to match the (honest) in-code docs, and add one test that actually races the two store hops against a clear.

ftchvs (commented):
Reviewed and pushed a fix in `2f4df81`.

Finding fixed: the hardening still captured the cache store generation after the main-actor clear epoch check. If a clear completed between those two points, a stale persist could capture the post-clear generation and commit after the clear.

Changes made:
- Added a second main-actor clear-epoch check immediately after store-generation capture and before each cache write.
- Covered both read-model and transaction cache persist paths.
- Added a source-level regression test that requires the post-generation-capture epoch recheck.

Validation:
- `git diff --check`
- `swift test --filter cachePersistsRecheckClearEpochAfterGenerationCapture`

No merge performed.

PR #665, opened 2026-06-25
refactor(window): per-window state on NavigationModel, retire single-surface flags (AND-600)

ftchvs:
## What & why

The menu-bar popover's presentation flag (`isPopoverPresented`) and the detached-dashboard intent (`isDashboardDetached`) lived on the app-global `AppState`. Both are **single-surface by construction** — a global property of the whole app can only ever describe one surface, which is the exact blocker to multiple independent windows.

This migrates both off `AppState` onto the **per-window `NavigationModel`** (the value-type `NavigationState` underneath), so two simultaneous windows hold independent presentation and detach state. The menu bar still owns exactly one `NavigationModel`, so the legacy popover / `--window-first off` escape-hatch path behaves byte-identically.

Stage-2 cleanup, **SAFE non-deleting part only** (AND-600). Out of scope and untouched (gated on on-hardware parity QA, AND-615): `PopoverWindowAnchor`, `AppActivationPolicyCoordinator`, the legacy `NSWindowController`s, and the `--window-first` escape hatch (AND-598/599).

## Change (files)

- `Sources/PlaidBarCore/Models/NavigationState.swift` — add `isPopoverPresented` (ephemeral) + `isDashboardDetached` (durable intent) with pure transitions (`setPopoverPresented`, `detachDashboard` — sets intent **and** dismisses popover so only one surface is up, `redockDashboard`). Codable round-trip preserved.
- `Sources/PlaidBar/App/NavigationModel.swift` — façade for both; persists the detached intent to the **unchanged** `dashboard.detached` key and applies the snapshot-render override (`DetachedDashboardPreferences.resolvedDetachedIntent`) at hydration. `isPopoverPresented` is deliberately not hydrated (a relaunch never auto-opens the popover).
- `Sources/PlaidBar/App/AppState.swift` — remove the two stored properties + the `dashboard.detached` key + the `loadSettings` detached block; thread the snapshot flag into the model at construction.
- `Sources/PlaidBar/App/PlaidBarApp.swift` — all reads/writes go through `appState.navigationModel.*`; the `menuBarExtraAccess(isPresented:)` binding is built explicitly off the `@Observable` model (a `let` `navigationModel` can't carry a writable key-path through the `@State` projection).
- `Sources/PlaidBar/App/DetachedDashboardCoordinator.swift` — `detach`/`redock`/`sync`/`handleMenuBarActivation`/`presentForLaunchOverride` drive the per-window model (via `detachDashboard()` / `redockDashboard()`).
- `Sources/PlaidBar/App/SnapshotRenderer.swift`, `Sources/PlaidBar/Settings/SettingsView.swift` — updated call sites; Settings binds via `@Bindable var nav = appState.navigationModel`.
- `Sources/PlaidBar/App/DetachedDashboardWindowController.swift`, `Sources/PlaidBarCore/Models/DetachedDashboardPreferences.swift` — doc-comment accuracy only.
- `Tests/PlaidBarCoreTests/NavigationSurfaceStateTests.swift` — **new suite** (9 tests).

No references to the single-surface flags remain on `AppState` (only doc-comments referencing the migration). Persistence footprint is unchanged: same `dashboard.detached` key, same raw value.

## Testing

```
swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors   # exit 0
swift test --filter NavigationSurfaceStateTests                                  # 9 passed
swift test                                                                       # 2192 passed, 0 failures
git diff --check                                                                 # clean
```

The new suite pins the migration's reason for existing — per-window independence: detaching/presenting one window's surface never touches another's, and `detachDashboard()` dismisses only its own popover. Existing `RouteNavigationTests` / `NavigationStateIndependenceTests` / `DetachedDashboardPreferencesTests` (50 tests) still pass.

> Gate note: the legacy popover / escape-hatch (`--window-first off`) build still compiles and works — the menu bar owns one `NavigationModel`, so detach/redock, the status-item click intercept, and on-disk persistence are byte-identical.

## Links

Closes AND-600

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -57,7 +57,6 @@ final class AppState {
         static let firstRunSnapshotDismissedContextPrefix = "firstRunSnapshot.dismissed.context"
         static let lastTransactionCacheContext = "cache.lastTransactionCacheContext"
         static let categoryBudgetCache = "cache.categoryBudgets"
-        static let dashboardDetached = DetachedDashboardPreferences.detachedStorageKey
     }
 
     // MARK: - State
@@ -115,7 +114,6 @@ final class AppState {
             self.error = sanitized
         }
     }
-    var isPopoverPresented = false
 
     /// Per-window navigation model (AND-594). Owns the destination plus
     /// the dashboard filter / account selection / heatmap metric that previously
@@ -127,7 +125,10 @@ final class AppState {
     /// `appState.dashboardFilter` / `dashboardSelectedAccountID` /
     /// `dashboardHeatmapMode`, which delegate here and persist to the original
     /// UserDefaults keys — so popover behavior and on-disk persistence are
-    /// unchanged.
+    /// unchanged. As of AND-600 it also owns the menu-bar popover's presentation
+    /// flag (`isPopoverPresented`) and the detached-dashboard intent
+    /// (`isDashboardDetached`), retired from `AppState` so each window holds
+    /// independent presentation and detach state.
     let navigationModel: NavigationModel
 
     /// Pending deep-link route awaiting the primary window (AND-597).
@@ -163,19 +164,14 @@ final class AppState {
         set { navigationModel.heatmapMode = newValue }
     }
 
-    /// When true, the dashboard lives in a floating desktop window instead of
-    /// the menu-bar popover (AND-384). Persisted so the window reopens on the
-    /// next launch. While detached, a click on the menu-bar item raises the
-    /// floating window rather than opening the popover; re-docking flips this
-    /// back to false and the popover resumes. Mirrors the `dashboard.detached`
-    /// `@AppStorage` key the Settings toggle writes, so the toggle and the
-    /// in-dashboard pin/re-dock controls stay in sync.
-    var isDashboardDetached = false {
-        didSet {
-            guard isDashboardDetached != oldValue else { return }
-            UserDefaults.standard.set(isDashboardDetached, forKey: Keys.dashboardDetached)
-        }
-    }
+    // AND-600: the menu-bar popover's presentation flag (`isPopoverPresented`) and
+    // the detached-dashboard intent (`isDashboardDetached`) no longer live on
+    // `AppState` as single-surface app-global storage. They moved onto the
+    // per-window ``NavigationModel`` (`appState.navigationModel.isPopoverPresented`
+    // / `.isDashboardDetached`), so a second window holds independent presentation
+    // and detach state. The legacy popover/escape-hatch path drives the menu bar's
+    // single model exactly as before; persistence of the detached intent is
+    // unchanged (still the `dashboard.detached` key, now owned by the model).
 
     /// Persisted across launches so configured installs boot straight into
     /// the dashboard instead of flashing first-run onboarding until the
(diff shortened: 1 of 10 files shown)


ftchvs:

**Verdict: LGTM** — clean, faithful, well-scoped migration. No correctness or regression concerns found; only minor coverage/parity notes below. Not approving; leaving open for human review.

### Scope (gate compliance) ✅
- `PopoverWindowAnchor`, `AppActivationPolicyCoordinator`, the legacy `NSWindowController`s, and the `--window-first` escape hatch are **untouched** — confirmed via `git diff --stat` (only the 10 declared files changed). This is correctly the SAFE non-deleting part; AND-615/598/599 deletions are not pulled forward.
- No single-surface flags remain on `AppState`: the stored `isPopoverPresented`/`isDashboardDetached` properties and the `Keys.dashboardDetached` entry are removed; all surviving references are doc-comments or the intended `navigationModel.*` / `$nav.isDashboardDetached` call sites (`git grep` verified).

### Per-window independence (the migration's reason for existing) ✅
- State now lives on the value-type `NavigationState`; `NavigationModel` is per-window (`private(set) var state`, `@Observable @MainActor`), so two windows hold independent presentation/detach state.
- `detachDashboard()` dismisses only its own popover (`isPopoverPresented = false` on the same struct), and the new tests pin exactly the cross-window contract that the old global flag could never express — incl. `detachDoesNotCrossDismissPopover`. Good adversarial framing.

### Behavioral parity ✅
- `detachDashboard()`/`redockDashboard()` mutate `state` directly then persist; the `isDashboardDetached` setter keeps the `!= newValue` guard that the old `AppState.didSet` had — same write-back behavior, same `dashboard.detached` key/raw value.
- Snapshot/window-first determinism is preserved: the `isRenderingSnapshot || isRenderingWindowFirst` predicate now threads into `NavigationModel(isRenderingSnapshot:)` and is applied via the *same* `DetachedDashboardPreferences.resolvedDetachedIntent` at hydration. `loadSettings` no longer re-resolves (verified it no longer touches the intent), so no double-resolution/ordering bug.
- `isPopoverPresented` is intentionally not hydrated (relaunch never auto-opens) — matches prior behavior.
- `popoverPresentedBinding` is a correct manual `Binding` workaround for the `let navigationModel` (can't project a writable key-path through a `let`); equivalent to the old `$appState.isPopoverPresented`.
- Strict-concurrency: no new types, no Sendable surface changed. Privacy Mask / color-only: N/A — these are presentation booleans, no financial values touched.

### Minor notes (non-blocking)
- **Injected-model + snapshot edge:** when a test injects a `NavigationModel` into `AppState`, the snapshot override is *not* re-applied (the injected model already hydrated with its own default `isRenderingSnapshot: false`), because resolution moved out of `loadSettings`. Production never injects, so this is harmless — but it's a subtle semantic shift from "always resolved in loadSettings" worth a one-line comment at the injection param.
- **Test altitude:** the new suite (and pre-existing `DetachedDashboardPreferencesTests`) cover the pure transitions and the resolver thoroughly, but nothing exercises the `NavigationModel` wrapper's `persistDashboardDetached` write-back / `isHydrating` guard or model-level hydration of the detached intent. That's the code that actually moved. The app target isn't in the test binary, so this is expected — flagging only so a reviewer knows the wrapper persistence path is covered transitively, not directly.

### Must-fix
None.

ftchvs:
Reviewed and pushed a test-target compile fix in `c374d70`.

Finding fixed:
- `NavigationSurfaceStateTests` used `JSONEncoder`/`JSONDecoder` without importing `Foundation`, which breaks the test target on a clean compile.
- The same file also had a `var` that was never mutated; strict warnings-as-errors would fail on that warning, so I changed it to `let` while keeping the test behavior unchanged.

Validation:
- `git diff --check`
- `swift test --filter NavigationSurfaceStateTests` (9 Swift Testing tests passed)
- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors`

Docs context checked: Context7 Swift strict-concurrency guidance. I did not merge the PR.

PR #666, opened 2026-06-25
feat(export): local CSV/JSON export for transactions, accounts, budgets (AND-645)

ftchvs:

## What & why

Settings → Local data already has an **Export & Backup** section (AND-492) that writes accounts, transactions, and balance history to local CSV/JSON via an NSSavePanel/NSOpenPanel — no upload, fits the local-first model. AND-645 completes that scope cut by adding the user's **saved category budgets** to the same export, so a portable backup matches what the user actually sees in-app (Budgets).

## Change

- **`PlaidBarCore/DataExportBuilder` (pure, Sendable):**
  - `budgetsCSV(_:)` — emits `category,category_name,monthly_limit` with rows **sorted by Plaid category raw value**, so output is deterministic regardless of the app's `[SpendingCategory: Double]` dictionary order. Reuses the existing RFC-4180 escaping + spreadsheet-formula-injection neutralization (`csvField`). A `[SpendingCategory: Double]` overload + `budgetDTOs(from:)` normalize the app's in-memory map.
  - JSON `Envelope` gains an optional `budgets: [CategoryBudgetDTO]` + `counts.budgets`; `schemaVersion` bumped **1 → 2**. Custom `init(from:)` decoders make both new fields decode as **empty/zero from a v1 file**, so older backups stay importable.
  - `isExportAllowed(shouldMaskFinancialValues:)` — the masked-state gating rule, lifted out of the SwiftUI view so it is **unit-testable**. Export writes real values to disk, so it must stay blocked while the UI is masked (Privacy Mask) or locked (App Lock); both surface through `shouldMaskFinancialValues`.
- **App (`Settings/SettingsView.swift`):** CSV export now also writes `budgets.csv`; JSON export includes budgets; `isExportDisabled` delegates to the Core decision; section + masked-state copy updated to mention budgets. The app target still only touches local DTOs it already holds — no new server/credential/token access.

## Testing

Ran with `dangerouslyDisableSandbox` (SwiftPM needs it locally).

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete!** (clean, no warnings-as-errors).
- `swift test --filter DataExportBuilderBudgetsTests` → **9 tests passed** (stable budget CSV order, escaping/neutralization, empty input, JSON envelope with/without budgets, v1 backward-compat decode, masked-gating decision).
- `swift test --filter DataExportBuilderTests --filter DataExportBuilderOverrideTests` → **15 tests passed** (existing export suites — envelope round-trip stays green after the schema bump).
- `git diff --check` → clean.

## Links

Closes AND-645

## Open questions / risks

- CSV uses the documented `category` raw value + `category_name`; the `monthly_limit` column has no currency column because budgets are always the account's local currency in this app — flag if a currency column is wanted.
- Privacy Mask currently **disables** export entirely (the existing AND-492 hard-gate) rather than offering an "export real values anyway" confirmation; this is the stricter, safer interpretation of the requirement and matches the existing behavior. Can switch to a confirmation dialog if preferred.
- No import path yet — this is export-only, as scoped. The JSON envelope is intentionally round-trippable for a future importer.

### DIFF
--- a/Sources/PlaidBar/Settings/SettingsView.swift
+++ b/Sources/PlaidBar/Settings/SettingsView.swift
@@ -731,7 +731,7 @@ struct GeneralSettingsView: View {
 
             Section("Export & Backup") {
                 VStack(alignment: .leading, spacing: Spacing.sm) {
-                    Text("Export a portable copy of your accounts, transactions, and balance history. VaultPeek never uploads these files — they are written only to the folder you choose. Pick a local folder to keep them on this Mac; a synced location (iCloud Drive, Dropbox, a network share) will copy them off-device.")
+                    Text("Export a portable copy of your accounts, transactions, budgets, and balance history. VaultPeek never uploads these files — they are written only to the folder you choose. Pick a local folder to keep them on this Mac; a synced location (iCloud Drive, Dropbox, a network share) will copy them off-device.")
                         .detailText()
                         .fixedSize(horizontal: false, vertical: true)
 
@@ -750,7 +750,7 @@ struct GeneralSettingsView: View {
                     if appState.shouldMaskFinancialValues {
                         Label(
                             appState.isContentLocked
-                                ? "Export is disabled while App Lock is locked, so real balances, accounts, and transactions are never written to disk. Unlock VaultPeek to export."
+                                ? "Export is disabled while App Lock is locked, so real balances, accounts, transactions, and budgets are never written to disk. Unlock VaultPeek to export."
                                 : "Export is disabled while Privacy Mask is on, so real balances are never written to disk. Turn off Privacy Mask to export.",
                             systemImage: appState.isContentLocked ? "lock" : "eye.slash"
                         )
@@ -815,7 +815,7 @@ struct GeneralSettingsView: View {
         NSWorkspace.shared.activateFileViewerSelecting([url])
     }
 
-    // MARK: - Export & Backup (AND-492)
+    // MARK: - Export & Backup (AND-492, budgets added AND-645)
 
     private var isExportDisabled: Bool {
         // Gate on the effective mask state, not just the manual Privacy Mask
@@ -823,8 +823,11 @@ struct GeneralSettingsView: View {
         // active. Settings is a separate scene reachable from the status-item
         // menu while the dashboard is locked, so without this a locked session
         // (Privacy Mask off) could still write the full accounts/transactions/
-        // balances export to disk, bypassing App Lock (Codex P1).
-        appState.shouldMaskFinancialValues
+        // balances/budgets export to disk, bypassing App Lock (Codex P1). The
+        // rule itself lives in `DataExportBuilder` so it is unit-testable.
+        !DataExportBuilder.isExportAllowed(
+            shouldMaskFinancialValues: appState.shouldMaskFinancialValues
+        )
     }
 
     /// Documented backup file path, e.g. `~/.vaultpeek/plaidbar-sandbox.sqlite`.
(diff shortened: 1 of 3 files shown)


ftchvs:

**Verdict: LGTM (minor)** — correct, well-scoped, privacy-safe. No must-fix items. A couple of optional polish notes below.

I built and ran the suites locally (`dangerouslyDisableSandbox`):
- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete!** (clean)
- `DataExportBuilderBudgetsTests` (9), `DataExportBuilderTests` (9), `DataExportBuilderOverrideTests` (6) → **all green**

### What I verified
- **Serializers are pure + in `PlaidBarCore` + tested.** `budgetsCSV`/`budgetDTOs`/`isExportAllowed` are `static` on the `enum DataExportBuilder` (no AppKit/I/O), reuse the existing RFC-4180 `csvField` escaping + formula-injection neutralization, and are covered by the new suite. Stable sort by `category.rawValue` makes output deterministic regardless of `[SpendingCategory: Double]` dictionary order. Verified the test fixtures against real raw values/displayNames in `SpendingCategory.swift` (`FOOD_AND_DRINK < GENERAL_MERCHANDISE < RENT_AND_UTILITIES < TRANSPORTATION`, "Bills & Utilities", etc.) — assertions are accurate.
- **Local-only.** Export still goes through the existing NSSavePanel/folder-write path; the app target only touches local DTOs it already holds. No new server/credential/token access. No upload.
- **Privacy Mask / App Lock honored.** Gating is now `!DataExportBuilder.isExportAllowed(shouldMaskFinancialValues:)`. `shouldMaskFinancialValues` is `financialPrivacyDisplayMode != .normal`, which is `true` for *both* Privacy Mask and App Lock — so a locked-but-unmasked Settings scene can no longer bypass the gate and write real budgets/balances to disk. The PR keeps the stricter "fully block while masked" stance (no silent "export anyway") — correct, safe interpretation.
- **No new ID/token leak.** Budgets CSV/JSON emit only `category`/`category_name`/`monthly_limit`. The pre-existing `account_id`/`item_id`/`mask` columns in `accountsCSV` are unchanged AND-492 scope, not introduced here. No secrets reach the app/Core.
- **Schema bump is backward-safe.** `schemaVersion 1→2`; custom `init(from:)` on `Envelope` + `Counts` use `decodeIfPresent ?? []/0`, so v1 backups decode cleanly (test covers it). Existing test asserts against `DataExportBuilder.schemaVersion` rather than a literal `1`, so the bump doesn't regress it. New `budgets:` params are all labeled and defaulted — no positional-arg breakage at existing call sites (full app build confirms).
- **Sendable / strict-concurrency:** clean. **Color-alone:** N/A — masked-state Label pairs text with `lock`/`eye.slash` SF Symbols, no color-only meaning.

### Minor (optional, non-blocking)
1. **Escaping test doesn't exercise `budgetsCSV` end-to-end.** `budgetsCSVEscapingAndNeutralization` calls `csvField` directly and (correctly) notes displayNames are constants so no real budget name can inject. Fair — but since `category_name` is the only free-ish text column, a one-line assertion that a `budgetsCSV([...])` row for a category whose displayName contained `&` stays literal would lock the wiring in. As-is the row-builder→`csvField` path is only proven transitively. Cosmetic.
2. **`decimalString` locale:** confirmed Swift's `String(format:"%.2f")` is locale-independent (emits `450.50` even under `LC_ALL=de_DE`), so no `,`-as-decimal CSV-corruption risk. Noting it only because it's the kind of thing worth a guard-comment if anyone later swaps in a `NumberFormatter`.

ftchvs (commented):
Review complete. No serious findings.

I checked the local export/privacy boundary: export remains user-chosen and local, the app target still does not call Plaid directly, CSV formula injection is neutralized, and masked financial values prevent export.

Validation reviewed:
- `git diff --check origin/main...origin/pr/666`
- strict `PlaidBar` build
- strict `PlaidBarServer` build
- `swift test --filter DataExportBuilder --skip-update --disable-keychain` (24 selected tests)

No merge performed.

PR #667, opened 2026-06-25
feat(settings): theme + accent color customization (AND-647)

ftchvs:

## What & why

AND-647 (post-1.0 deferred candidate, "Dark/light theme customization") asked to extend the existing appearance preference (follow-system / light / dark via `AppAppearance`) with user-facing theme customization — at minimum an accent color — surfaced in Settings, while keeping the single authoritative `NSApp.appearance` owner and the accessibility rule that meaning is never conveyed by color alone.

This ships a user-selectable **accent color** (System + 8 named accents). It is decorative/brand only: it tints hero glyphs, active controls, selection washes, and the app-wide `.tint` — it never carries financial or status meaning.

## Change

- **PlaidBarCore (pure, Sendable):** new `AppAccentColor` enum + `AppAccentSwatch` in `Sources/PlaidBarCore/Models/AppearancePreferences.swift`, alongside the existing appearance enums. Provides `storageKey` (`appearance.accentColor`), `defaultValue` (`.system`), `title`, per-case `swatch` (sRGB triple, `nil` for System), and a pure `resolvedSwatch(cliOverride:storedAccent:)` resolver — the override-wins precedence mirrors `AppAppearanceMode.resolvedScheme`.
- **App bridge:** `SemanticColors.resolvedAccent(for:)` + `Color(accentSwatch:)` in `Theme/DesignTokens.swift` are the single place the Core choice meets SwiftUI (parity with the existing `ForcedColorScheme → ColorScheme` bridge). `SemanticColors.brand` now resolves to `Color.accentColor`, so every brand surface re-tints live.
- **Single appearance owner preserved:** the accent is applied as SwiftUI `.tint(...)` only, via a new `appliesAppAccent()` modifier (`AppAccentApplier`) that mirrors `appliesAppTextSize()`, wired at every scene/window root (glance, popover, Settings, primary Window, and the three detached AppKit-hosted windows). No new `NSApp.appearance` writer is introduced — `AppAppearance.applyToNSApp` remains the sole owner. "System" applies no tint override (inherits the macOS accent). A `--accent` CLI flag (QA/screenshot aid) wins over the stored choice, mirroring `--appearance`.
- **Settings UI:** accent `Picker` in the Display section — every row labels the accent by **name** with a swatch icon (never color alone), plus an accessible `AccentSwatchPreviewRow` (swatch + name caption + a sample "Active" control with a checkmark glyph). Explanatory copy states the accent is decorative and never changes over/under-budget, gain/loss, or status colors.

Accessibility / privacy: meaning never depends on the chosen accent — over/under budget, gain/loss, currency, and sync status keep their own semantic colors and non-color cues. The Settings preview uses synthetic labels only (no real financial data), so it is Privacy-Mask / App-Lock safe.

## Testing

Run from the worktree with the sandbox disabled.

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete!** (clean, Swift 6 strict concurrency, warnings-as-errors).
- `swift test --filter AppAccentColorTests` → **11 tests passed** (pure preference → swatch mapping, CLI-override precedence, in-gamut + distinct swatches, defaults, titles, identifiers, storage-key namespacing).
- `swift test --filter AppearancePreferencesTests` → **12 tests passed** (no regression in the sibling appearance enums).
- `git diff --check` → clean (no whitespace errors).

Not run: `./Scripts/smoke-sandbox.sh` (app/Core-only change; no server surface touched).

## Links

Closes AND-647.

## Open questions / risks

- The 8 accent swatches are hand-tuned sRGB triples that track Apple's system palette; they are decorative only, so exact hue is not load-bearing. A future pass could swap to a `Color` asset catalog for automatic light/dark variants if desired.
- The monochrome menu-bar **label** glyph is deliberately left untinted to avoid fighting the menu bar's template-image rendering; the accent tints content surfaces only.
- `SemanticColors.brand` moving from a fixed `.systemBlue` to `Color.accentColor` means the default install now follows the user's macOS accent (previously always blue). This is the intended "System" default and matches platform convention, but is a visible default-appearance change for existing users.

### DIFF
--- a/Sources/PlaidBar/App/CategoryDashboardWindowController.swift
+++ b/Sources/PlaidBar/App/CategoryDashboardWindowController.swift
@@ -181,6 +181,8 @@ final class CategoryDashboardWindowController: NSObject, NSWindowDelegate {
                 // Category Dashboard scales with the same control as the popover
                 // (AND-570). Reads @AppStorage, so it tracks live changes.
                 .appliesAppTextSize()
+                // Tint with the same accent preference as every other surface (AND-647).
+                .appliesAppAccent()
                 .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .topLeading)
         )
     }
--- a/Sources/PlaidBar/App/DetachedDashboardWindowController.swift
+++ b/Sources/PlaidBar/App/DetachedDashboardWindowController.swift
@@ -420,6 +420,9 @@ final class DetachedDashboardWindowController: NSObject, NSWindowDelegate {
                 // Apply the in-app text-size preference so the floating desktop
                 // dashboard scales with the same control as the popover (AND-570).
                 .appliesAppTextSize()
+                // Tint the floating desktop dashboard with the same accent
+                // preference as every other surface (AND-647).
+                .appliesAppAccent()
                 // The panel owns its own width via resize; let the dashboard fill
                 // it rather than imposing the popover's screen-anchored width math.
                 .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .topLeading)
--- a/Sources/PlaidBar/App/PlaidBarApp.swift
+++ b/Sources/PlaidBar/App/PlaidBarApp.swift
@@ -137,6 +137,7 @@ struct PlaidBarApp: App {
                 .forcedAppColorScheme(Self.forcedColorScheme)
                 .appliesAppAppearance()
                 .appliesAppTextSize()
+                .appliesAppAccent()
         } else {
             MainPopover()
                 .environment(appState)
@@ -176,6 +177,7 @@ struct PlaidBarApp: App {
                 // so this control is the only way users can enlarge VaultPeek's
                 // text (AND-570).
                 .appliesAppTextSize()
+                .appliesAppAccent()
         }
     }
 
@@ -497,6 +499,9 @@ struct PlaidBarApp: App {
                 // Scale the Settings window too, so the user sees the text-size
                 // change reflected in the very control they are adjusting (AND-570).
                 .appliesAppTextSize()
+                // Tint Settings too, so the accent picker's effect is visible in
+                // the very window the user is adjusting (AND-647).
+                .appliesAppAccent()
         }
 
         // Window-first primary workspace (Epic 1 / AND-591). The scene is
@@ -527,6 +532,7 @@ struct PlaidBarApp: App {
                 // competing setter to flash a wrong first-paint theme.
                 .appliesAppAppearance()
                 .appliesAppTextSize()
+                .appliesAppAccent()
                 // Elevate `.accessory → .regular` while this window is on screen and
                 // drop back when the last managed window closes, via the shared
                 // refcounted coordinator (AND-620). SwiftUI gives a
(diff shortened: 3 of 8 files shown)


ftchvs:

**Verdict: LGTM (minor).** The architecture is sound and the headline risks are clean. The single `NSApp.appearance` owner is preserved — `AppAppearance.applyToNSApp` stays the sole writer and the accent ships purely as SwiftUI `.tint(...)` via `AppAccentApplier`, so there is no second appearance writer and no first-paint theme flash. The never-meaning-by-color-alone rule holds: I traced all 61 `SemanticColors.brand` call sites; `brand` is consistently the *neutral / info / in-progress* tone, while genuine semantics (negative / warning / positive) live on their own fixed colors, each already paired with a glyph + text (e.g. `AlertsDestinationView.accent`, `InsightsAIInsightView.phaseTint`/`phaseGlyph`/`statusLabel`). Changing the accent cannot flip the read of a value. Persistence mirrors the existing `@AppStorage(storageKey)` appearance pattern, and `--accent` mirrors `--text-size`/`--appearance` (override-wins, parsed in Core via the pure `resolvedSwatch`). Core enum is `Sendable`, 11 unit tests cover resolution/precedence/gamut.

**Findings (advisory, non-blocking):**

1. **`brandSecondary` collision is real for two accents.** `brandSecondary` stays `.systemOrange`, and the doc-comment claims it is held fixed "so the sandbox-mode indicator does not collide with a warm primary accent." But the user can now pick **Orange** (`brand` → orange triple) or **Red** for `brand`, which sit right next to `brandSecondary`'s orange. Where both render in the same surface (sandbox-mode hero vs brand hero) they may become near-indistinguishable. This is decorative-only so it is not a correctness bug, but the comment's premise (that holding `brandSecondary` fixed *prevents* collision) is now false for the orange/red accents.

2. **`graphite` accent collapses active-vs-secondary affordance.** `graphite` (≈0.557 gray) is close to `Color.secondary`/`.separator`. Selection washes that lean on `brand.opacity(0.10–0.16)` (e.g. `GoalsDestinationView`, `InsightsAIInsightView` chips, `AlertsDestinationView`) will read as a faint gray over neutral chrome. Selection is still carried by checkmark glyph + weight in the chips I checked, so it does not violate color-alone — but the "active control" affordance gets weak. Worth a visual QA pass, especially under Increase Contrast.

3. **Increase Contrast does not touch the accent.** `increasedContrast` is computed in `AppearanceSettingsView` but only feeds the preview card stroke; the resolved accent swatch is contrast-agnostic. The hand-tuned sRGB triples are not adjusted when Increase Contrast is on, so a low-contrast accent (graphite, teal) against glass can fall below the affordance the user expects from that setting. Not a regression (System accent had the same property), but the PR's contrast claim is about semantic colors, not the new decorative tint.

4. **App-layer bridge is untested.** Tests cover the pure Core mapping only. `Color(accentSwatch:)`, `SemanticColors.resolvedAccent(for:)`, and the `.tint` wiring at the 8 scene/window roots have no coverage. A dropped `.appliesAppAccent()` on one of the detached AppKit windows would silently leave that window untinted with no failing test. Acceptable (view code), but the "tests exercise the change" bar is met for Core only.

5. **Hardcoded sRGB triples vs asset catalog (already noted in PR).** Fixed triples don't get automatic light/dark variants; pink/red (`1.0,0.176,0.333` / `1.0,0.231,0.188`) are close enough that on a glass wash at low opacity they may be hard to tell apart by name-blind users — though every picker row carries the name, so this is cosmetic.

**Privacy / security:** Clean. `AccentSwatchPreviewRow` uses synthetic labels only (Privacy-Mask / App-Lock safe); no financial value is tinted by the accent; no secrets touched; change is app + Core only, server untouched.

**Must-fix:** none. **Recommend before merge:** a quick on-hardware visual QA of the `graphite` and `orange` accents (findings 1–2) under both light/dark and Increase Contrast, since those are the two accents most likely to weaken an affordance. Do not block on it.

ftchvs (commented):
Review complete. No serious findings.

I checked the accent/theme changes against the Plaid/privacy boundaries and the HIG color guidance: the accent remains decorative/tinting, does not move Plaid tokens or local data, and does not make finance/risk/error meaning depend on color alone.

Validation reviewed:
- `git diff --check`
- diff-only sensitive-data scan
- `swift build --target PlaidBar --skip-update --disable-keychain`
- `swift test --filter AppAccentColorTests --skip-update --disable-keychain` (11 tests)
- strict `PlaidBar` build with complete concurrency and warnings-as-errors

No merge performed.

PR #668, opened 2026-06-25
feat(currency): multi-currency account display + per-currency aggregates (AND-643)

ftchvs:
**DEFERRED feature (re-opens scope cut from v1).** From the post-1.0 deferred product candidates list (AND-643, priority Low, "Provider & Platform Expansion"). The maintainer explicitly authorized building it. **Opened as DRAFT** — it is large and carries one open product decision (the conversion source). The native-display + per-currency grouping path is complete and tested; cross-currency conversion ships as a tested offline stub pending a product call.

## What & why

VaultPeek currently assumes every balance and transaction is USD. Plaid actually reports a per-balance `iso_currency_code` (and an `unofficial_currency_code` for crypto/points), so a user with a EUR or GBP account sees their balance silently rendered with a `$`. This adds a real currency model so:

- accounts and transactions display in their **native** currency, and
- aggregates (net worth, and the same math safe-to-spend rides on) group **per currency**, converting to a single headline only when rates are available — never fabricating a number, and never collapsing two currencies into one.

This keeps the no-cloud posture: there is **no live FX fetch**. Conversion is a pluggable source whose only shipped implementation is an offline/stub rate table.

## Change

All new logic lives in `PlaidBarCore` (pure + `Sendable`); the app/server boundary is untouched.

- **`Models/CurrencyCode.swift`** — normalized ISO/unofficial code with an explicit `.unknown` bucket (an absent code is its own group, never coerced to USD), `Codable`/`Comparable`, plus a `$`-or-code `symbolHint` and a spoken `accessibleName` so currency identity is carried by **text/symbol, never color** (ACCESSIBILITY.md).
- **`Utilities/CurrencyAggregation.swift`** — sums `(amount, currency)` entries into per-currency `Subtotal`s and a best-effort `ConvertedTotal` (`.exact` / `.converted(unpricedCurrencies:)` / `.unavailable`). Native subtotals are **always** preserved. Defines `CurrencyConversionSource` (protocol), `NoConversionSource` (default — identity only, so mixed currencies degrade to subtotals-only), and `StaticCurrencyConversionRates` (offline USD-based stub table, with a `sampleOffline` set). A converted headline is only produced when a *foreign* currency was actually priced — identity-only matches with foreign balances left over degrade to `.unavailable` rather than silently dropping them.
- **`Utilities/MultiCurrencyBalancePresentation.swift`** — `netWorth(accounts:)` (debt signed negative, currencies kept separate), formatted subtotal rows, and an honest `Headline` disclosure (exact / "approximate, converted using offline rates, excludes X" / "shown per currency below"). Respects Privacy Mask: figures hidden, **currency identity stays visible**.
- **`Formatters.currency(_:in:format:)`** — native-currency formatting; bare number (no symbol) for `.unknown`.
- **`BalanceDTO`/`TransactionDTO`** — `currency` accessor wrapping the existing raw `iso_currency_code` (no schema change).
- **Server (`AccountRoutes`)** — falls back to `unofficial_currency_code` when the ISO code is absent so crypto/points balances carry a native identity. (Server keeps owning Plaid decode; the app still only sees the DTO.)

Wiring these presenters into specific views (Dashboard hero, account rows, transaction rows) is intentionally **not** in this PR — that depends on the conversion-source decision below.

## Testing

```
# → Build complete!  (Swift 6 strict-concurrency gate, warnings-as-errors)

swift test --filter MultiCurrencyTests
# → Test run with 20 tests passed after 0.012 seconds.

git diff --check
# → clean (no whitespace errors)
```

The 20 tests cover: code normalization + unknown-≠-USD invariant, symbol/accessible-name cues, `Codable` round-trip, native + unknown formatting, DTO accessors, the static/no-op conversion sources, single/empty/mixed aggregation (all-priceable, partially-priceable with remainder, and unavailable), net-worth signing across currencies, and the masked/accessible presentation rows + disclosure copy.

`./Scripts/smoke-sandbox.sh` was **not** run: it requires Plaid sandbox creds (unavailable here), and the only server change is a pure pass-through fallback already exercised by the strict build.

## Done / left

**Done:** currency model, per-currency grouping/subtotals, offline-stubbed conversion with safe degradation, native + masked + accessible presentation, formatter, DTO accessors, server unofficial-code fallback, 20 passing tests.

**Left (gated on the open decision):** binding the presenters into live views; a real offline rate table (and how/when it is refreshed); making the reporting/home currency user-selectable.

## Links

Closes AND-643

## Open questions / risks

- **Conversion source (the one open product decision):** how should cross-currency aggregates be priced without breaking the no-cloud posture? Options: (a) ship a periodically-bundled offline rate table updated with app releases; (b) let the user paste/enter rates manually; (c) never convert — always show per-currency subtotals. This PR implements the plumbing for all three (pluggable source + always-present subtotals) but ships only the `sampleOffline` stub, which must **not** be presented as authoritative — the disclosure copy already labels converted totals "approximate, offline rates."
- **Rate staleness / honesty:** any converted figure needs a visible "approximate, as of <date>" treatment before it reaches a view. The headline disclosure provides the copy; a real rate-date is still needed.
- **Reporting currency is hard-coded to USD** (`CurrencyCode.usd`). The type is ready for a user-selected home currency; that UI is out of scope here.


ftchvs:
🤖 Automated adversarial review (pre-merge, advisory) — not an approval, do not merge on this.

**Verdict: LGTM as a draft (minor notes).** I reproduced the green gates locally on `origin/feat/and-643-multi-currency`: the strict-concurrency build (`-strict-concurrency=complete -warnings-as-errors`) is clean, and all 20 `MultiCurrencyTests` pass (0.005s). The architecture is honest and well-scoped: per-currency subtotals are always preserved, `.unknown` is rigorously kept distinct from USD, there's no live FX fetch (pluggable source, offline stub only), and currency identity is carried by text/symbol + spoken `accessibleName` rather than color. The headline disclosure copy labels converted totals as "approximate, offline rates" and lists excluded currencies. This is solid plumbing.

**Findings (all non-blocking for a draft):**

1. **The dishonest aggregate it's meant to fix is still live post-merge.** The PR intentionally wires none of this into views, so the shipping app keeps using `MenuBarSummary.netCash`/`netWorth` (`Sources/PlaidBarCore/Utilities/MenuBarSummary.swift:38-55`), which blindly sums `Double` across currencies and renders with `$` via `Formatters.currency(_, format:)`. A EUR+USD user still sees a fabricated single `$` number on the Dashboard/menu bar today. That's the actual user-facing bug; this PR adds the honest path beside it but doesn't switch the call sites. Worth stating plainly in the PR so it isn't mistaken for "AND-643 fixed" — it's "AND-643 plumbing landed, fix pending the wiring PR." `WealthSummaryPresentation.swift:173` still calls `netCash`.

2. **Non-USD / abbreviated formatting is locale-dependent and a bit rough.** `Formatters.currency(_:in:format:)` for a resolved non-USD code routes to the existing `NumberFormatter`-based path which uses the default (current) locale — so grouping separators and symbol placement vary by host locale, and the `.abbreviated` branch produces `EUR1.2K` (raw code glued to the number, no space) and for `.unknown` an empty-symbol `1.2K`. Tests assert `"1,200"`/`!contains("$")`, which would break under a non-US locale. Not a correctness bug for the shipped USD default, but pin the test locale (you already do this for dates via `en_US_POSIX`) or the abbreviated currency rows will look off when wired. Low priority while unwired.

3. **Privacy Mask: figures masked, identity shown — verify this is the intended policy.** `subtotalRows`/`headline` correctly replace the amount with `PrivacyMaskPresentation.compactValue` while leaving the currency name/symbol visible (you can see "you hold EUR" without the figure). That matches the PR's stated intent and the established mask behavior, and no raw value leaks. Just flagging it as a deliberate disclosure choice for the privacy reviewer to bless, since "currency present" is itself a (small) signal. The masked path is exercised by a test. No leak found.

4. **Server fallback is sound.** `AccountRoutes.swift` `isoCurrencyCode ?? unofficialCurrencyCode` is valid — `Balances` carries both fields (`PlaidModels.swift:222-223`), it's a pure pass-through, and no secret/token crosses the boundary. App still only sees the DTO. Good.

**Secrets/concurrency:** No secrets touched; all new logic is pure `Sendable` in `PlaidBarCore`; strict build is warnings-clean. No `nonisolated(unsafe)`, no shared mutable state.

**Tests:** They genuinely exercise the change (normalization, unknown≠USD invariant, Codable round-trip, all aggregation branches incl. partial-priceable remainder and identity-only→`.unavailable` degradation, net-worth debt signing, masked rows, disclosure copy). The one gap is the locale assumption in #2.

**Must-fix before this leaves draft / before wiring:** (a) when binding presenters into views, flip the live `netCash`/`netWorth` call sites off the blind-sum path or the dishonest `$` aggregate ships; (b) pin test locale for the non-USD/abbreviated formatting assertions; (c) resolve the open conversion-source product decision before any converted headline reaches a view, and add the "as of <date>" rate-date the PR notes is still missing. Nothing here blocks keeping it as a draft.

ftchvs (commented):
Review found a serious mixed-currency correctness issue and I pushed a fix in `b2231be`.

Problem fixed: account rows, dashboard/account hero totals, menu-bar balance text, App Group snapshots, App Intents, and Control Center safe-to-spend could collapse USD/EUR/etc. into one arbitrary currency. The branch now preserves per-currency subtotals, formats account-level values in native currency, and returns per-currency/mixed-currency messaging instead of scalar value intents when conversion is unavailable.

Validation:
- `git diff --check`
- `swift test --filter "MultiCurrencyTests|FinanceSnapshotBuilderTests|FinanceIntentQueriesTests|DashboardDrillInSurfaceTests" --skip-update --disable-keychain` (59 tests)

No merge performed.

ftchvs:
Merged `origin/main` (advanced ~16 PRs since this branch was cut) and resolved all conflicts, preserving both sides' intent.

**Conflict resolution (DashboardDestinationView hero metrics):** kept BOTH the merged `FigureProvenance` popovers (AND-641) and this PR's multi-currency aggregate display + per-currency safe-to-spend headline (AND-643). Net-worth and safe-to-spend metrics now carry provenance AND multi-currency-aware detail/value.

**Regression fix (must-fix #2 / dominant-path guard):** `b2231be` routed per-account formatting through `Formatters.currency(_, in: account.balances.currency)`. Plaid frequently returns accounts with a nil `isoCurrencyCode`, which resolves to `.unknown` and dropped the `$` symbol — regressing the dominant single-USD display (`$500` → `500`, `$1,000` → `1000`). Added `AccountPresentation.displayCurrency(for:)` which falls back to the app reporting currency (USD) for a lone account's own display, restoring byte-identical `$` output for single-USD users. The `.unknown`-distinct-from-USD invariant is retained for cross-account *aggregation* (so subtotals never silently absorb an unknown). This was caught by the existing Core suite (`Account presentation derives dashboard row labels`, drill-in accessibility) which the merge surfaced as failing.

**Validation:** strict-concurrency build (`-strict-concurrency=complete -warnings-as-errors`) clean; full suite **2389 tests pass** (Keychain-env tests skipped, not failed).

**Open follow-up (by design):** the FX **conversion-source** product decision remains open. This PR ships per-currency DISPLAY + per-currency subtotals only — no live/cloud FX fetch, no converted single-number headline reaches a view. The "as of <date>" rate-date and the converted-headline path stay deferred until that source decision is made (track as a follow-up to AND-643).

PR #669, opened 2026-06-25
feat(budgets): per-category budget alerts (AND-642)

ftchvs:

## What & why

VaultPeek already computes per-category budgets (`CategoryBudgetPlanner` / `CategoryBudgetPresentation`) and shows their pressure bands in the Budgets workspace, but nothing told the user when a category crossed its limit while the app sat in the menu bar. AND-642 adds local notifications that nudge the user when a budgeted category **nears** or **exceeds** its monthly budget — closing the loop between the budget UI and the alerting pipeline.

## Change

- **Core (pure, deterministic):** new `CategoryBudgetAlertEvaluator` in `PlaidBarCore`. Given a finished, override-aware `CategoryBudgetPresentation` and an injected `now`/`Calendar` (no hidden `Date()`), it emits one alert per category in the `nearing`/`over` band, worst-first. Band classification reuses `CategoryBudgetStatus` so the alert verdict can never drift from the in-app budget UI. Suggested (unsaved) budgets are excluded by default; the near-threshold is configurable.
- **Notification pipeline:** added a `categoryBudgetAlert` trigger kind to `NotificationTriggerSelection`, plus a per-family enable gate (`categoryBudgetAlert`) and `budgetAlertNearThreshold` on `NotificationTriggers`. The evaluator is invoked alongside the existing watchlist/recurring/financial triggers.
- **De-dup:** keyed on `category + month + band`. A category fires **once** at `nearing` and **once more** if it escalates to `over` (distinct band key), never on every refresh while it stays in the same band, and re-arms with a fresh month key next month. The band crossing is a one-shot (`clearsWhenResolved == false`).
- **Privacy:** notification bodies never contain an amount. Under Privacy Mask the category name is omitted too (status verb only — "A budget category is over its monthly limit"), so the lock screen never leaks which category is under pressure. Over vs nearing is conveyed by body text + title + severity, never color.
- **App wiring:** `AppState` gains a persisted `notifyCategoryBudget` toggle, threads the live `categoryBudgetPresentation` and `shouldMaskFinancialValues` into `evaluateNotifications`, and `NotificationService.evaluateTriggers` forwards both. The app target stays localhost-only — no Plaid secrets touched.
- **Settings UI:** new "Budget alerts" section with a "Category budget warnings" toggle and an explanatory caption noting the lock-screen/Privacy-Mask behavior.

## Testing

Ran from the worktree with sandbox disabled:

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete**, no warnings/errors.
- `swift test --filter CategoryBudgetAlertEvaluatorTests` → **12 tests passed** (crossing bands, suggested exclusion, threshold override, worst-first ordering, deterministic month key, escalation de-dup, disabled-family suppression, unmasked body has no amount, masked body omits category + amount, dedup keys hide amounts).
- `swift test --filter NotificationTriggerEvaluationTests` → **9 passed** (no regression from the `NotificationTriggers` / `evaluate` signature change).
- `swift test --filter "CategoryBudget"` → **55 passed** (all budget families green).
- `git diff --check` → clean.

`./Scripts/smoke-sandbox.sh` not run: this is an app + Core workstream with no server/Plaid changes, and the smoke preflight needs the server path.

## Links

Closes AND-642.

## Open questions / risks

- The near-threshold defaults to `CategoryBudgetStatus.nearingThreshold` (0.8) to match the UI; it is plumbed through `NotificationTriggers` but not yet exposed as a user-facing slider (kept minimal). A follow-up could surface it next to the other notification thresholds.
- Alerts derive from `categoryBudgetPresentation`, which includes planner *suggestions*; these are intentionally excluded from alerting (`includeSuggested: false`) so a user only gets nudged about budgets they actually set.
- New-install behavior: the delivered-dedup set persists in `UserDefaults`, so a fresh install whose first sync immediately shows an over-budget category will fire one alert per such category (one-shot, deduped) — consistent with the existing large-transaction/watchlist one-shots.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -42,6 +42,7 @@ final class AppState {
         static let notifyRecurringChargeDueSoon = "notifyRecurringChargeDueSoon"
         static let notifyBrokenConnection = "notifyBrokenConnection"
         static let notifyWatchlist = "notifyWatchlist"
+        static let notifyCategoryBudget = "notifyCategoryBudget"
         static let watchlistTargets = "watchlistTargets"
         static let privacyMaskEnabled = "privacyMaskEnabled"
         static let appLockEnabled = UserDefaultsAppLockSettingsStore.defaultStorageKey
@@ -387,6 +388,14 @@ final class AppState {
         didSet {
             guard notifyWatchlist != oldValue else { return }
             UserDefaults.standard.set(notifyWatchlist, forKey: Keys.notifyWatchlist)
+        }
+    }
+    /// Gate for per-category budget alerts: nudge when a budgeted category nears
+    /// or exceeds its monthly limit (AND-642).
+    var notifyCategoryBudget: Bool = true {
+        didSet {
+            guard notifyCategoryBudget != oldValue else { return }
+            UserDefaults.standard.set(notifyCategoryBudget, forKey: Keys.notifyCategoryBudget)
         }
     }
     /// User-defined per-merchant / per-category spend watches (AND-501).
@@ -790,6 +799,9 @@ final class AppState {
         }
         if defaults.object(forKey: Keys.notifyWatchlist) != nil {
             notifyWatchlist = defaults.bool(forKey: Keys.notifyWatchlist)
+        }
+        if defaults.object(forKey: Keys.notifyCategoryBudget) != nil {
+            notifyCategoryBudget = defaults.bool(forKey: Keys.notifyCategoryBudget)
         }
         loadWatchlistTargets(defaults: defaults)
         loadAppLockPreferences(defaults: defaults)
@@ -3014,6 +3026,7 @@ final class AppState {
             loginRequired: notifyBrokenConnection,
             itemError: notifyBrokenConnection,
             watchlist: notifyWatchlist,
+            categoryBudgetAlert: notifyCategoryBudget,
             largeTransactionThreshold: largeTransactionThreshold,
             lowBalanceThreshold: lowBalanceThreshold,
             creditUtilizationThreshold: creditUtilizationThreshold
@@ -3024,7 +3037,9 @@ final class AppState {
             recurringTransactions: recurringTransactions,
             itemStatuses: itemStatuses,
             watchlistTargets: watchlistTargets,
+            budgetPresentation: categoryBudgetPresentation,
             isSyncStale: isSyncStale,
+            privacyMaskActive: shouldMaskFinancialValues,
             config: config
         )
     }
--- a/Sources/PlaidBar/Services/NotificationService.swift
+++ b/Sources/PlaidBar/Services/NotificationService.swift
@@ -13,7 +13,9 @@ protocol NotificationServiceProtocol {
         recurringTransactions: [RecurringTransaction],
         itemStatuses: [ItemStatus],
         watchlistTargets: [WatchlistTarget],
+        budgetPresentation: CategoryBudgetPresentation,
         isSyncStale: Bool,
+        privacyMaskActive: Bool,
         config: NotificationTriggers
     ) async
 }
@@ -123,7 +125,9 @@ final class NotificationService: NotificationServiceProtocol {
         recurringTransactions: [RecurringTransaction],
         itemStatuses: [ItemStatus],
         watchlistTargets: [WatchlistTarget] = [],
+        budgetPresentation: CategoryBudgetPresentation = .empty,
         isSyncStale: Bool,
+        privacyMaskActive: Bool = false,
         config: NotificationTriggers
     ) async {
         let evaluation = NotificationTriggerSelection.evaluate(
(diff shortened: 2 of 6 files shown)


ftchvs:

**Verdict: concerns** — one real privacy regression worth fixing before merge; the rest is solid. Do not treat this as an approval.

The Core evaluator design is clean: `CategoryBudgetAlertEvaluator` is a pure, `Sendable`, injected-`now`/`Calendar` enum that reuses `CategoryBudgetStatus` band logic, the per-`category+month+band` dedup keys are SHA256-hashed (no amounts leak into keys), `clearsWhenResolved == false` correctly makes each band a one-shot, suggested budgets are excluded by default, and the family gate + threshold are plumbed through `NotificationTriggers`. Wiring in `AppState.evaluateNotifications` and `NotificationService.evaluateTriggers` is correct and stays app/Core-only (no Plaid secrets touched). Tests genuinely exercise crossing, escalation de-dup, family suppression, ordering, and the masked path.

### Must-fix

**1. Unmasked notification body leaks the category name onto the lock screen — diverges from the established hard privacy rule.**
`NotificationTriggerSelection.categoryBudgetAlertDecision` (Sources/PlaidBarCore/Utilities/NotificationTriggerSelection.swift ~line 354) emits, when `privacyMaskActive == false`:

> "Food & Drink is over its monthly budget. Open VaultPeek for details."

That string becomes `content.body` of a `UNMutableNotificationContent` (Sources/PlaidBar/Services/NotificationService.swift:171), which renders on the **lock screen** by default. But Privacy Mask is an *in-app display toggle* — it does not govern whether the physical device is locked when the banner is delivered. Gating the category name on `!privacyMaskActive` therefore still leaks "which category is under financial pressure" to anyone glancing at a locked/asleep device whenever the user simply hasn't turned Mask on.

This is exactly the rule the sibling watchlist feature enforces *unconditionally*: `watchlistDecision` (same file, ~line 485) keeps copy generic regardless of mask — "A spending category you're watching crossed its monthly limit. Open VaultPeek for details." — with the comment "Lock-screen copy stays generic — no merchant name, category, or amount (privacy hard rule)... The exact framing lives in-app where the device is unlocked." `NotificationTriggerEvaluationTests` (line ~282-300) asserts that **no** decision body/title/dedupKey contains the merchant/category/account name or an amount, and it runs with the default `privacyMaskActive: false`.

The new path escapes that guard only because that test passes no `budgetPresentation`, and the PR's own `unmaskedBodyHasNoAmount` test then codifies the leak by asserting `body.contains("Food & Drink")`. Net: this feature is the one notification family that names the category on the lock screen.

Recommended fix: drop the category name from the body unconditionally (keep the generic "A budget category is over/nearing its monthly limit" copy you already wrote for the masked branch for *all* cases), and delete the `privacyMaskActive` branch in this decision. That also lets you remove the `privacyMaskActive` plumbing through `evaluate`/`evaluateTriggers` and removes a mask-toggle dependency from the body (note the dedup key is already mask-independent, which is good — but the body shouldn't depend on it either). If product genuinely wants the named variant, it belongs in the in-app surface, not the notification.

### Minor / non-blocking

- **De-dup floods:** the one-shot keying looks correct — same band on later refreshes is filtered by `deliveredDedupKeys`, and only an `under → nearing → over` escalation produces a new (distinct-band) key, re-arming next month via the month key. No per-refresh flood. Good. One residual the PR already calls out: a fresh install whose first sync is already over-budget fires one alert per over category; acceptable and consistent with existing one-shots.
- **Threshold-override band re-derivation:** `evaluate` re-derives the band from `item.fractionUsed` against the caller threshold rather than reusing `item.status`. Correct for a *stricter* near-threshold, but the doc-comment says values are "clamped to 0...1" while `band(forFraction:)` still treats `fraction > 1.0` as `.over` — so a threshold of 1.0 makes `nearing` an empty band (anything `>= 1.0` and `<= 1.0` is exactly 1.0 → nearing; `> 1.0` → over). Fine, just confirm that's intended (raising the slider to 100% effectively means "only alert when over, plus exactly-at-limit").
- **Stale presentation:** `categoryBudgetPresentation` is a cached computed property invalidated on transaction/budget/rule changes; notification eval reads it live, so no concurrency issue, but worth a glance that the cache is fresh at eval time (it is, since the same `didSet`s drive both).
- Strict-concurrency/Sendable: clean — `Alert` is `Sendable`, evaluator is stateless.

Nice work on the determinism and de-dup contract; the single blocker is the lock-screen category leak.

ftchvs:
✅ Addressed the must-fix privacy finding in 05d1e0b.

The category-budget notification body is now **unconditionally generic** — it never names the category (or an amount), regardless of Privacy Mask state — because a delivered notification renders on the lock screen, which the in-app Privacy Mask / App Lock can't guard. This brings it in line with every other trigger and the `NotificationTriggerEvaluationTests` invariant.

- Dropped the `privacyMaskActive` branch + parameter from `categoryBudgetAlertDecision`.
- Updated the unmasked-body test to assert the lock-screen-safe body; over/nearing decisions are now distinguished by band severity/title (bodies are identical across categories; dedup keys stay SHA256-hashed).

Validation: strict-concurrency build clean; `CategoryBudgetAlertEvaluatorTests` 12/12 and `NotificationTriggerEvaluationTests` 9/9 pass.

ftchvs (commented):
Review complete on current head. No serious findings.

I checked the category-budget alert privacy path after the prior leak was fixed: notification bodies are generic and value-free, the masked path remains wired through `shouldMaskFinancialValues`, and I did not find Plaid credential/token movement, app-target Plaid access, or color-only risk-state communication in the touched UI.

Validation reviewed:
- `git diff --check origin/main...origin/pr/669`
- strict `PlaidBar` build passed on the prior head before the branch advanced
- focused alert/notification tests were identified as the next pre-merge gate for the latest head

No merge performed.

PR #670, opened 2026-06-25
fix(webhooks): dormant ordering/state defects + gated real signature verifier (AND-634, AND-646)

ftchvs:

## What & why

Two coupled, deferred webhook-path items in the `PlaidBarServer` target.

**AND-634 (primary, fully tested)** fixes three real but currently-dormant defects in the server webhook path. They are latent because inbound webhook processing ships disabled (the receiver's signature validator always throws). Each fix lands with a regression test that fails on the pre-fix code:

1. **Clock-mixing out-of-order key.** `WebhookEventStore.record` computed ordering by comparing a stored event's `receivedAt` (delivery clock) against a new signal's `eventAt` (Plaid's event clock) — two unrelated clocks — and could spuriously flag (or miss) reordering.
2. **`needsPollingSync` masked by a later non-sync webhook.** The status path collapsed each item to its single max-`effectiveDate` event, so a later `PENDING_EXPIRATION` hid an earlier still-pending `SYNC_UPDATES_AVAILABLE`, silently dropping the sync signal.
3. **Concurrent-delivery race.** `record` suspends between the existence check and the save, and `apply` does an unlocked status read-modify-write, so two concurrent distinct-hash deliveries for the same item could both resolve `.stored` and interleave their status mutations (last-writer-wins).

**AND-646 (secondary, gated / non-activating)** wires a real ES256 webhook signature verifier to replace `UnconfiguredPlaidWebhookSignatureValidator` — but only when the operator explicitly opts in. The shipped default is byte-for-byte unchanged: the receiver stays dormant.

## Change

- **Fix 1** — `WebhookEventStore.record`: out-of-order is decided only when BOTH the stored latest event and the incoming signal carry an `eventAt`; otherwise the delivery is treated as in-order.
- **Fix 2** — new `WebhookEventStore.latestSyncEventsByItem()` filters to `needsSync` events *before* the max reduce, keeping the sync signal sticky. `StatusRoutes.needsPollingSync` reads from this sync-only projection; the latest-overall event still drives the display fields (`lastWebhookAt`, `lastWebhookEvent`). The sticky signal clears once the item's `updatedAt` advances past it.
- **Fix 3** — per-item async FIFO lock in `WebhookEventStore` (`withItemLock`, mirroring `TokenStore`'s lock idiom). `WebhookRoutes.receive` holds it across `record` + `apply`, making the out-of-order check, save, and status RMW atomic per item; distinct items stay concurrent.
- **AND-646** — new `PlaidWebhookSignatureVerifier.swift`: `ES256PlaidWebhookSignatureValidator` (pure, testable P-256/SHA-256 signature check over the JWS signing input), `PlaidWebhookKeySource` abstraction + `StaticPlaidWebhookKeySource`, and `PlaidWebhookJWK` (JWK → `P256.Signing.PublicKey`). Gated behind new `PlaidWebhookVerificationConfig` (`PLAIDBAR_WEBHOOK_VERIFICATION` opt-in; optional pinned key via `PLAIDBAR_WEBHOOK_SIGNING_JWK`), resolved in `ServerConfig.load`. `App.webhookVerifier(config:logger:)` returns the existing `StrictPlaidWebhookVerifier()` (dormant) unless explicitly enabled.

Scope is confined to the SERVER target. No app/`PlaidBarCore` changes, no secrets, no new persisted state. Any future hosted relay stores nothing — this is a local verification posture only.

## Testing

All run with the SwiftPM sandbox disabled (the repo's known requirement).

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete!** (clean, no warnings).
- `swift test --filter "WebhookDefectAndVerifierTests"` → **11 tests passed** (new suite: the three AND-634 regressions incl. concurrent serialization, plus AND-646 ES256 accept/reject, JWK round-trip, strict-verifier end-to-end, and the gating/dormant-default assertions).
- `swift test --filter "WebhookReceiverTests"` → **9 tests passed** (existing suite, no regression).
- `git diff --check` → clean.
- `./Scripts/smoke-sandbox.sh` → could not run the full preflight here: it exits early requesting `PLAID_CLIENT_ID`/`PLAID_SECRET`, which are unavailable in this environment. The changes do not touch `/health` or `/api/status`; they are confined to the (dormant-by-default) `/webhooks/plaid` receiver and the `needsSync` computation in the status snapshot.

## Links

Closes AND-634
Closes AND-646

## Open questions / risks

- The real ES256 validator's cryptographic core is fully covered with a synthetic keypair, but the *production* key source — fetching Plaid's signing key by `kid` from `/webhook_verification_key/get` (which needs the server's Plaid credentials) — is intentionally left as the `PlaidWebhookKeySource` seam plus an operator-pinned-JWK path, not a live fetcher. Enabling end-to-end production verification still requires provisioning a key (or implementing a Plaid-backed key source) and flipping the opt-in; until then the receiver stays dormant.
- AND-634 fixes are inert in shipped behavior today (the receiver is disabled), so they cannot regress current users; they harden the path for whenever AND-646 is activated.

### DIFF
--- a/Sources/PlaidBarServer/App.swift
+++ b/Sources/PlaidBarServer/App.swift
@@ -1,4 +1,5 @@
 import ArgumentParser
+import CryptoKit
 import FluentSQLiteDriver
 import Foundation
 import Hummingbird
@@ -138,7 +139,7 @@ struct PlaidBarServer: AsyncParsableCommand {
         )
         .register(with: router)
         WebhookRoutes(
-            verifier: StrictPlaidWebhookVerifier(),
+            verifier: Self.webhookVerifier(config: serverConfig, logger: logger),
             tokenStore: tokenStore,
             eventStore: webhookEventStore
         )
@@ -185,6 +186,57 @@ struct PlaidBarServer: AsyncParsableCommand {
                     break
                 }
             }
+        }
+    }
+
+    /// Selects the webhook JWT verifier (AND-646). GATED: by default
+    /// (`webhookVerification.enabled == false`) this returns the exact same
+    /// `StrictPlaidWebhookVerifier()` the server has always wired — its default
+    /// `UnconfiguredPlaidWebhookSignatureValidator` always throws, so the
+    /// receiver stays dormant and shipped behavior is byte-for-byte unchanged.
+    /// The real ES256 validator is only constructed when the operator explicitly
+    /// opts in; if the opt-in is present but supplies no parseable signing key,
+    /// the validator is still wired but trusts no key (it rejects every
+    /// delivery), so processing remains inert from the user's perspective.
+    static func webhookVerifier(
+        config: ServerConfig,
+        logger: Logger
+    ) -> any PlaidWebhookVerifier {
+        guard config.webhookVerification.enabled else {
+            return StrictPlaidWebhookVerifier()
+        }
+        let keySource = StaticPlaidWebhookKeySource(
+            keysByID: pinnedWebhookKeys(config: config, logger: logger)
+        )
+        logger.info("Plaid webhook signature verification ENABLED (opt-in).")
+        return StrictPlaidWebhookVerifier(
+            signatureValidator: ES256PlaidWebhookSignatureValidator(keySource: keySource)
+        )
+    }
+
+    /// Parses the optional operator-pinned signing key (JWK JSON) into a
+    /// `kid`-indexed key set. Returns an empty set (verifier rejects everything)
+    /// if absent or unparseable, never crashing the server boot.
+    private static func pinnedWebhookKeys(
+        config: ServerConfig,
+        logger: Logger
+    ) -> [String: P256.Signing.PublicKey] {
+        guard let json = config.webhookVerification.signingKeyJWKJSON,
+              let data = json.data(using: .utf8)
+        else {
+            logger.warning("Webhook verification enabled but no signing key configured; all deliveries will be rejected.")
+            return [:]
+        }
+        do {
+            let jwk = try JSONDecoder().decode(PlaidWebhookJWK.self, from: data)
+            guard let kid = jwk.kid, !kid.isEmpty else {
+                logger.warning("Configured webhook signing JWK has no `kid`; all deliveries will be rejected.")
+                return [:]
+            }
+            return [kid: try jwk.publicKey()]
+        } catch {
+            logger.warning("Failed to parse configured webhook signing JWK: \(String(describing: error)); all deliveries will be rejected.")
+            return [:]
         }
     }
 }
--- a/Sources/PlaidBarServer/Config/DeploymentMode.swift
+++ b/Sources/PlaidBarServer/Config/DeploymentMode.swift
@@ -104,3 +104,55 @@ struct RemoteBridgeConfig: Sendable, Equatable {
         )
     }
 }
+
+/// Opt-in configuration for real Plaid webhook signature verification (AND-646).
+///
+/// GATED / NON-ACTIVATING BY DEFAULT. The shipped server keeps the dormant
+/// webhook receiver: it wires `UnconfiguredPlaidWebhookSignatureValidator`,
+/// which always throws, so no inbound webhook is ever accepted unless the
+/// operator explicitly opts in. This config records *whether* the real ES256
+/// validator should be wired and, optionally, a statically pinned EC public key
+/// (JWK JSON) to verify against. Until `enabled` is `true`, nothing constructs
+/// the real validator and the receiver stays off exactly as before.
+///
+/// `enabled` is parsed from `PLAIDBAR_WEBHOOK_VERIFICATION` (`on`/`true`/`1`);
+/// any other value (including absent) resolves to disabled — a malformed config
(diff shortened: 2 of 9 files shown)


ftchvs:
🤖 Automated adversarial review (pre-merge, advisory) — not an approval.

**Verdict: LGTM (minor advisories only).** I reproduced the build and both test suites on the PR head (`674e8ff`) with the SwiftPM sandbox disabled:

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete!** (clean, no warnings)
- `swift test --filter WebhookDefectAndVerifierTests` → **11 passed**
- `swift test --filter WebhookReceiverTests` → **9 passed** (no regression)

Scope is genuinely server-only (`git diff --name-only origin/main 674e8ff` touches only `Sources/PlaidBarServer/*` + the new test). No `PlaidBar`/`PlaidBarCore`/`PlaidBarCache` changes, so no Privacy-Mask / color-meaning / UI surface is in play, and no secret crosses the localhost boundary. The pinned JWK is an EC **public** key, read from env only on explicit opt-in — correct.

### AND-634 defects — each fix verified correct and tested
1. **Clock-mixing out-of-order key** — `WebhookEventStore.record` now decides ordering only when **both** the stored latest event and the incoming signal carry a real `eventAt` (`if let latestEventAt, let signalEventAt`), else treats as in-order. This replaces the old `effectiveDate > …` comparison that fell back to `receivedAt` and mixed the delivery clock with Plaid's event clock. Out-of-order events are still **persisted** (flagged, not dropped), so no data loss. Covered by both the "only-when-both" and "still-fires-when-both" regression tests.
2. **Sticky `needsSync`** — new `latestSyncEventsByItem()` filters to `needsSync` **before** the max-reduce, so a later `PENDING_EXPIRATION` no longer masks an earlier pending `SYNC_UPDATES_AVAILABLE`; display fields still come from the latest-overall event. Clears once `item.updatedAt` advances past the sync event. Both the masking and the clear-after-refresh paths are tested.
3. **Per-item async FIFO lock** — `withItemLock` wraps `record`+`apply` in `WebhookRoutes.receive`. The lock is actor-isolated (acquire/release run on the executor, no interleaving except the `withCheckedContinuation` append), `defer { releaseItemLock }` releases on throw, and ownership hand-off keeps the id in `lockedItemIds` for the resumed waiter — correct, no lost-wakeup, no leak. Distinct items stay concurrent (tested).

### AND-646 verifier — dormant-by-default and crypto sound
- Gating is airtight: the **only** construction site of `ES256PlaidWebhookSignatureValidator` is `App.webhookVerifier`, guarded by `config.webhookVerification.enabled`, which defaults to `.disabled` and flips only on explicit `on/true/1/yes`. Blank/garbage/absent → disabled (tested). `defaultVerifierStaysDormant` proves the shipped path still throws `signatureVerificationUnavailable`.
- Signature path is safe: `alg` is hard-pinned to `ES256` in `StrictPlaidWebhookVerifier` before any crypto, so no `alg:none`/RS256 confusion; `kid` only selects among statically pinned trusted keys (unknown `kid` → `unknownKeyID` reject, never a fallback-accept); signature is raw 64-byte `r||s` over `SHA256(header.payload)`; body-hash gate runs before crypto. Tamper/unknown-kid/missing-kid/non-EC-JWK all rejected (tested).

### Minor / advisory (non-blocking)
- **Concurrency test is a smoke test, not a true interleave repro.** `concurrentDistinctDeliveriesAreSerialized` asserts "2 rows + one of two valid final statuses" — that assertion passes even on the pre-fix code (the unique-constraint recovery already prevents lost rows, and either final status is "valid"). It exercises the lock path but does not *prove* the race is fixed the way the PR text implies the AND-634 regressions do. Consider asserting ordering/serialization more directly (e.g. instrument that the two `apply`s never interleave) if you want this to be a true regression.
- **Enabled-but-no-key is a foot-gun.** Opting in without `PLAIDBAR_WEBHOOK_SIGNING_JWK` (or with an unparseable one) logs a `warning` and rejects every delivery — correct fail-closed posture, but an operator could silently disable their webhook path. The warning is good; just flagging that there's no live Plaid key-source seam yet (already noted in the PR's open questions).
- The ISO8601 formatter is re-instantiated per call in a couple of spots (`idempotencyHash`, `decodeDate`) — pre-existing, not introduced here, ignorable.

### Must-fix
None. The dormant defaults are preserved byte-for-byte, the crypto is correctly pinned, and the AND-634 fixes are correct and (mostly) well-tested. Recommend tightening the concurrency-test assertion before relying on it as a regression guard, but that does not block merge.

ftchvs (commented):
Review found a webhook verification hardening issue and I pushed a fix in `7aa5ab3`.

Problem fixed: the strict Plaid webhook verifier compared the JWT body hash to the actual body hash with normal string equality. Plaid's current verification guidance requires constant-time hash comparison, so the verifier now uses a length-checked byte comparison and has regression coverage.

Validation:
- `git diff --check`
- `swift test --filter "verifierBodyHashComparison|strictVerifierEndToEndWithES256"`
- parallel server/privacy review also passed `git diff --check` and strict `PlaidBarServer` build

No merge performed.

PR #671, opened 2026-06-25
feat(investments): Plaid Investments holdings + net-worth inclusion (AND-644)

ftchvs:

Opened as a **DRAFT**: this is the large AND-644 Investments product. The
pure logic, server endpoints, app surfacing, and demo path are complete and
fully tested, but the live Plaid path (`/investments/holdings/get`,
`/investments/transactions/get`) cannot be exercised without Plaid sandbox
credentials and a sandbox item linked with the `investments` product, so I
am leaving it in draft for on-creds QA before merge.

## What & why

Adds the Plaid Investments product end-to-end — holdings, securities, and
investment transactions — surfacing brokerage positions and folding
investment account value into net worth. Closes the post-1.0 deferred
candidate AND-644.

The `investment` account category already existed (`AccountType.investment`,
its icon, and the balance-mix segment), so investment **accounts** already
folded into net worth. This PR adds the missing **product depth**: the
actual positions (holdings + securities) and investment transactions, with a
holdings detail surface.

## Change

**Core (`PlaidBarCore`)** — pure, `Sendable`, most-tested layer
- `Models/InvestmentDTO.swift` — `HoldingDTO`, `SecurityDTO`,
  `InvestmentTransactionDTO`, and the `InvestmentsResponse` wire shape.
  `HoldingDTO` derives `marketValue` (institution value → quantity×price → 0)
  and `unrealizedGain` (nil without cost basis).
- `Utilities/InvestmentHoldingsPresentation.swift` — joins holdings to
  securities, sorts by descending market value, and produces per-account /
  portfolio rollups. Gain/loss is **never color-alone** (ACCESSIBILITY.md):
  a `Direction` enum carries a shape glyph (arrow up/down/dash) + a
  sign-prefixed amount + a spoken VoiceOver word. Every currency/quantity
  surface is Privacy-Mask-aware.

**Server (`PlaidBarServer`)** — owns the Plaid access token; ships only
secret-free DTOs
- `Routes/InvestmentRoutes.swift` — `GET /api/investments/holdings` and
  `GET /api/investments/transactions`, both behind `APITokenMiddleware`.
  Best-effort per item: an item linked without the `investments` scope
  returns a Plaid product error that is swallowed per item, so a scope gap
  never fails the whole request (mirrors `AccountRoutes.listLiabilities`).
- `Plaid/PlaidClient.swift` — `/investments/holdings/get` and
  `/investments/transactions/get` with protocol defaults so test doubles and
  non-Plaid providers opt out cleanly.
- `Plaid/PlaidModels.swift` — request/response models for both calls.
- `App.swift` — registers `InvestmentRoutes`.

**App (`PlaidBar`)** — localhost only; never sees Plaid secrets
- `Services/ServerClient.swift` + `ServerEndpoint` — `getInvestmentHoldings()`.
- `App/AppState.swift` — `investments` state + `refreshHoldings()` called
  alongside `refreshLiabilities()` on the live and auto-refresh paths;
  seeded in `--demo`.
- `Views/AccountDetailFlyout.swift` — a Holdings section in the account
  inspector for investment accounts, rendering the Core rows. Gain cue uses
  glyph + signed text with color layered additively.
- `Utilities/DemoFixtures.swift` — a demo Fidelity brokerage account plus
  four holdings whose values reconcile to the account balance; net worth
  updated accordingly.

Security boundary preserved: all Plaid secret/token handling stays
server-side; the app and Core never see a `client_secret` or `access_token`.

## Testing

```
# → Build complete!

swift test --filter InvestmentHoldingsPresentationTests \
           --filter InvestmentRoutesTests \
           --filter DemoFixturesTests \
           --filter WealthSummaryPresentationTests
# → Test run with 37 tests passed

swift test --filter PlaidBarCoreTests     # → 1951 tests passed
swift test --filter PlaidBarServerTests   # → 173 tests passed (incl. new InvestmentRoutesTests)

git diff --check                          # → clean

./Scripts/smoke-sandbox.sh
# → exits cleanly: "Missing Plaid sandbox credentials." Cannot exercise the
#   new /api/investments/* endpoints without creds (expected for this draft).
```

New coverage: `InvestmentHoldingsPresentationTests` (market value / gain /
sort / Privacy Mask / missing-security fallback / summary rollups /
no-cost-basis / direction + formatting helpers) and `InvestmentRoutesTests`
(Plaid→DTO mapping, unknown-type fallback, investment-transaction mapping,
per-item merge de-dup, lookback window).

## Links

Closes AND-644.

## Open questions / risks

- **Live validation pending (why this is a draft).** The mapping is written
  to Plaid's documented `/investments/holdings/get` /
  `/investments/transactions/get` shapes but has not been run against a real
  sandbox item with the `investments` product. On-creds QA should confirm
  field names decode under `convertFromSnakeCase` and that values reconcile.
- **Investment transactions endpoint is wired but not yet surfaced** in the
  UI (holdings only). The route + DTO + client default exist so the app can
  adopt it without another server change.
- **Net-worth inclusion** currently flows through the investment **account
  balance** (already in net worth). Holdings are shown for detail and
  reconciliation; if an institution reports an account balance that diverges
  from the sum of holdings, the account balance remains the source of truth
  for net worth (matching how cash/credit work today).
- **Short positions** (negative quantity/value) flow through `marketValue`
  as-is; the asset-total clamp in `WealthSummaryPresentation` already guards
  net worth against negative asset contributions.


ftchvs:
🤖 Automated adversarial review (pre-merge, advisory) — not an approval, do not merge on this alone.

**Verdict: minor (LGTM-with-nits for a draft).** The scaffold is coherent, mirrors the existing Account/Liability route + client patterns faithfully, keeps Plaid secrets server-side, folds investment value into net worth without double-counting, and is genuinely tested at the pure-logic + mapping layers. Nothing here destabilizes existing account flows. Findings below are advisory; the live-Plaid decode caveat in the PR body is the real gate.

### Verified good
- **Secret boundary intact.** `client_id`/`secret` live only in `PlaidAuthenticatedRequest` / `PlaidInvestmentTransactionsRequest` inside `PlaidClient`; the app-facing `InvestmentsResponse`/`HoldingDTO`/`SecurityDTO`/`InvestmentTransactionDTO` carry no secrets, tokens, or CUSIP/ISIN/SEDOL. App `ServerClient.getInvestmentHoldings()` only talks localhost.
- **Net worth is correct, no double-count.** `WealthSummaryPresentation.totalAssets` (main) reduces over non-debt account *balances*; investment value enters once via the brokerage `AccountDTO`. Holdings are detail-only — `totalMarketValue(holdings:)` is not wired into net worth. Demo `netWorth` 17,604.24 → 60,104.24 (+42,500) reconciles, and `DemoFixturesTests.netWorthMatchesAccountBalances` derives the total dynamically so it stays green. `GlanceSnapshotTests`' hardcoded 17,604.24 literals don't reference `DemoFixtures`, so no regression.
- **Accessibility.** Gain/loss is never color-alone: `Direction` carries a shape glyph (arrow.up.right / arrow.down.right / minus) + sign-prefixed text + spoken word; `HoldingRowView` layers `SemanticColors` additively. New `.investment` category stays icon-based (`chart.line.uptrend.xyaxis`), non-color-coded.
- **Privacy Mask.** `marketValueText`, `quantityText`, `gainText`, summary totals, and the a11y label are all masked at the Core layer (tests assert `PrivacyMaskPresentation.compactValue` for each). View does no masking of its own.
- **Concurrency.** All new DTOs/models are `Sendable`; route is `struct ... : Sendable`; mapping is pure static. `refreshHoldings()` mirrors `refreshLiabilities()` guard-for-guard.

### Minor / advisory (non-blocking)
1. **Masked holdings section vs. placeholder inconsistency** — `AccountDetailFlyout`: when `shouldMaskFinancialValues` is on, the rest of the inspector collapses into `privateDetailsPlaceholder()`, but `holdingsSection()` still renders (with masked values) *above* it. Values are safely masked, so it's not a leak — but it's a UX inconsistency (a masked Holdings block sitting next to a "private details hidden" placeholder). Consider gating the holdings section behind `!shouldMaskFinancialValues` for parity, or fold it into the placeholder branch.
2. **`jsonResponse` duplicated** — `InvestmentRoutes` defines its own private `jsonResponse<T:Encodable>` identical to `AccountRoutes.jsonResponse`. It already reuses `AccountRoutes.deterministicItems` / `.defaultMaxConcurrentItemRefreshes`; reuse the existing helper instead to avoid drift.
3. **`merge` drops first-wins detail silently** — securities de-dup keeps the first occurrence; holdings are appended with no de-dup. Fine under the one-account-per-item invariant, but a defensive comment (or de-dup by `HoldingDTO.id`) would harden it if an item is ever returned twice.
4. **`getInvestmentTransactions` route catches *all* errors per item** (auth/network/decode → `[]`), same as `listLiabilities`. Accepted pattern, but it means a genuine outage looks identical to "no investments scope." Not new, just flagging for the on-creds QA pass.
5. **Transactions endpoint maps only `investmentTransactions`**, dropping the response's `securities` — acceptable since the route is wired-but-unsurfaced per the PR body; just confirm the consumer joins securities when it's adopted.

### Must-fix before un-drafting
- **Live decode validation (already called out).** The Plaid `convertFromSnakeCase` decode of `/investments/holdings/get` + `/investments/transactions/get` has not run against a real sandbox item with the `investments` product. Confirm field decode + value reconciliation on-creds before flipping out of draft — that's the actual blocker, not the code.

No correctness bug found in the pure/mapping layers. Good draft.

ftchvs (commented):
Review found two privacy issues and I pushed fixes in `09e0565` and `5a812ba`.

Problems fixed:
- SwiftUI row identity and missing-security fallback text no longer expose raw Plaid account/security identifiers.
- Privacy Mask no longer reveals holdings composition: the inspector suppresses the holdings section while masked, and Core masked rows omit security names, tickers, and security type labels.

Validation:
- `git diff --check`
- `swift test --filter InvestmentHoldingsPresentationTests --skip-update --disable-keychain` (16 tests)

No merge performed.

PR #672, opened 2026-06-25
feat(budgeting): per-month budgets + rollover envelope carry (AND-548)

ftchvs:

## What & why

Moves budgeting from a single, undated per-category limit (v1) to **per-month budgets with optional rollover ("envelope carry")**, per AND-548. Builds directly on the merged AND-546 schema — `MonthlyBudgetV2(month, categoryId, monthlyLimit, rollover)` already models the per-(month, category) row with a per-row rollover flag; the foundation only *stored* those fields. This PR adds the **math and the per-month editing rules** on top, as pure `Sendable` `PlaidBarCore` logic.

Purely **additive**: no existing file is modified. A user who does not opt into v2 (or has no v2 data) reads/writes none of this and sees byte-identical v1 behavior.

## Change

- **`RolloverBudgetPlanner` (PlaidBarCore, new)** — pure, deterministic envelope carry across a category's chronological `MonthlyBudgetV2` rows:
  - `available = baseLimit + carriedIn`; `remaining = available - spent`; `carriedOut = rolloverActive ? remaining : 0`, fed into the next month's `carriedIn`.
  - **Unspent *and* overspend carry** — the remainder is signed, so over-budget months subtract from the next month (true envelope).
  - **Per-category opt-out**: the row's `rollover` flag (a category opts out by setting it `false` on its rows) *and* a global `optedOutCategoryIds` toggle that forces `carryOut == 0` for a whole category.
  - A **budget-row gap** (a missing intervening month) or a **rollover-off month** resets the chain — a later month starts from its bare limit.
  - **Historical immutability policy** `isMonthEditable(_:asOf:)` (current + future editable, past frozen) and **month-key arithmetic** (`nextMonthKey` / `previousMonthKey` / `monthKey(for:calendar:)`) as the single source of truth.
  - No hidden `Date()` — explicit `Calendar` / `asOf` month key throughout.
- **`MonthlyBudgetEditor` (PlaidBarCore, new)** — pure upsert / remove / forward-seed over `BudgetingV2Schema` that **enforces immutability**: an edit to a frozen historical month is a no-op (schema returned byte-identical). Validates the limit (finite, `>= 0`) and that the category exists; `rolloverTemplateToNextMonth` copies the current month's budgets forward without clobbering an existing forward edit.

## Testing

Exact commands and results:

```
Build complete! (126.76s)

$ swift test --filter RolloverBudgetPlannerTests --filter MonthlyBudgetEditorTests
Test run with 31 tests passed after 0.006 seconds.

$ git diff --check
(clean)
```

31 new deterministic Core tests: carry of unspent / overspend, compounding across months, Dec→Jan year boundary, gap-resets-carry, row-flag and global opt-out (incl. independent per-category in `resolveCarryByCategory`), unsorted/duplicate input degradation, editability policy + fail-closed on malformed keys, editor immutability + input validation + forward-seed, and an edit-then-resolve end-to-end.

## Links

Closes AND-548

## Follow-ups

- Wire these into the app's per-month Budgets UI and the `BudgetingV2Store` save path (later AND-524 subtree; out of scope here — this PR is the pure Core engine).
- `overrideAwareSpend` (AND-546) is the intended `spendByMonth` source; a thin adapter that fans it out per month/category can live with the UI wiring.

### DIFF
--- a/Sources/PlaidBarCore/Utilities/MonthlyBudgetEditor.swift
+++ b/Sources/PlaidBarCore/Utilities/MonthlyBudgetEditor.swift
@@ -1,0 +1,173 @@
+import Foundation
+
+/// Pure, deterministic per-month budget editing for budgeting v2 (AND-548).
+///
+/// The carry math lives in ``RolloverBudgetPlanner``; this is the small companion
+/// that *mutates the stored set* of ``MonthlyBudgetV2`` rows while enforcing the
+/// AC's immutability rule: **historical months are frozen, the current and future
+/// months are editable**. Every entry point returns a new ``BudgetingV2Schema``
+/// (value semantics) and leaves the input untouched, so it stays `Sendable` and
+/// trivially testable — no hidden `Date()`; "now" is the explicit `asOf` month key.
+///
+/// Editing a frozen historical month is rejected by returning the schema unchanged
+/// (a no-op), so a stale UI action can never rewrite history. Whether a month is
+/// editable is decided by ``RolloverBudgetPlanner/isMonthEditable(_:asOf:)`` — the
+/// single source of truth shared with the editor's enable/disable state.
+public enum MonthlyBudgetEditor {
+
+    /// The outcome of an edit attempt: the (possibly unchanged) schema plus whether
+    /// the edit was applied. `applied == false` means the target month was frozen
+    /// history and the schema is byte-identical to the input.
+    public struct EditResult: Sendable, Hashable {
+        public let schema: BudgetingV2Schema
+        public let applied: Bool
+
+        public init(schema: BudgetingV2Schema, applied: Bool) {
+            self.schema = schema
+            self.applied = applied
+        }
+    }
+
+    /// Upsert a per-month limit (and rollover flag) for one category, but only when
+    /// `month` is editable as of `asOf` (current or future). A frozen historical
+    /// month is a no-op.
+    ///
+    /// - Parameters:
+    ///   - schema: the current v2 snapshot.
+    ///   - categoryId: the ``BudgetCategoryV2/id`` to budget. Must exist in the
+    ///     schema's category table; an unknown id is rejected (no-op) so a budget can
+    ///     never reference a missing category.
+    ///   - month: the `YYYY-MM` month to set.
+    ///   - monthlyLimit: the new limit. A non-finite or negative value is rejected
+    ///     (no-op) — a budget limit is `>= 0`.
+    ///   - rollover: whether this month's remainder carries into the next month.
+    ///   - asOf: the current `YYYY-MM` (injected "now").
+    /// - Returns: the updated schema and whether the edit applied.
+    public static func setBudget(
+        in schema: BudgetingV2Schema,
+        categoryId: String,
+        month: String,
+        monthlyLimit: Double,
+        rollover: Bool,
+        asOf: String
+    ) -> EditResult {
+        guard
+            RolloverBudgetPlanner.isMonthEditable(month, asOf: asOf),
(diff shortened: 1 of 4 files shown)


ftchvs:

**Verdict: LGTM — no must-fixes.** The carry math, historical-immutability policy, and determinism all hold up under adversarial probing. Merge-ready from a correctness standpoint; findings below are minor/advisory only.

## What I verified (locally, on a fresh `origin/feat/and-548-per-month-budgets` worktree)
- **Builds clean** under the CI gate: `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → `Build complete!`. Both new types are pure `Sendable` `PlaidBarCore` logic.
- **All 31 new tests pass** (17 `RolloverBudgetPlannerTests` + 14 `MonthlyBudgetEditorTests`), matching the PR body.
- **Builds on the merged AND-546 schema, not a reinvention.** `MonthlyBudgetV2(month, categoryId, monthlyLimit, rollover)`, `BudgetingV2Schema.category(id:)`, `BudgetingV2Migration.seed(carryingForward:month:)` all exist on `origin/main` and are consumed as-is.
- **Purely additive / byte-identical v1.** `git diff --name-status origin/main..HEAD` = 4 **added** files, **zero** existing files modified. Nothing is wired into the app/store yet, so no opt-in path is reachable and the flag-OFF/no-v2-data build is unchanged. (Confirms the "additive" claim.)

## Focus-area findings
- **Carry-forward math (over/underspend, opt-out):** correct. `remaining` is signed so overspend subtracts from next month (true envelope); rollover-off and global `optedOut` both force `carryOut == 0`. I added extra probes the suite doesn't have — a **rollover-off month mid-chain** (contiguous, but breaks the carry: Feb `carriedIn==500`, `carriedOut==0`, Mar starts bare 500 ✓) and a **refund / negative spend** (`remaining` and carry increase ✓) — both behave correctly.
- **Month-boundary / gap resets:** Dec→Jan crossing and missing-intervening-month resets both correct; `YYYY-MM` lexicographic ordering is sound for the 4-digit-year domain.
- **Historical immutability:** `isMonthEditable` = `month >= asOf` with fail-closed on malformed keys; editor no-ops on frozen months and returns a byte-identical schema (asserted via `schema == before`, which works because the schema is `Hashable`/`Equatable`). Solid.
- **Determinism:** no hidden `Date()` (grep finds only doc-comment mentions); the only date entry point takes an injected `Calendar`. No logging/`print`/`os_log` — no financial-data leak surface in this Core layer.

## Minor / advisory (non-blocking)
1. **`MonthResult.id` is not part of `Equatable`.** The struct synthesizes equality over stored properties only; the computed `var id` doesn't participate. Harmless here (dedup prevents `id` collisions), but if anything later keys/dedups `MonthResult` by `id` assuming it's identity, that's a latent footgun. Consider a doc note.
2. **`removeBudget` returns `applied: true` even when no matching row existed** (documented as intentional — post-state is "no budget"). A caller that gates a disk write on `applied` will do a redundant no-op write. Cheap to tighten (`applied = removedSomething`) if the UI wiring cares.
3. **Dead-defensive branch:** in `resolveCarry`, the gap reset runs before `guard let row = rowByMonth[month]`, but `month` always comes from `rowByMonth.keys`, so the guard can never fail. Not a bug — just unreachable.
4. **Wiring gap is acknowledged** (the engine isn't connected to `BudgetingV2Store`/UI; `overrideAwareSpend` adapter is a follow-up). Flagging only so reviewers don't expect end-to-end behavior from this PR — it's the pure engine, as stated.

No correctness, concurrency, privacy, or accessibility blockers found.

PR #663, opened 2026-06-25
feat(budgeting): persisted Category/Budget schema + migration foundation (AND-546)

ftchvs:
## What & why

**DEFERRED epic AND-524 — reverses the v1 launch cutline; foundation only, dependents AND-547..553 follow after review.** Open as a DRAFT.

This is the **schema foundation (AND-546)** for budgeting v2: introduce persisted `Category` / `CategoryGroup` / `Budget(month, rollover)` tables to replace the flat-enum `SpendingCategory` primary key — **without changing v1 behavior unless a user opts in**. The new tables seed one-to-one from the existing `SpendingCategory` set so current categorization is preserved bit-for-bit. It also fixes the known override-unaware spend math in `CategoryBudgetPlanner`.

Explicitly **out of scope** here (separate later PRs): the editor UI, per-month budgets, rebalance, splits, rules manager, and sync (AND-547…AND-553).

## Change (files)

**PlaidBarCore (pure, `Sendable`, unit-tested):**
- `Sources/PlaidBarCore/Models/BudgetingV2Schema.swift` — new value types: `BudgetCategoryGroupV2`, `BudgetCategoryV2`, `MonthlyBudgetV2` (month `YYYY-MM` + limit + rollover), and the `BudgetingV2Schema` snapshot. Each seeds one-to-one from the closed `SpendingCategory` / `CategoryGroup` taxonomy keyed by stable `rawValue`, so a v1 budget and every categorized transaction maps to its v2 row with no reclassification. Schema-version tagged.
- `Sources/PlaidBarCore/Utilities/BudgetingV2Migration.swift` — pure, deterministic **forward seed** (`seed(carryingForward:month:)`, additive — never mutates a v1 record), **reverse opt-out** (`reverseToV1Budgets` recovers v1 limits losslessly for a month), and `needsMigration` for self-healing.
- `Sources/PlaidBarCore/Utilities/CategoryBudgetPlanner.swift` — **override-aware spend fix**: new public `overrideAwareSpend(transactions:month:metadata:rules:calendar:)` that *always* resolves per-transaction user category overrides, recategorization rules, and budget exclusions before bucketing (via the existing `EffectiveCategoryResolver` path). The v1 default `netSpendByCategory` path (both args `nil`) is left **byte-for-byte unchanged**, so not-opted-in users are unaffected. Adds a private `monthStartDate(fromMonthKey:)` parser (malformed key → empty, no crash).

**PlaidBarCache (app-only file-backed store):**
[...]

PR #673, opened 2026-06-25
fix(currency): honest mixed-currency menu-bar glance (AND-643 follow-up)

ftchvs:
## What & why

#668 (AND-643 multi-currency) made the **dashboard** currency-honest, but the
**menu-bar glance** still degraded a mixed-currency balance to the literal
`"By currency"` — no number at all. Single-currency users were unaffected, but a
user whose accounts span multiple currencies got a useless glance. This finishes
the wiring so the constrained menu-bar surface shows an honest, currency-aware
figure without ever fabricating a cross-currency `$` total.

## Change

All in `PlaidBarCore` (pure, `Sendable`, tested) plus the menu-bar text path:

- **`MultiCurrencyBalancePresentation.glance(from:format:privacyMaskEnabled:)`** (new)
  - All accounts share **one** currency (or a real conversion exists) → returns
    the single figure verbatim. Single-USD output is **byte-identical** to before
    (`Formatters.currency(amount, in: .usd, …)` collapses to the legacy
    `Formatters.currency(amount, …)` path).
  - Accounts span **multiple** currencies with no offline rate → shows the
    **dominant** currency's subtotal (largest by absolute value) in that
    currency's own symbol, plus a non-color `"+"` marker
    (`multiCurrencyMarker`) signalling other currencies exist. **No fabricated
    cross-currency `$` sum**, no invented cloud FX.
  - Carries an accessibility label naming the figure as the dominant-currency
    subtotal and noting the others are shown separately (never meaning-by-color).
- **`MenuBarSummary.text`** routes `netWorth`/`netCash`/`totalCash` through
  `glance(…).text` instead of `displayText`.
- **`MenuBarSummary.accessibleValueText`** (new) returns the glance's spoken
  label for balance modes (plain `text` otherwise); `AppState.menuBarAccessibleValueText`
  threads it into `menuBarAccessibilityLabel` so VoiceOver doesn't read a bare
  `"€1,200+"` whose `+` and currency are visual-only. Same app-lock / Privacy
  Mask masking as `menuBarText`.

Additive: a single-currency user (and the masked/locked/empty states) sees
byte-identical prior behavior. The dashboard `headline`/`subtotalRows`/`displayText`
APIs are untouched.

## Testing

Exact commands (run with the sandbox disabled per repo setup):

```
# → Build complete! (strict concurrency, warnings-as-errors)

swift test --filter MultiCurrencyTests
# → Test run with 25 tests passed (4 new)

swift test --filter MenuBar
# → Test run with 37 tests passed (announcement/glance/review goldens intact)

swift test --filter PlaidBarCoreTests
# → Test run with 2089 tests passed

git diff --check
# → clean
```

New tests in `MultiCurrencyTests`:
- single-USD menu-bar text byte-identical to the legacy formatter across
  `.compact`/`.full`/`.abbreviated`, with no `"+"` marker;
- mixed-currency picks the dominant subtotal + `"+"` marker + correct symbol and
  emits no wrong `$` (no `7,000` blind sum);
- dominant pick is by **absolute** magnitude (a large negative-debt subtotal wins);
- VoiceOver label present, names the dominant subtotal, discloses the others.

## Links

Closes AND-643 (menu-bar follow-up to #668).

## Follow-ups

- If/when an offline FX source is wired into the menu-bar path, the `.converted`
  branch already routes through `glance` and will show a single disclosed
  approximate total instead of the dominant-subtotal fallback — no further glance
  change needed.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -1120,6 +1120,31 @@ final class AppState {
         )
     }
 
+    /// Spoken form of the menu-bar value. Matches ``menuBarText`` for every mode
+    /// except mixed-currency balances, where it names the dominant currency's
+    /// subtotal in words (so VoiceOver does not read a bare `"€1,200+"` whose `+`
+    /// and currency are visual-only). Goes through the same app-lock/icon-only
+    /// masking as ``menuBarText`` so a locked or masked state never leaks a figure.
+    var menuBarAccessibleValueText: String {
+        let safeToSpend: Double? = menuBarSummaryMode == .safeToSpend
+            ? currentSafeToSpendAmount()
+            : nil
+        let rawText = MenuBarSummary.accessibleValueText(
+            mode: menuBarSummaryMode,
+            accounts: accounts,
+            transactions: transactions,
+            currencyFormat: balanceFormat,
+            isInitialLoad: isBootLoadInFlight,
+            privacyMaskEnabled: shouldMaskFinancialValues,
+            precomputedSafeToSpend: safeToSpend
+        )
+        return appLockPreferences.menuBarText(
+            currentText: rawText,
+            isAppLocked: isAppLocked,
+            isIconOnly: menuBarSummaryMode == .iconOnly
+        )
+    }
+
     var menuBarStatusPresentation: MenuBarStatusPresentation {
         MenuBarStatusPresentation.evaluate(
             isDemoMode: isDemoMode,
@@ -1187,7 +1212,7 @@ final class AppState {
         // status to keep VoiceOver in sync with the badge sighted users see.
         MenuBarAnnouncement.accessibilityLabel(
             mode: menuBarSummaryMode,
-            valueText: menuBarText,
+            valueText: menuBarAccessibleValueText,
             reviewCount: transactionReviewCount,
             diagnosticsSummary: diagnosticsSummary,
             attentionText: menuBarAttentionText,
--- a/Sources/PlaidBarCore/Utilities/MenuBarSummary.swift
+++ b/Sources/PlaidBarCore/Utilities/MenuBarSummary.swift
@@ -220,24 +220,24 @@ public enum MenuBarSummary {
         case .netWorth:
             guard !accounts.isEmpty else { return PlaidBarConstants.appName }
             guard !privacyMaskEnabled else { return PrivacyMaskPresentation.heroValue }
-            return MultiCurrencyBalancePresentation.displayText(
+            return MultiCurrencyBalancePresentation.glance(
                 from: MultiCurrencyBalancePresentation.netWorth(accounts: accounts),
                 format: currencyFormat
-            )
+            ).text
         case .netCash:
             guard !accounts.isEmpty else { return PlaidBarConstants.appName }
             guard !privacyMaskEnabled else { return PrivacyMaskPresentation.heroValue }
-            return MultiCurrencyBalancePresentation.displayText(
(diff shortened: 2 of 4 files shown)


ftchvs:

**Verdict: minor / approve-worthy.** I tried to break the four load-bearing claims (single-USD byte-identical, no fabricated cross-currency `$`, dominant symbol correct, VoiceOver label present) and they hold under the diff + base code. Findings below are advisory; no must-fixes.

### Verified
- **Single-currency byte-identical (no regression to the dominant path).** For all-USD, `CurrencyAggregation.aggregate` hits the `subtotals.count <= 1` fast path → `.exact(_, .usd)`, so `headline.formattedTotal` is non-nil and `glance(...).text` returns it verbatim — structurally identical to the old `displayText` (`headline.formattedTotal ?? "By currency"`). The marker is only appended in the `formattedTotal == nil` branch, which single-currency never reaches. Guards for empty accounts / Privacy Mask / app-lock are unchanged and sit *upstream* of `glance`, so masked/locked/empty states stay byte-identical.
- **No fabricated cross-currency `$`.** Mixed + unpriceable → `.unavailable` → `formattedTotal == nil` → `dominantSubtotal` rendered via `Formatters.currency(amount, in: dominant.currency, …)`. No blind sum; the `7,000` test pins this.
- **Dominant symbol + by-absolute-value pick** are correct (`abs()` reduce; negative-debt test covers it). Marker `"+"` is text, not color — satisfies ACCESSIBILITY.md.
- **VoiceOver label present & honest.** `accessibleValueText` routes balance modes to `glance(...).accessibilityLabel`, which names the dominant currency in words and discloses omitted currencies; `menuBarAccessibilityLabel` threads it in. Masking parity holds (same `privacyMaskEnabled` guard + `appLockPreferences.menuBarText` wrapper as `menuBarText`).
- **Builds on merged AND-546 (#663) + #668 (both in `main`)** — not a reinvention. Single clean commit.

### Advisory (non-blocking)
1. **`menuBarHelpText` still uses `menuBarText`, not the accessible variant.** Likely intentional (tooltip should mirror the visible glyph), but worth a one-line comment to make the asymmetry with `menuBarAccessibilityLabel` deliberate rather than look like a missed call site.
2. **Single-USD test references `MenuBarSummary.netCash(from:)` while exercising mode `.netWorth`.** Those use *different* debt math (`abs(current)` vs `AccountPresentation.displayBalance`/`isDebt`). They coincide for this fixture (test passes), but the equality is somewhat coincidental. The genuinely structural guarantee is `glance.text == headline.formattedTotal == old displayText`; consider asserting against `MultiCurrencyBalancePresentation.displayText(from: netWorth(accounts:), …)` so the regression test can't silently drift if either debt formula changes.
3. **Pre-existing (out of scope, flagging for awareness):** `.netCash` mode routes through `netWorth(accounts:)` aggregation — present on `main` before this PR, not introduced here. Not this PR's job, but the "netCash uses netWorth" naming is a latent footgun.
4. **`glance`'s `privacyMaskEnabled` branch is effectively dead via the menu-bar caller** (upstream guard returns `heroValue` first). Harmless defense-in-depth; only matters if a future caller invokes `glance` without the upstream mask guard — keep it.
5. No test exercises the `.converted` (real offline-FX) branch of `glance`. The PR body notes it routes through `headline.formattedTotal` so it's covered by the exact path, but a 2-currency `StaticCurrencyConversionRates.sampleOffline` test would lock in "no `+` marker when a real conversion exists."

No correctness, privacy, concurrency, or color-only-meaning regressions found. Recommend merge after optionally tightening finding #2.

PR #674, opened 2026-06-25
feat(budgeting): rules manager UI — list/edit/reorder/delete (AND-551)

ftchvs:

## What & why

`TransactionRule`s are create-only inline in v1 — the only way to make one is to correct a transaction in the Review inbox (`createRule` / `InlineCategoryRulePrompt`). Once created there was no way to see, change, or remove them. This adds a management surface so the user can audit and curate the rules quietly steering their categorization and budgets.

Builds directly on the merged AND-546 schema/override-aware spend math: the conflict-resolution doctrine here is the *exact* precedence `EffectiveCategoryResolver.resolve` already applies (most recently created matching rule wins per field), surfaced verbatim so the UI's "this rule wins" copy can never drift from what actually happens to budgets.

## Change

Pure logic in `PlaidBarCore` (Sendable, unit-tested), thin renderer + store mutators in the app — per CLAUDE.md.

- **`PlaidBarCore/Utilities/TransactionRuleManager.swift`** (new, pure):
  - `sortedForDisplay` / `rows` — newest-first display order (= precedence order).
  - `matchDescription` + `effects` — privacy-safe text for each rule's match and effect. Meaning is in the text; icons/tints are decoration only (ACCESSIBILITY.md).
  - `shadowedFields` / `overlaps` — per-field conflict detection (AC #4). When two overlapping rules set the same field, the most recently created wins; the loser is flagged "overridden by the X rule", never hidden or deleted.
  - `provenance(for:in:)` — the winning rule per field for a given transaction, identical to the resolver's `firstRuleValue` precedence (AC #3).
  - `validate` / `applyingEdit` / `deleting` — editing preserves `id` + `createdAt` so an edit never silently re-ranks a rule; deleting only drops the matcher.
- **`Sources/PlaidBar/App/AppState.swift`** — `updateTransactionRule` / `deleteTransactionRule`: reuse the existing `transactionRules` store and serial persistence writer, undoable via the existing review-undo stack. Deleting a rule deliberately does **not** touch review metadata, so already-reviewed transactions stay reviewed (AC #2).
- **`Sources/PlaidBar/Settings/RulesSettingsView.swift`** (new) + a new "Rules" tab in `SettingsView` — list (match + effect chips + conflict warnings), add/edit sheet with live validation, delete with a confirmation that explicitly states reviewed transactions are not re-opened. Suppressed under Privacy Mask / App Lock.
- **`Sources/PlaidBarCore/Models/Route.swift`** — added `rules` to `SettingsRouteTab` to keep deep-link tab parity with the app's `SettingsTab`.

**Additive / byte-identical for non-opted-in users:** a user with no rules sees an empty state and nothing new is written to disk; the existing inline create-path and spend math are unchanged.

## Testing

```
# → Build complete! (no errors, no warnings)

swift test --filter TransactionRuleManagerTests
# → Test run with 21 tests passed (display order, effects/match copy, per-field
#   shadowing incl. token-containment overlap, provenance — cross-checked against
#   EffectiveCategoryResolver — validation, edit identity/normalization, deletion)

swift test --filter EffectiveCategoryResolverTests --filter RouteDeepLinkTests --filter TransactionReviewInboxTests
# → Test run with 75 tests passed (no regression from Route / AppState edits)

git diff --check
# → clean
```

## Links

Closes AND-551

## Follow-ups

- A deep-link / button from the transaction inspector's existing provenance popover into the matching rule row in this manager (the `provenance` API is already in place for it).
- Reordering as an *explicit* precedence override (drag-to-rank) is intentionally out of scope here — precedence is `createdAt`-derived and an edit preserves it; a manual rank field would be a schema addition under the AND-524 epic.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -4604,6 +4604,41 @@ final class AppState {
         persistReviewStorage()
     }
 
+    // MARK: - Rules management (AND-551)
+
+    /// Update (or insert) a categorization rule from the rules-manager surface.
+    ///
+    /// Routes through the pure `TransactionRuleManager.applyingEdit`, which
+    /// normalizes the rule's free-text fields and **preserves the existing rule's
+    /// `id` and `createdAt`** so an edit never silently re-ranks the rule relative
+    /// to its peers (conflict precedence is by `createdAt`). The change is undoable
+    /// (one ⌘Z restores the prior rule set) and persisted through the same serial
+    /// writer as every other review-storage mutation. Caches invalidate via
+    /// `transactionRules.didSet`.
+    ///
+    /// Editing a rule deliberately does **not** touch review metadata: re-running
+    /// the matcher over already-reviewed transactions is the resolver's job at
+    /// aggregation time, not a retroactive metadata rewrite here.
+    func updateTransactionRule(_ rule: TransactionRule) {
+        pushReviewUndoState()
+        transactionRules = TransactionRuleManager.applyingEdit(rule, to: transactionRules)
+        persistReviewStorage()
+    }
+
+    /// Delete a categorization rule from the rules-manager surface.
+    ///
+    /// Per AC #2, deleting a rule must **not** retroactively un-review past
+    /// transactions: this only drops the rule from the matcher set
+    /// (`TransactionRuleManager.deleting`), leaving every `.reviewed` /`.ignored`
+    /// metadata record exactly as it was. Undoable and persisted like every other
+    /// review mutation.
+    func deleteTransactionRule(id: UUID) {
+        guard transactionRules.contains(where: { $0.id == id }) else { return }
+        pushReviewUndoState()
+        transactionRules = TransactionRuleManager.deleting(ruleID: id, from: transactionRules)
+        persistReviewStorage()
+    }
+
     /// Marks a batch of inbox rows reviewed in a single, undoable step (AND-528).
     ///
     /// The blast radius — exactly which transaction ids resolve — is decided by the
--- a/Sources/PlaidBar/Settings/RulesSettingsView.swift
+++ b/Sources/PlaidBar/Settings/RulesSettingsView.swift
@@ -1,0 +1,382 @@
+import PlaidBarCore
+import SwiftUI
+
+/// Settings surface that lets the user manage categorization rules — list, edit,
+/// and delete the `TransactionRule`s that were previously create-only inline
+/// (AND-551, DEFERRED v2 / AND-524).
+///
+/// Thin renderer over `TransactionRuleManager`: that pure type owns the display
+/// order, the match/effect copy, and the conflict (shadowing) detection, so the
+/// view never re-derives "which rule wins". Mutations route through
+/// `AppState.updateTransactionRule` / `deleteTransactionRule`, which reuse the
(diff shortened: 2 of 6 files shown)


ftchvs:

**Verdict: minor — no must-fix blockers found. Two advisory items below; both are pre-existing-scope, not regressions introduced here.**

Verified the three primary review focuses against `main` and by building/testing the branch worktree:

- **Delete must not retroactively un-review** ✅ `AppState.deleteTransactionRule` calls only `TransactionRuleManager.deleting` (a pure `filter`) + `persistReviewStorage`; it never touches `transactionReviewMetadata`. The pure deletion is covered by `deleteRemovesOnlyTarget` / `deleteAbsent`. AC #2 holds.
- **Conflict resolution deterministic + shown** ✅ `sortedForDisplay` is byte-identical to `EffectiveCategoryResolver.resolve`'s sort (`createdAt` desc, then `id.uuidString` desc tiebreak), and `provenance` mirrors `firstRuleValue` per field. The `provenanceMatchesResolver` test cross-checks the winning category against the resolver directly — good defense against drift. Conflicts are surfaced (never hidden) via `FieldConflict` / `shadowWarning`.
- **Reuses existing store, no parallel store** ✅ Mutators write the existing `transactionRules` `@Observable` and route through the shared `pushReviewUndoState` / `persistReviewStorage`; undo snapshots already carry `(metadata, rules)`, so edit/delete are undoable with no new plumbing. Builds on merged AND-546 schema — not a reinvention.

Other checks: `swift build --target PlaidBarCore -strict-concurrency=complete -warnings-as-errors` is clean; `swift test --filter TransactionRuleManagerTests` = 21/21 green. Non-opted-in path writes nothing new to disk (no `.onAppear` writes; default `selectedTab` stays `.general`). Editor normalizes on save via `applyingEdit` → `normalized`, and validation runs on the normalized form, so validated and persisted state agree. `id`/`createdAt` preserved on edit so an edit never silently re-ranks (`editPreservesIdentity`).

**Advisory 1 — rule-derived rename (`merchantName`) is an unwired effect.** Neither `EffectiveCategoryResolver.resolve` nor `TransactionReviewInbox` ever reads `TransactionRule.merchantName`; the inbox derives the display name from `metadata?.userMerchantName`, not from a matching rule. This PR's editor now lets a user create a rule whose *only* effect is "Rename to X" (category off, transfer/exclude unset) and `validate` accepts it as effective — but that rename will never take effect anywhere. This is pre-existing (the field has been write-only since the inline create path), so not a blocker for AND-551, but consider either (a) wiring rule-derived rename into the resolver/inbox, or (b) not presenting standalone rename as a valid effect until it is wired, to avoid promising an effect the app doesn't apply. `shadowedFields` correctly omits the rename field (nothing to shadow since it's unresolved) — internally consistent.

**Advisory 2 — PR body slightly oversells "deep-link tab parity."** `SettingsRouteTab` has no consumer in the app target (no bridge sets `@AppStorage("settings.selectedTab")` from it); adding `case rules` keeps the documented raw-value parity invariant with `SettingsTab`, which is correct and harmless, but it does not by itself wire a deep link into the new tab. Raw values are strings so insertion order between the two enums is irrelevant. No action needed — just flagging the wording.

Nit: the title says "list/edit/reorder/delete" but reordering is explicitly out of scope (correctly — precedence is `createdAt`-derived); consider dropping "reorder" from the title.

PR #675, opened 2026-06-25
feat(review): confidence-sorted inbox + optional high-confidence auto-review (AND-553)

ftchvs:

## What & why

A cheap budgeting-v2 smart-sort for the Review Inbox plus an opt-in auto-review pass, both built on the `isLowConfidenceCategory` flag that already rides on `TransactionDTO` (server-derived from Plaid PFCv2 LOW/UNKNOWN confidence). The data exists; this just sorts on it and, optionally, clears the safest rows for the user.

Strictly additive: a user who never touches the new sort control or settings toggle sees byte-identical prior behavior — the default sort is the historical order and auto-review is OFF by default.

## Change

**PlaidBarCore (pure, `Sendable`, unit-tested):**
- `ReviewInboxSorting` / `ReviewInboxSortOrder` — `.priority` (default; the historical urgency → newest → id comparator extracted verbatim) and `.confidenceLowFirst` (floats Plaid LOW/UNKNOWN categorizations to the top, with the historical order as the in-band tiebreaker so nothing else reorders). `TransactionReviewInbox.evaluate` gains a `sortOrder` parameter defaulting to `.priority`.
- `ReviewAutoReviewPolicy` / `ReviewAutoReviewPreference` — conservative eligibility: a row is auto-reviewable only when high-confidence, categorized (not `.other`), NOT an on-device suggestion, ordinary value, and **never** a transfer (by `isTransfer` flag, `.possibleTransfer` reason, or transfer category) nor carrying any high-priority reason. Preference defaults OFF.
- `TransactionReviewMetadata` gains an additive `autoReviewed` flag (decodes as `false` when absent — pre-AND-553 records behave identically); `TransactionReviewItem` surfaces it as `wasAutoReviewed`.

**App:**
- `AppState`: UserDefaults-backed `reviewInboxSortOrder` (invalidates the cached snapshot on change) and `autoReviewHighConfidenceEnabled`. `applyHighConfidenceAutoReview()` marks the policy's eligible ids reviewed via the *same* approve path as a manual review, in **one** undoable batch (so a single ⌘Z / undo reverts the whole pass), stamping `autoReviewed = true`. Every manual review/ignore/recategorize/rename/transfer path clears the flag. The pass runs when the toggle is switched on and whenever new transactions sync (a no-op while disabled).
- `ReviewInboxView`: a confidence-aware sort control (label text + glyph, never color alone) and an "Auto-reviewed" badge on reopened auto-reviewed rows, folded into the VoiceOver row summary.
- `Settings → General → Review Inbox`: the auto-review toggle with explanatory copy.

**Security / privacy / a11y:** all new logic is in `PlaidBarCore` (pure, `Sendable`); no secrets touch the app/Core boundary. The sort control and badges are suppressed under Privacy Mask / App Lock like the rest of the inbox. Meaning is always carried by text + glyph, never color alone.

## Testing

Exact commands and results (run with the sandbox disabled per repo convention):

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete!**
- `swift test` → **Test run with 2411 tests passed** (37 new across 3 suites: `ReviewInboxSortingTests`, `ReviewAutoReviewPolicyTests`, `ReviewAutoReviewMetadataTests`).
- `git diff --check` → clean.

New tests assert: low-confidence-first ordering and the historical default order are byte-identical to before; auto-review **never** clears transfers (flag/reason/category), unusual, low-confidence, uncategorized/Other, NL-suggested, or already-resolved rows; the eligible-id set is exactly the safe rows; the additive `autoReviewed` field round-trips and a legacy record (no key) decodes as `false`; and `evaluate` threads the flag onto `wasAutoReviewed`.

## Links

Closes AND-553

## Follow-ups

- The "Auto-reviewed" badge surfaces only on rows that *reopen* after auto-review (a freshly auto-reviewed row leaves the inbox like any reviewed row). A future power-review-table column could list auto-reviewed rows for at-rest inspection if desired.
- A one-shot confirmation banner ("Auto-reviewed N transactions — Undo") on the inbox could make the bulk-undo even more discoverable; deferred to keep this diff scoped.

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -59,6 +59,8 @@ final class AppState {
         static let windowFirstOrientationDismissedContextPrefix = "windowFirstOrientation.dismissed.context"
         static let lastTransactionCacheContext = "cache.lastTransactionCacheContext"
         static let categoryBudgetCache = "cache.categoryBudgets"
+        static let reviewInboxSortOrder = ReviewInboxSortOrder.storageKey
+        static let reviewAutoReviewHighConfidence = ReviewAutoReviewPreference.storageKey
     }
 
     // MARK: - State
@@ -107,6 +109,42 @@ final class AppState {
         }
     }
     var hasLoadedTransactionReviewStorage = false
+
+    /// How the Review Inbox orders its rows (AND-553). UserDefaults-backed so the
+    /// choice persists, and changing it invalidates the memoized snapshot so the
+    /// next read re-sorts. Defaults to `.priority` — the historical order — so a
+    /// user who never touches the control sees today's ordering exactly. Seeded
+    /// from UserDefaults in `init` (below) so a stored choice is honored at boot.
+    var reviewInboxSortOrder: ReviewInboxSortOrder = .priority {
+        didSet {
+            guard reviewInboxSortOrder != oldValue else { return }
+            _cachedTransactionReviewInboxSnapshot = nil
+            UserDefaults.standard.set(reviewInboxSortOrder.rawValue, forKey: Keys.reviewInboxSortOrder)
+        }
+    }
+
+    /// Whether the opt-in high-confidence auto-review pass is enabled (AND-553).
+    /// Off by default: a not-opted-in user has nothing resolved on their behalf,
+    /// matching today's behavior byte-for-byte. UserDefaults-backed so the choice
+    /// persists; seeded in `init`. Turning it ON triggers a one-shot
+    /// `applyHighConfidenceAutoReview()` so the inbox immediately reflects the
+    /// choice; turning it OFF stops future auto-review but leaves already-cleared
+    /// rows as they are (they remain ⌘Z / bulk-undoable).
+    var autoReviewHighConfidenceEnabled: Bool = false {
+        didSet {
+            guard autoReviewHighConfidenceEnabled != oldValue, !isLoadingReviewPreferences else { return }
+            UserDefaults.standard.set(autoReviewHighConfidenceEnabled, forKey: Keys.reviewAutoReviewHighConfidence)
+            if autoReviewHighConfidenceEnabled {
+                applyHighConfidenceAutoReview()
+            }
+        }
+    }
+
+    /// Suppresses the `autoReviewHighConfidenceEnabled` didSet side effects while
+    /// seeding the stored value at boot, so loading the persisted preference does
+    /// not itself trigger an auto-review pass before data is loaded.
+    private var isLoadingReviewPreferences = false
+
     var isLoading = false
     /// True from launch until the first `loadInitialData()` pass completes.
     /// While booting, data surfaces render loading/skeleton states instead
@@ -827,6 +865,7 @@ final class AppState {
         loadWatchlistTargets(defaults: defaults)
         loadAppLockPreferences(defaults: defaults)
         loadLocalAISettings(defaults: defaults)
+        loadReviewPreferences(defaults: defaults)
         // Balance history
         loadPersistedBalanceHistory()
         loadPersistedWeeklyReviewState()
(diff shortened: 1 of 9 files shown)


ftchvs:

**Verdict: LGTM (minor)** — no must-fixes. Strictly-additive, well-tested, and the safety-critical eligibility contract holds. Verified locally on `origin/feat/and-553-confidence-sorted-inbox`: strict-concurrency build (`-strict-concurrency=complete -warnings-as-errors`) → **Build complete!**; the three new suites pass (22 test cases green). Two advisory notes below; neither blocks merge.

### What I checked (all pass)
- **Auto-review never touches transfers / unusual / large / low-confidence / Other / NL-suggested / already-resolved rows.** `ReviewAutoReviewPolicy.isAutoReviewable` (`Sources/PlaidBarCore/Utilities/ReviewAutoReviewPolicy.swift`) gates on all of: `status == .needsReview`, `!isLowConfidenceCategory`, real non-`.other` category, `!isNLSuggestedCategory`, `!isTransfer`, `!.possibleTransfer`, `!isTransferCategory`, `!.unusualAmount`, and a defense-in-depth `!reasonCodes.contains(where: \.isHighPriority)`. Note `isHighPriority` (verified on `main`: priority-0 reasons + `.unusualAmount`) already subsumes the explicit `.unusualAmount`/`.possibleTransfer` guards — harmless redundancy, intentional belt-and-suspenders. Covered by `ReviewAutoReviewPolicyTests` including flag/reason/category transfer paths.
- **Bulk-undo restores exactly.** `applyHighConfidenceAutoReview` (`AppState.swift`) pushes exactly one `pushReviewUndoState()` snapshot, then loops `approveReviewItemWithoutUndo(_, autoReviewed: true)` (which deliberately does *not* push per-row). `undoLastReviewAction` restores the whole `transactionReviewMetadata` array wholesale → one ⌘Z reverts the entire pass to the exact prior state (including `autoReviewed=false`). Correct.
- **Sort stability.** Both `.priority` and the `.confidenceLowFirst` in-band fallback use `priorityLess` with a total `id <` tiebreaker, so ordering is deterministic regardless of `sorted(by:)` not being a stable sort. Transaction ids are unique → no ambiguity.
- **Not-opted-in / no-new-data path byte-identical.** `sortOrder` defaults to `.priority` (verbatim-extracted historical comparator; `evaluateDefaultSortUnchanged` asserts implicit==explicit). `applyHighConfidenceAutoReview` early-returns on `!autoReviewHighConfidenceEnabled`, so the new call appended to `seedReviewMetadataForNewTransactions` is a pure no-op when disabled. `autoReviewed` decodes as `false` when absent (legacy-record test passes).
- **Builds on merged AND-546 schema, not a reinvention.** `isLowConfidenceCategory`, `EffectiveCategoryResolver.isTransferCategory`, and `isHighPriority` all already exist on `origin/main`. This PR consumes them; it does not redefine them.
- **Privacy/a11y.** Sort control gated on `!shouldMaskFinancialValues && totalCount > 0`; rows (and thus the auto-reviewed badge) are replaced by `privateInboxPlaceholder` under mask, so nothing leaks. Badge + sort labels pair glyph **and** text (never color alone); `wasAutoReviewed` is folded into the VoiceOver row summary. Good.

### Advisory (non-blocking)
1. **Background auto-review displaces the manual undo target.** Auto-review runs on every sync via `seedReviewMetadataForNewTransactions` and pushes onto the *same* single-level `reviewUndoStack` the "Undo last review" action pops (no redo stack; capped at 20). So after a sync auto-clears rows, the user's next Undo reverts the *sync's* auto-review batch rather than their last manual action — and the manual entry can be buried or evicted. The batch being undoable is arguably desirable, but it silently changes what ⌘Z does after a sync. Consider whether the silent sync-time pass should push an undo snapshot at all, or whether the deferred "Auto-reviewed N — Undo" banner should own that affordance.
2. **No surfaced trace for freshly auto-reviewed rows (acknowledged in the PR body).** A row auto-cleared on sync leaves the inbox with zero indication it was automatic vs. manual; the `wasAutoReviewed` badge only appears if the row later *reopens*. The deferred confirmation banner / at-rest review-table column would close this discoverability gap.

Do not merge on my say-so — advisory only.

PR #676, opened 2026-06-25
feat(budgeting): custom/renamed/recolored categories + groups editor (AND-547)

ftchvs:
## What & why

**DEFERRED v2 (AND-524 epic) — re-opens scope cut from v1.** The maintainer authorized the deferred budgeting-v2 epic.

AND-547 adds the editor to create/rename/recolor/re-emoji/reorder categories and define/reorder groups, built on the **merged AND-546 schema** (#663, `199bca9` — persisted `Category`/`CategoryGroup`/`Budget(month, rollover)` tables + override-aware spend math). It is **additive and opt-in**: a user who does not enable custom categories sees **byte-identical** standard budgeting (no v2 snapshot is written, the v1 `/api/budgets` path is untouched, and opting out restores v1 losslessly).

## Change

**`PlaidBarCore` (pure, `Sendable`, unit-tested) — all logic lives here:**
- **`BudgetCategoryEditor`** — stateless CRUD/validation over a `BudgetingV2Schema` (value semantics; no hidden `Date()`/UUID/I/O — identity is passed in). Implements the four ACs:
  1. **Create a custom category** with name + emoji + color + group (custom ids are `cat_<uuid>` so they never collide with a Plaid `rawValue`).
  2. **Rename / recolor / re-emoji a SYSTEM category without breaking its `systemKey → Plaid` mapping** — `seededFromCategory` is preserved verbatim; only presentation fields change, so every already-categorized transaction still buckets to the same Plaid key.
  3. **Create / rename / reorder groups** and **move categories between groups**.
  4. **Delete a custom category, reassigning its budgets** to a fallback (default: the always-present `.other` system category), **summing** colliding month/category limits so no budgeted dollar disappears — **no orphaned spend**. The returned `DeletionResult` carries the `deleted → reassigned` mapping so a caller can re-point transaction-level overrides the same way. System categories/groups are never deletable.
- **`BudgetingV2Schema`** — additively extends `BudgetCategoryV2` with `emoji`, `colorHex`, and a within-group `sortIndex`, plus a derived `isCustom` on both category and group. A **custom decoder defaults the new keys**, so a v2 snapshot written *before* AND-547 still decodes (self-healing, no destructive reseed).
- **`BudgetingV2Migration.seed`** — stamps each seeded leaf with a contiguous per-group `sortIndex` so the editor has a stable starting order to reorder against.

**`PlaidBar` (app UI):**
- **`CategoryEditorModel`** (`@Observable @MainActor`) — owns the opt-in `BudgetingV2Store`, routes every mutation through `BudgetCategoryEditor`, persists best-effort (in-memory schema updates regardless so the UI reflects an edit even if a disk write fails — the snapshot is disposable). Opt-in carries v1 budgets forward; opt-out recovers them and clears the snapshot.
- **`CategoryEditorView` + `CategoryFormSheet`** — a new Settings **"Categories"** tab: opt-in prompt, per-group sections, drag-to-reorder, move-to-group, an accessible named-swatch color picker, and emoji field. Identity is always carried by **name + glyph, never color alone**; errors are text. Financial labels inherit the shared Privacy Mask / App Lock gating via the Settings scene.
- **`AppState+CategoryEditor`** — builds the model scoped to the active Plaid environment via the existing `readModelCacheKey()` seam, behind `try?` (a store failure leaves the editor read-only and the app on v1).

The server, Plaid client, and Keychain/localhost boundary are unchanged. No secrets touch the app/Core. Synthetic/demo data only.

## Testing

```
swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors   # Build complete, 0 errors/warnings
swift test --filter BudgetCategoryEditorTests          # 20 passed (4 ACs + validation + additivity)
swift test --filter BudgetingV2SchemaBackCompatTests   # 4 passed (pre-AND-547 snapshot still decodes; round-trip)
swift test --filter BudgetingV2EditorStoreTests        # 2 passed (edited schema persists; opt-out recovers v1)
swift test --filter CategoryBudget --filter Budget     # 202 passed (no regression)
git diff --check                                       # clean
```

Key guarantees covered by tests: a renamed system category keeps `seededFromCategory`/Plaid key and still reverse-migrates to v1; deleting a custom category conserves total budgeted dollars (sums collisions); a pre-AND-547 snapshot decodes with the new fields defaulted; seeded categories keep their v1 chart colors and a contiguous per-group order.

## Links

Closes AND-547

## Follow-ups

- Wiring the `DeletionResult` reassignment mapping into the transaction-override store so a custom-category delete also re-points `TransactionReviewMetadata.userCategory` references (the pure mapping is returned and tested; the app-side re-point is a small additional seam, deferred to keep this diff scoped).
- Per-month budgets / rebalance / splits / rules manager / sync (AND-548…AND-553) continue to build on this editor + schema.

### DIFF
--- a/Sources/PlaidBar/App/AppState+CategoryEditor.swift
+++ b/Sources/PlaidBar/App/AppState+CategoryEditor.swift
@@ -1,0 +1,44 @@
+import Foundation
+import PlaidBarCache
+import PlaidBarCore
+
+/// AppState wiring for the budgeting-v2 categories & groups editor (AND-547 —
+/// deferred epic AND-524).
+///
+/// Reuses the AND-546 seams: it opens the opt-in ``BudgetingV2Store`` against the
+/// active data directory and scopes it with the same ``readModelCacheKey()`` the
+/// other disposable caches use, so the v2 schema is per-Plaid-environment. Like the
+/// read-model and transaction caches the store is opened behind `try?`; a failure
+/// leaves the editor empty/read-only and the app on v1 budgeting (additive,
+/// fallback-safe).
+extension AppState {
+    /// Builds the editor model for the active environment, or a read-only model
+    /// (`store: nil`) when v2 is unavailable (demo mode, no server context yet, or
+    /// the store fails to open). The model loads its snapshot on appear, so a fresh
+    /// instance per editor presentation is fine and keeps no long-lived disk handle.
+    ///
+    /// The store open is best-effort and never throws to the caller; a `nil` store
+    /// yields an editor that can render the opt-in prompt but performs no I/O.
+    @MainActor
+    func makeCategoryEditorModel() -> CategoryEditorModel {
+        guard !isDemoMode, let cacheKey = readModelCacheKey() else {
+            // No environment context yet (or demo): editor stays read-only and v1
+            // budgeting is untouched. A non-nil cacheKey is required so a seeded
+            // snapshot is scoped; without one there is nothing to read or write.
+            return CategoryEditorModel(store: nil, cacheKey: "")
+        }
+        let store = try? BudgetingV2Store(onDiskIn: activeStorageDirectoryURL)
+        return CategoryEditorModel(store: store, cacheKey: cacheKey)
+    }
+
+    /// The current month as `YYYY-MM`, used to carry v1 budgets forward at opt-in and
+    /// recover them on opt-out. Uses the same date convention the rest of the app
+    /// sorts/keys budgets by.
+    @MainActor
+    func currentBudgetMonthKey(asOf date: Date = Date(), calendar: Calendar = .current) -> String {
+        let components = calendar.dateComponents([.year, .month], from: date)
+        let year = components.year ?? 1970
+        let month = components.month ?? 1
+        return String(format: "%04d-%02d", year, month)
+    }
+}
--- a/Sources/PlaidBar/Services/CategoryEditorModel.swift
+++ b/Sources/PlaidBar/Services/CategoryEditorModel.swift
@@ -1,0 +1,251 @@
+import Foundation
+import OSLog
+import PlaidBarCache
+import PlaidBarCore
+
+/// App-side state for the budgeting-v2 categories & groups editor (AND-547 —
+/// deferred epic AND-524).
+///
+/// Owns the opt-in ``BudgetingV2Store`` and the in-memory ``BudgetingV2Schema`` the
+/// editor view renders, and routes every mutation through the pure
+/// ``BudgetCategoryEditor`` in `PlaidBarCore` (all validation/CRUD logic is there
+/// and unit-tested; this model only persists the result and surfaces errors).
+///
+/// ## Opt-in & additive
+/// The editor is **opt-in**: until the user explicitly opens it and seeds v2, no v2
+/// snapshot exists and v1 budgeting is byte-for-byte unchanged. `load()` only reads
+/// an existing snapshot; `optIn()` seeds one (carrying any v1 budgets forward).
+/// Opting out (`optOut(month:)`) recovers the v1 budgets and clears the snapshot,
+/// restoring v1 untouched.
+///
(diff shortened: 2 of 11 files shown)


ftchvs:

**Verdict: LGTM with minor advisories.** The four focus risks all hold up under scrutiny, the change is genuinely additive/opt-in, and it builds on the merged AND-546 schema/store rather than reinventing it. Two small invariant/UX gaps and one gating note below — none block merge.

### Focus-area verification (all pass)
- **systemKey → Plaid mapping survives rename/recolor/re-emoji** ✅ `editCategory` (`BudgetCategoryEditor.swift`) rewrites only presentation fields and re-emits `seededFromCategory` + `id` verbatim. Covered by `editSystemCategoryPreservesMapping` and the back-compat round-trip. Renaming a system category cannot reclassify a transaction.
- **Delete reassigns, never orphans** ✅ `deleteCustomCategory` folds each `(month, deleted)` budget onto `(month, target)`, **sums** colliding limits, OR's rollover, defaults the target to the always-present non-deletable `.other`, rejects self/missing targets, and returns the `deleted → reassigned` mapping for transaction-override re-pointing. System categories are hard-blocked via `isCustom` (derived from `seededFromCategory`, so it can't be spoofed). Tested in `deleteCustomCategoryReassignsBudgets` / `DefaultFallback` / `InvalidTarget` / `systemCategoryNotDeletable`.
- **No silent SpendingCategory case loss** ✅ N/A — the diff contains no exhaustive `switch` over `SpendingCategory`; identity flows through `rawValue`/`seededFromCategory`, and custom ids use the `cat_`/`grp_` prefixes so they never collide with a Plaid `rawValue`.
- **Byte-identical / opt-in** ✅ Every mutation is reachable only after `optIn` seeds a v2 snapshot; `makeCategoryEditorModel` short-circuits demo mode and a missing env to a read-only `store: nil` model; the store opens behind `try?`; v1 `/api/budgets` + `CategoryBudgetDTO` are never written. App-side symbols (`readModelCacheKey`, `activeStorageDirectoryURL`, `categoryBudgets`, actor `BudgetingV2Store`) all exist on `main`; `await store.*` is correct against the `actor`.

### Minor advisories (non-blocking)
1. **`deleteCustomGroup` skips the duplicate-name guard that `moveCategory` enforces** (`BudgetCategoryEditor.swift`, ~L452). Folding a custom group's categories into the fallback group blindly appends them — if the custom group held a "Misc" and `.other` already has a "Misc", you get two same-named categories in one group, an invariant `moveCategory` (L242 `hasDuplicateCategoryName`) explicitly prevents. No data/budget loss, but the post-delete schema can violate the per-group unique-name rule. Consider de-duping (e.g. suffix or merge) on fold, or at least a test asserting the resulting state.
2. **Opt-out silently drops live custom-category budgets.** `reverseToV1Budgets` only recovers budgets whose category has a `seededFromCategory`; a budget the user set on a *custom* category has no v1 home and is dropped (documented as intentional). That's defensible, but it's real dollars vanishing from the user's point of view. The opt-out UI copy ("restores your saved budgets") doesn't warn that custom-category budgets won't come back — recommend a one-line caveat in the opt-out footer/confirmation.
3. **Privacy Mask / App Lock gating is asserted in prose, not in this view.** `CategoryEditorView`/`CategoryFormSheet` render category labels + (eventually) limits — financial data — but rely entirely on the Settings scene environment for masking. Worth a QA-matrix line or a test confirming the Categories tab is actually masked/locked, since it's a new financial surface.

### Notes (no action)
- `isValidEmoji`'s `scalar.value >= 0x203C || isEmojiPresentation` correctly rejects ASCII digits/`#`/`*` (which carry `isEmoji` but not emoji presentation). Fine.
- `reorderCategories`/`reorderGroups` stable-sort unlisted ids after listed ones, so a partial reorder never drops a row. Good.

Test coverage (20 + 4 back-compat + 2 store) directly exercises every focus path. No must-fixes.

PR #677, opened 2026-06-25
feat(review): opt-in server-synced review state (AND-552)

ftchvs:
**DEFERRED v2 (AND-524 epic) — re-opens scope cut from v1.** The maintainer authorized re-opening this cloud surface, approved **only as strictly opt-in**.

## What & why

VaultPeek is local-first: the transaction **review state** — the per-transaction overrides a user sets in the Review Inbox (chosen category, merchant rename, transfer flag, budget exclusion, free-text note, review status) plus their categorization rules — is app-local JSON only, and the local server never sees it. AND-552 adds an **optional** multi-device sync of exactly that state through a new `/api/review` surface so a user can carry their review work across their own machines.

This re-opens a cloud surface, so it is **strictly opt-in**. With the toggle OFF (the default for a fresh install and every existing user), the app never constructs or sends a review snapshot, the new tables stay empty, and behavior is **byte-identical** to before AND-552 — nothing leaves the device without explicit consent.

## Change

Built on the merged AND-546 schema/security split; no Plaid/Keychain/localhost boundary changed.

- **PlaidBarCore (pure, Sendable):**
  - `ReviewSyncDTO` — `ReviewStateSnapshotDTO` + per-record envelopes (`ReviewMetadataRecordDTO`, `ReviewRuleRecordDTO`), each carrying an `updatedAt` LWW clock. Schema-versioned like `BudgetingV2Schema`.
  - `ReviewStateConflictResolver` — **conflict resolution = per-record last-writer-wins** keyed on `updatedAt`: distinct ids are unioned; for the same id the newer write wins (tie → the upload). Pure and deterministic, so a merge never invents a category the user did not pick (that's why LWW per-record, not per-field).
  - `ServerSyncedReviewFeatureFlag` — the opt-in flag, **default OFF** (mirrors `WindowFirstFeatureFlag`; `--server-synced-review on|off` CLI/QA override).
  - `ReviewSyncGate` — the pure "opt-out ⇒ skip, no network" decision, kept in Core so the security-critical off-path is unit-testable (the app target is `@main`/untestable).
  - `ReviewSyncAssembler` — in-memory review state ⇄ wire snapshot. `ServerEndpoint.reviewStateURL`.
- **PlaidBarServer:** `ReviewMetadataModel` / `ReviewRuleModel` Fluent models + **additive** `CreateReviewState` migration (creates two new tables, touches nothing existing); `ReviewStateStore` actor (`snapshot` / LWW `merge` / `clearAll`); `ReviewRoutes` (`GET`/`PUT`/`DELETE /api/review`) behind the **existing `APITokenMiddleware`**, with a 4 MiB body cap and a schema-version reject. Wired in `App.swift`.
- **PlaidBar app:** `ServerClient` review CRUD; `ReviewSyncService` gated through `ReviewSyncGate`, so the off path makes no request.
- **Docs:** `SECURITY.md` trust-boundary entry (the server stores user category overrides/renames/notes — display-safe, never a Plaid token — only when enabled; opt-out clears it); README Server API Reference rows for `/api/review`.

## Testing

Exact commands and results (Swift commands run with the sandbox disabled, per repo memory):

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete** (the only warning is the pre-existing, unrelated `async-http-client` unused-dependency note).
- `swift test --filter "ReviewStateSyncTests" --filter "ReviewStateStoreTests"` → **23 tests passed**. Covers: LWW (distinct-id union, newer-wins both directions, tie→incoming, idempotent re-merge, deterministic ordering); wire-DTO JSON round-trip; opt-in flag default OFF + resolution; **sync gate skips when not opted in (opt-out sends nothing)**; assembler round-trip; store CRUD; `/api/review` PUT/GET/DELETE merge round-trip through the wired route; auth-required; malformed-body + future-schema reject.
- `swift test` (full suite) → **2412 tests passed** (server target 194, core target 2100, all green — assembler move + service add caused no regression).
- `git diff --check` → clean.

## Links

Closes AND-552.

## Follow-ups

- Wire `ReviewSyncService` into `AppState`'s review-action path + a Settings opt-in toggle with a consent prompt (the gate, client, and assembler are in place; the AppState/Settings UI hookup is the remaining glue and a natural sibling of AND-553's rules-manager UI).
- Optional: a periodic/background `pull` + local-vs-server reconciliation pass for near-real-time multi-device convergence (today convergence happens on the next explicit push/pull).

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,6 +596,14 @@ when the local app created the Hosted Link session.
 | `GET` | `/api/budgets` | List saved category budgets (display-safe monthly limits only) |
 | `PUT` | `/api/budgets/:category` | Upsert a category's monthly limit (body `{ "monthlyLimit": Double }`); income/transfer categories are rejected |
 | `DELETE` | `/api/budgets/:category` | Remove a category's budget |
+| `GET` | `/api/review` | Pull the opt-in synced review-state snapshot (overrides + rules); empty unless the user enabled server-synced review. Display-safe values only — no Plaid tokens |
+| `PUT` | `/api/review` | Upload a device review-state snapshot; the server merges it (per-record last-writer-wins) and returns the union. Opt-in only |
+| `DELETE` | `/api/review` | Clear all synced review state (opt-out / reset) |
+
+> `/api/review` is the **opt-in** server-synced review surface (AND-552). It is
+> off by default (`ServerSyncedReviewFeatureFlag`); with sync off nothing is sent
+> or stored and behavior is byte-identical to local-first. See `SECURITY.md` for
+> the trust-boundary change (the server then stores user category overrides).
 
 ## Roadmap
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -108,6 +108,32 @@ Use GitHub private vulnerability reporting if available, or contact the reposito
   Keychain support can store token bytes in SQLite under `PLAIDBAR_DATA_DIR`.
   To keep the derived cache and any fallback token rows on-device, point
   `PLAIDBAR_DATA_DIR` at a local, non-synced directory.
+- **Trust-boundary change — opt-in server-synced review state (AND-552).** By
+  default the local server **never** stores transaction *review state* — the
+  per-transaction overrides a user sets in the Review Inbox (chosen category,
+  merchant rename, transfer flag, budget exclusion, free-text note, review
+  status) and their categorization rules. That state is app-local JSON only.
+  AND-552 adds an **optional** multi-device sync that, *only when the user
+  explicitly opts in*, persists exactly that review state on the local server
+  (the `review_metadata` / `review_rules` SQLite tables, served by
+  `/api/review` behind the same bearer-token middleware as every other `/api`
+  route). This deliberately **expands the server's stored data surface beyond
+  Plaid limits and token references**: with sync on, the server SQLite holds the
+  user's category overrides, merchant renames, and notes — display-safe derived
+  values, **never** a Plaid token, access secret, or Keychain material. The
+  opt-in is gated by `ServerSyncedReviewFeatureFlag` (default **OFF**); with it
+  off — the default for a fresh install and every existing user — the app never
+  constructs or sends a review snapshot, the `/api/review` tables stay empty,
+  and behavior is byte-identical to before AND-552, so **no review data leaves
+  the device without explicit consent**. Like the read-model cache, the synced
+  tables live in the environment-scoped SQLite under `PLAIDBAR_DATA_DIR`
+  (default local `~/.vaultpeek/`), so pointing the data dir at a cloud-synced
+  folder can copy this review state off-device (see the `PLAIDBAR_DATA_DIR`
+  caveat above). Opting back out clears the server-side tables
+  (`DELETE /api/review`). Conflicting edits across devices resolve by per-record
+  last-writer-wins on each record's `updatedAt` clock
+  (`ReviewStateConflictResolver`), so a sync never invents a category the user
+  did not choose.
 - `/api/status` is authenticated and exposes readiness metadata: version,
   environment, credential availability, storage path, linked item count,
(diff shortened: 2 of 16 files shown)


ftchvs:

**Verdict: LGTM with minor advisory notes.** I tried hard to break the opt-out/byte-identical and consent claims and could not. The default-off path is genuinely inert, conflict resolution is pure and deterministic, the trust-boundary doc is solid, and this builds on the merged AND-546 schema rather than reinventing it. No must-fixes. Notes below are nice-to-haves; do not block merge on them.

### What I verified (the load-bearing claims hold)
- **Opt-out sends nothing / byte-identical behavior — TRUE.** `ServerSyncedReviewFeatureFlag.defaultValue == false`; every read/write in `ReviewSyncService` guards on `ReviewSyncGate.allowsNetwork(isOptedIn:)`, returning `nil` without touching `ServerClient` when off. Critically, `ReviewSyncService` has **zero callers** anywhere (grep across `Sources/PlaidBar/`) — it is not wired into AppState/startup/any timer, so the off-path is not just gated, it is entirely unreached. Covered by `gateSkipsWhenNotOptedIn` / `gateDefaultSkips`.
- **Consent gating — adequate at the plumbing layer.** The flag default + gate enforce off-by-default. Note the user-facing consent prompt/Settings toggle is explicitly deferred (PR "Follow-ups"); acceptable here only because nothing invokes the service yet. The consent UI must land before any auto-push wiring — please keep AND-553's hookup blocked on it.
- **Flag parity — correct.** `ServerSyncedReviewFeatureFlag` mirrors `WindowFirstFeatureFlag` exactly (same `object(forKey:)` nil-check, same CLI>stored>default precedence) with `defaultValue=false`; unparseable CLI tokens never silently enable.
- **Conflict resolution — deterministic.** Pure per-record LWW on `updatedAt`, tie→incoming, results sorted by `id` / `uuidString`. Tested both directions + tie + idempotent + distinct-union.
- **Builds on AND-546, not a reinvention.** `BudgetingV2Schema` is on `main`; `ReviewStateStore`/`ReviewRoutes`/`CreateReviewState`/`ReviewStateStoreTests` mirror the `BudgetStore`/`BudgetRoutes`/`CategoryBudget*` patterns 1:1. Additive migration touches no existing table. Route sits behind the existing `APITokenMiddleware` (auth test present).
- **No new status leak.** `/api/status` is unchanged; it exposes nothing about review sync state.

### Advisory findings (minor, non-blocking)
- **[minor] "byte-identical" is behavioral, not on-disk.** `CreateReviewState` runs on **every** server (App.swift:56), so even never-opted-in users get two empty `review_metadata`/`review_rules` tables + a `_fluent_migrations` row. The claim is true for behavior/network/no-data-egress, but the SQLite file schema does change for all users. This matches the accepted additive-migration precedent (BudgetStore did the same) — just consider tightening the wording to "behaviorally byte-identical; additive empty tables only."
- **[minor] `ReviewRoutes.decodeSnapshot` lacks a lower schema bound.** It rejects `schemaVersion > current` but accepts `0`/negative. Not exploitable (auth-gated, display-safe data), but a `>= 1` floor would fail closed symmetrically. (`ReviewStateModel.swift` / `ReviewRoutes.swift`)
- **[minor] undecodable-row handling in `ReviewStateStore`.** `snapshot()` silently drops rows whose payload won't decode (mirrors `BudgetStore`), and `merge()` derives `base` from `snapshot()` — so a future-schema stored row is invisible to the merge and can be silently overwritten or orphaned. Benign at v1; worth a comment for when schemaVersion bumps.
- **[note, not a bug] `clearReviewState()` is intentionally not flag-gated** so opt-out teardown works when the flag is off. Correct and documented — flagging only so a future reviewer doesn't "fix" it.

Nice work on keeping the security-critical off-path in a pure, unit-testable `ReviewSyncGate`.

PR #678, opened 2026-06-25
feat(budgeting): transaction splits — one transaction into N category allocations (AND-550)

ftchvs:

## What & why

AND-399 explicitly cut transaction splits from v1 as "invasive to the transaction model." This PR re-opens that scope under the deferred budgeting-v2 epic (AND-524): split a single transaction into N category allocations (e.g. a Target run = Groceries + Home), so spend lands in the right buckets instead of one catch-all category.

The hard constraint is **additivity**: this touches every consumer of `TransactionDTO` spend (budgets, summaries, exports), so a user who does not split anything must see **byte-identical** prior behavior. The implementation is a child-allocation model layered on top of the parent transaction — the parent is never mutated, re-keyed, or removed.

## Change

All logic lives in `PlaidBarCore` (pure, `Sendable`, no I/O, no hidden `Date()`):

- **Model — `Models/TransactionSplit.swift`**: `TransactionSplit` (keyed by parent `transaction_id`) + `TransactionSplitAllocation` (own category + own signed amount + own exclude flag). The defining **sum invariant** — allocations sum to the parent amount within a cent — is enforced by `isBalanced(tolerance:)` / `isValid`. An empty or unbalanced split is not valid.
- **Resolver — `Utilities/TransactionSplitResolver.swift`**: the single additive seam every spend consumer routes through. `spendRows(for:)` returns:
  - **no split** → exactly one parent row, byte-identical to today;
  - **valid split** → one row per allocation, so **rollups count the parts, not the parent** (total spend is conserved because the parts sum to the parent);
  - **malformed split** → falls back to the parent row, so **editing/removing a split restores the original** automatically.
- **Split-aware spend** (optional `splits:` param, defaults to empty):
  - `CategoryBudgetPlanner` — `overrideAwareSpend` / `presentation` / `mergedPresentation` / internal `netSpendByCategory`.
  - `SpendingSummary` — `spendingByCategory` / `periodSummary`.
  - A split allocation buckets by its **own** declared category (the user chose it explicitly when splitting, so it bypasses per-parent override/rule resolution), honors its own exclude flag, and never counts income/transfer categories.
- **Exports respect splits** — `DataExportBuilder.splitsCSV` (one row per allocation, joined by `transaction_id`; the transactions CSV keeps its one-row-per-transaction contract) and an additive `splits` array in the JSON envelope (`schemaVersion` bumped 2 → 3; decodes as empty from older backups).

Persistence (an app-only `PlaidBarCache` store) and the editor UI are intentionally out of scope here — this PR is the model + the split-aware spend math + exports, matching how AND-546 landed the schema foundation.

## Testing

```
# → Build complete!

swift test --filter TransactionSplitTests
# → Suite "Transaction splits (AND-550)" passed — 19 tests

swift test
# → Test run with 2408 tests passed

git diff --check
# → clean
```

New suite `TransactionSplitTests` covers: the sum invariant (balanced/unbalanced/empty/sub-cent drift/negative refund), resolver expansion (no-split single row, valid-split per-allocation rows, malformed fallback), split-aware spend (rollups count parts, total conserved, per-part exclude honored, transfer/income parts dropped), restore-on-remove, and v1-safety (no-split spend is byte-identical; legacy JSON envelope decodes with empty splits).

## Links

Closes AND-550

## Follow-ups

- Persistence: an opt-in `PlaidBarCache` split store (mirrors `BudgetingV2Store`) + wiring splits into the app's spend calls.
- Editor UI: a split sheet in the Transaction Workspace inspector (`evenSplit` is provided as a seed/default).
- Review engine: surface a split badge / route split rows through the Review Inbox display (the inbox already groups by raw category; a split-aware inbox view is a separate surface).
- Charts: thread `splits` into the dashboard chart aggregations that call `SpendingSummary`.

### DIFF
--- a/Sources/PlaidBarCore/Models/TransactionSplit.swift
+++ b/Sources/PlaidBarCore/Models/TransactionSplit.swift
@@ -1,0 +1,193 @@
+import Foundation
+
+/// Splitting a single transaction into N category allocations (AND-550).
+///
+/// ## What this is (and the v1-safety contract)
+///
+/// A **child-allocation** model layered *additively* on top of ``TransactionDTO``.
+/// The parent transaction is never mutated, re-keyed, or removed: a split is a
+/// separate, opt-in record keyed by the parent's `transaction_id`. A user who
+/// does **not** split anything has no ``TransactionSplit`` records, so every
+/// consumer of transaction spend (budgets, summaries, review, exports, charts)
+/// sees byte-identical prior behavior. This is the deferred budgeting-v2 (AND-524)
+/// re-opening of scope that AND-399 explicitly cut.
+///
+/// A split divides the parent's signed amount (Plaid convention: positive = money
+/// out, negative = money in) into ``TransactionSplitAllocation`` parts. Each part
+/// carries its **own** budget category and its **own** budget-exclude flag, so a
+/// `$150` Target run can be `$90` Groceries + `$60` Home, or one part can be
+/// excluded from budgets while the rest counts. The defining **sum invariant** is
+/// that the parts' signed amounts sum back to the parent amount (within a cent),
+/// so splitting never creates or destroys money — it only re-attributes it.
+///
+/// ## Why a value type in `PlaidBarCore`
+///
+/// Pure, `Sendable`, no I/O, no hidden `Date()`. The split-aware spend math
+/// (``TransactionSplitResolver``) and every aggregation that reads it stay
+/// testable and shared across processes, per the project's "logic in Core"
+/// doctrine. Persistence (the app-only `PlaidBarCache` store) and the editor UI
+/// are separate, later surfaces — this file is the model + the sum invariant.
+
+// MARK: - Allocation (one category slice of a split)
+
+/// One slice of a split transaction: a signed amount attributed to a single
+/// budget category, with its own budget-exclusion flag.
+///
+/// Each allocation has a stable `id` so the editor can address an individual part
+/// (rename its category, toggle its exclude flag, change its amount) without
+/// reordering the rest. The `amount` is **signed**, in the parent's currency, and
+/// follows the same Plaid convention as ``TransactionDTO/amount`` (positive =
+/// money out). For a normal expense the parts are positive and sum to the
+/// positive parent amount; the model does not assume a sign so a refund (negative
+/// parent) splits the same way.
+public struct TransactionSplitAllocation: Codable, Sendable, Hashable, Identifiable {
+    /// Stable identity for this allocation row, independent of its position in the
+    /// split. A fresh UUID by default so a newly-added part never collides with an
+    /// existing one.
+    public let id: UUID
+    /// The budget category this slice is attributed to. The parent's own
+    /// ``TransactionDTO/category`` is ignored once a split exists — each part
+    /// declares its own bucket.
+    public let category: SpendingCategory
+    /// The signed amount of this slice, in the parent's currency (Plaid
+    /// convention: positive = money out). The parts' amounts must sum to the
+    /// parent amount (the split's sum invariant).
+    public let amount: Double
+    /// Whether this individual slice is excluded from budget/spend rollups. A
+    /// part can be excluded while its siblings still count — e.g. a reimbursable
+    /// line item carved out of a shared receipt. Defaults to `false` so an
+    /// allocation counts unless explicitly excluded.
+    public let excludedFromBudgets: Bool
+
+    public init(
+        id: UUID = UUID(),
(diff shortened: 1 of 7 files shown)


ftchvs:

**Verdict: CONCERNS** — one real cross-consumer correctness bug (transfer/income split allocations counted as spend in `SpendingSummary` but not in budgets), plus a latent conservation gap and a couple of minor notes. The model + resolver design is sound, the no-split path is genuinely byte-identical, exports are backward-compatible, and the build + 19-test suite pass locally. Not a blocker for the model itself, but the summary inconsistency should be fixed before this feeds dashboard charts.

What I verified (holds up):
- **No-split byte-identical**: `netSpendByCategory`, `overrideAwareSpend`, and `spendingByCategory` legacy/override paths reduce to the original `+= transaction.amount` / `+= displayAmount` with identical exclusion filters when `splits == []`. The resolver's `parentRow` carries parent id/amount/category verbatim. ✓
- **Sum invariant / restore-on-remove**: `isValid` gates application; malformed/empty/unbalanced → single parent row → editing or removing a split restores the original. ✓
- **Exports**: `schemaVersion` 2→3 additive, `decodeIfPresent ?? []`/`?? 0`, deterministic sort by `transaction_id`. v2 envelope decodes clean. ✓
- **Builds on AND-546 (#663), not a reinvention**: `BudgetingV2Schema` names "splits" as future scope but defines no split type — `TransactionSplit` is genuinely new. ✓
- Tests exercise the change (sum invariant, resolver expansion, budget rollups, exclude flag, restore, JSON round-trip). ✓

---

**MUST-FIX — `SpendingSummary` counts transfer/income split allocations as spend (cross-consumer inconsistency)**
`Sources/PlaidBarCore/Utilities/SpendingSummary.swift` — the split-allocation branches (both the legacy `guard metadata != nil…` path and the override path) filter only `isSplitExcluded`, then bucket by `row.category ?? .other`. They do **not** drop income/transfer categories. `CategoryBudgetPlanner.netSpendByCategory`'s split branch *does* (`if excludedCategories.contains(category) { continue }`, where `excludedCategories = [.income, .transfer, .transferOut]`).

`expenseTransactions(from:)` only filters at the **parent** level, so a `.shopping` parent passes and its `.transfer` allocation survives in the summary. Verified by probe — same split, two answers:
```
PROBE summary food=120.0 transfer=80.0     // SpendingSummary: transfer counted as spend
PROBE budget  food=120.0 transfer=absent   // CategoryBudgetPlanner: transfer dropped
```
This contradicts the PR body ("a split allocation … never counts income/transfer categories") and inflates summarized/charted spend vs. budgets for the identical split. The new suite's `splitIntoExcludedCategoryDropsOut` only covers `overrideAwareSpend`, so the gap is untested. Fix: apply the same `excludedCategories`-style drop to split allocations in `SpendingSummary`, and add a summary-path transfer/income test mirroring the planner's.

---

**SHOULD-FIX (latent, follow-up) — stale-split conservation gap; the documented detection is never wired up**
`Models/TransactionSplit.swift` claims persisting `expectedTotal` "lets the resolver detect a stale split if the underlying charge later changes amount." But `TransactionSplitResolver.spendRows(for:)` only checks `split.isValid` (allocations vs. the *stored* `expectedTotal`) — it never compares `expectedTotal` to the live `transaction.amount`. If Plaid re-posts a charge at a different amount (the same pending→posted scenario this code already handles for metadata via `pendingTransactionId`), the split silently applies stale allocations and total spend is no longer conserved against the live parent. Persistence is out of scope here, so this can't bite yet, but either wire up the `expectedTotal != transaction.amount` staleness check (fall back to parent) or drop the doc claim so the follow-up doesn't assume it exists.

---

**MINOR — `periodSummary` previous-total isn't split-aware**
`currentTotal` is computed from split-aware `categories`, but `previousTotal = previousExpenses.reduce { $0 + $1.displayAmount }` is raw. A previous-period split with an excluded allocation would be reflected in current but not previous totals, skewing the delta. Low severity (totals, not attribution) but inconsistent.

**MINOR — split allocations key off the v1 flat `SpendingCategory`**
AND-546 introduced `BudgetCategoryV2` (stable-id) explicitly to "replace the flat-enum `SpendingCategory` primary key." Fine for this PR (all spend consumers still use the v1 enum), but the persistence/editor follow-up should reconcile so splits don't get stranded on the old key.

**NIT** — `TransactionSplitResolver.index` "first split wins" on duplicate parent id is reasonable defensiveness, but a store-level uniqueness guarantee (the planned `PlaidBarCache` split store) should make duplicates impossible rather than silently dropped.

PR #679, opened 2026-06-25
feat(budgeting): rebalance — redistribute budget within a fixed total (AND-549)

ftchvs:

> Stacked on **AND-548** (per-month budgets). PR base is `feat/and-548-per-month-budgets`; this PR's own diff is the two new rebalance files. Review/merge AND-548 first.

## What & why

Copilot's **Rebalance**: move budget from an under-spent category to an over-spent one **while keeping that month's total budget constant**. Closes AND-549, the next dependent of the merged AND-546 schema (`MonthlyBudgetV2(month, categoryId, monthlyLimit, rollover)`) after AND-548's per-month rows. Purely additive `PlaidBarCore` math — no existing file is touched, so a v1 / not-opted-in user keeps **byte-identical** behavior.

## Change

New `BudgetRebalancePlanner` (pure, `Sendable`, deterministic — no hidden `Date()`):

- **`apply(_:to:asOf:)`** — moves `$X` from a source category to a destination for one month: source limit `−X`, destination limit `+X`, so the month total (`Σ monthlyLimit`) is **provably unchanged** (AC 1). Rejected as a no-op (schema byte-identical) when the month is frozen history (`RolloverBudgetPlanner.isMonthEditable`), the endpoints match, the amount is non-finite/non-positive, either endpoint is unbudgeted that month, or the amount exceeds the source limit (a source can never go negative).
- **`suggestRebalances(...)`** — proposes moves from categories with a **surplus** (limit above override-aware spend, pulling at most a cushion fraction of the headroom) toward categories **trending over** (spend above limit) (AC 2). Greedy and conservation-safe: applying every suggestion never changes the month total. Deterministic ranking (destination-overage-first, ties by category id); a frozen month yields none.
- **Undo** (AC 3) — every `BudgetRebalanceMove` has an exact `inverse`; the `ApplyResult` returns it as the undo token. Re-applying it restores the pre-move limits **byte-for-byte**, so a rebalance is fully undoable and the dashboard re-renders from the returned `BudgetingV2Schema` immediately.
- **`monthTotal(...)`** — exposed as the invariant helper callers/tests assert against.

Accessibility/privacy: pure value types, no UI in this PR; `RebalanceSuggestion` carries explicit `sourceSurplus` / `destinationOverage` context so a future UI can label source/destination roles without relying on color alone. Spend is supplied by the caller (typically `CategoryBudgetPlanner.overrideAwareSpend`), so Privacy Mask / App Lock gating stays at the surfaces that render it.

## Testing

Exact commands and results:

```
# → Build complete! (no warnings/errors)

swift test --filter BudgetRebalancePlannerTests
# → 17 tests passed

swift test --filter RolloverBudgetPlannerTests --filter MonthlyBudgetEditorTests
# → 31 tests passed (stacked AND-548 suites, no regression)

git diff --check
# → clean
```

The 17 new tests cover: the conservation invariant under chained moves, drain-to-zero, rejection of overdraw / frozen / missing-row / degenerate (self-move, zero, negative, non-finite) moves, lossless undo round-trip + double-inverse identity, surplus→overage suggestions, applying-all-suggestions conserves the total, no-overage / no-surplus / frozen yields none, and deterministic stable ranking. All deterministic (explicit `asOf` month key + caller-supplied spend; synthetic data only).

## Links

Closes AND-549.

## Follow-ups

- Wire `BudgetRebalancePlanner` into the Budgets workspace surface (apply/undo affordance + suggestion chips) — UI epic, separate PR.
- Persist applied moves through `PlaidBarCache.BudgetingV2Store` so a rebalance survives relaunch (the math here is store-agnostic by design).

### DIFF
--- a/Sources/PlaidBarCore/Utilities/BudgetRebalancePlanner.swift
+++ b/Sources/PlaidBarCore/Utilities/BudgetRebalancePlanner.swift
@@ -1,0 +1,397 @@
+import Foundation
+
+/// Pure, deterministic budget **rebalance** math for budgeting v2
+/// (AND-549 — deferred epic AND-524).
+///
+/// ## What this adds on top of AND-546 / AND-548
+///
+/// AND-546 persists ``MonthlyBudgetV2`` rows keyed by `(month, category)`; AND-548
+/// (``RolloverBudgetPlanner`` / ``MonthlyBudgetEditor``) adds the per-month carry
+/// math and immutability-aware editing. This planner is Copilot's **Rebalance**:
+/// move budget *from* an under-spent category *to* an over-spent one **within a
+/// single month, keeping that month's total budget constant**. It only ever
+/// reshuffles the existing `monthlyLimit` values for one month — it never changes
+/// the sum, never touches spend, and never edits a frozen historical month.
+///
+/// ## The invariant (AC 1)
+///
+/// A rebalance moves an amount `X` from a source category to a destination
+/// category for one `month`:
+///
+/// ```text
+/// source.monthlyLimit -= X
+/// dest.monthlyLimit   += X
+/// ```
+///
+/// so the month's total budget (`Σ monthlyLimit` over that month's rows) is
+/// **provably unchanged** — every applied move is `+X` and `−X` on the same total.
+/// ``apply(_:to:asOf:)`` enforces this exactly; ``BudgetRebalanceMove/amount`` is
+/// validated `> 0`, finite, and `<= source.monthlyLimit` so a source can never go
+/// negative and the conservation holds bit-for-bit.
+///
+/// ## Suggestions (AC 2)
+///
+/// ``suggestRebalances(...)`` proposes moves from categories with a **surplus**
+/// (limit comfortably above the month's override-aware spend) toward categories
+/// **trending over** (spend already above, or projected to exceed, the limit). It
+/// is greedy and total-preserving: it pulls only from real surplus and gives only
+/// what closes a real overage, so applying every suggestion leaves the month total
+/// unchanged. Suggestions are read-only proposals — nothing is mutated until the
+/// user applies a move.
+///
+/// ## Undo (AC 3)
+///
+/// Every applied ``BudgetRebalanceMove`` has an exact inverse
+/// (``BudgetRebalanceMove/inverse``: swap source/dest, same amount). Re-applying
+/// the inverse restores the pre-move limits **byte-for-byte**, so a rebalance is
+/// fully undoable and the dashboard reflects either state immediately (the result
+/// is just a new ``BudgetingV2Schema`` the read model re-renders from).
+///
+/// ## Determinism
+///
+/// Like ``RolloverBudgetPlanner`` and ``CategoryBudgetPlanner``, there is no hidden
+/// `Date()`: editability is gated by the explicit `asOf` month key, and spend is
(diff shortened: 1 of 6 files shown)


ftchvs:

**Verdict: LGTM — no must-fixes.** The three focus invariants hold under scrutiny, the change is purely additive (byte-identical for not-opted-in / v1 users), it builds correctly on the merged AND-546/AND-548 schema rather than reinventing it, and the new tests exercise the behavior. I reproduced the green state locally.

### Verification performed
- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete!** (clean, no warnings).
- `swift test --filter BudgetRebalancePlannerTests` → **17 / 17 passed**.
- Confirmed referenced symbols exist on the base branch and are used, not redefined: `MonthlyBudgetV2` / `BudgetingV2Schema` (`Sources/PlaidBarCore/Models/BudgetingV2Schema.swift`), `RolloverBudgetPlanner.isMonthEditable(_:asOf:)` (`month >= asOf`, malformed `asOf` → `false`), `BudgetingV2Migration.seed()`, `CategoryBudgetPlanner.overrideAwareSpend`.
- Diff touches only the two new files (`BudgetRebalancePlanner.swift`, tests) — no existing source modified, so the not-opted-in path is byte-identical. ✅

### Focus invariants — checked
- **Total-preserving (AC 1):** `apply` writes source `−amount` / dest `+amount` with the *same* `Double`, so `Σ monthlyLimit` cancels exactly (FP-safe). ✅
- **No negative budgets:** guarded by `amount <= sourceRow.monthlyLimit`; suggestion amounts are additionally capped at `monthlyLimit`. I traced the chained-apply case where one source funds multiple destinations: cumulative pulls ≤ `pullable` ≤ `monthlyLimit`, and each successive move ≤ current (decremented) limit, so no applied suggestion is ever rejected and "apply-all conserves total" holds. ✅
- **Undo (AC 3):** `inverse` swaps source/dest at equal amount; `apply`→`apply(inverse)` restores the schema (`==` asserted in `undoRoundTrips`); double-inverse identity covered. ✅

### Advisory (non-blocking) findings
1. **Doc vs. code mismatch (`BudgetRebalancePlanner.swift`):** docstring says destinations are categories "spend already above, **or projected to exceed**, the limit," but the code only flags `remaining < 0` (already over) — there is no projection. Tighten the prose or drop "projected to exceed" to avoid implying a forecast that isn't there.
2. **Duplicate-row collapse (`apply` / `row`):** `row(...)` is `.last`-wins, and `apply` filters out *all* rows matching the source/dest `(month, category)` then re-appends one each. If a malformed schema ever held duplicate rows for one `(month, category)`, an *applied* move would silently collapse them and shift `monthTotal`. The `MonthlyBudgetV2.id` contract says rows are unique so this is out-of-contract input, but a one-line note (or an early uniqueness assumption) would document the dependency. Not exercised by any test.
3. **Negative-spend (refund) draws (`suggestRebalances`):** when `spend < 0`, `remaining > monthlyLimit` and the `min(remaining * frac, monthlyLimit)` cap lets a refunded source be pulled up to its *entire* limit (not half), i.e. the "leave-a-cushion" intent is bypassed. Behavior is still conservation-safe; flagging only as a product nuance for the future UI.

None of the above block merge. The stacked-base caveat in the PR body is correct — land AND-548 first.

PR #689, opened 2026-06-26
fix(budgeting): rebalance can target zero-limit destination categories (AND-672)

ftchvs:
## Summary

The budget **rebalance** suggester (`suggestRebalances` in `BudgetRebalancePlanner`, AND-549) silently dropped destination categories whose current `monthlyLimit` is `0`. A brand-new or just-zeroed category that is overspent could therefore never **receive** redistributed budget from a rebalance — surprising for a feature framed as "redistribute within a fixed total."

The candidate pre-filter (`schema.budgets.filter { $0.month == month && $0.monthlyLimit > 0 }`) excluded every zero-limit row **before** the source/destination split. That exclusion is correct for the **source** side (you can't pull budget from a category that has none) but wrong for the **destination** side.

This change applies the zero-limit exclusion to **sources only**, making a zero-limit overspent category an eligible destination again. The month-total invariant is preserved.

## Architectural rationale

Pure, deterministic logic stays in `PlaidBarCore` (`Sources/PlaidBarCore/Utilities/BudgetRebalancePlanner.swift`), per the project's "shared logic in Core" convention — no view or route is touched. The fix is a one-line filter relaxation plus comments; the source/destination ranking, greedy walk, and conservation guarantees are unchanged.

The source side stays safe **by construction**, not by an explicit re-filter: the source branch computes `pullable = min(remaining * maxSurplusFraction, monthlyLimit)`, which is `0` when `monthlyLimit == 0`, and is gated behind `if pullable > 0`. So a zero-limit row can never become a source. Only the destination branch (`remaining < 0`) newly sees zero-limit overspent rows.

## Design decisions

- **Smallest behavior-correct change.** Relaxed the pre-filter from `month && limit > 0` to `month`, rather than restructuring the source/destination loop. The existing `if pullable > 0` guard already enforces the source-side zero-limit exclusion, so no redundant guard was added.
- **Invariant untouched.** `apply(_:to:asOf:)` already supports a zero-limit destination (`destination.monthlyLimit + amount`) and rejects any move that would push a source negative. Every applied move remains `+X`/`-X` on the same month total, so `monthTotal` is provably unchanged.
- **No new public surface.** Signature, defaults, and determinism (ranking by amount desc, then category id asc) are all preserved; the 17 existing AND-549 tests still pass unchanged.

## Testing notes

Exact commands and results (run with the sandbox disabled per repo quirk):

```
# → Build complete! (CI strict-concurrency gate passes)

swift test --filter "zeroLimitDestinationReceivesShare"
# → 1 test passed (after fix)

swift test --filter "BudgetRebalancePlannerTests"
# → Test run with 18 tests passed (17 existing + 1 new)
```

**Fails-before / passes-after confirmation:** with the source fix stashed (pre-fix code) but the new test present, `swift test --filter "zeroLimitDestinationReceivesShare"` **failed** with 4 issues — the zero-limit `travel` destination was dropped and `suggestions` came back empty. Restoring the fix makes it pass. This proves the AND-549 suite was previously blind to the drop (its suggestion tests pin neither destination count nor redistributed amount for a zero-limit target).

New test: `zeroLimitDestinationReceivesShare` — rebalances into a set that includes a zero-limit overspent category and asserts (a) it receives a positive share and (b) the grand total is invariant after applying every suggestion.

## Linked issue

- AND-672 — https://linear.app/andeslab/issue/AND-672

## Known limitations

- The new test exercises a single zero-limit overspent destination plus one surplus source. It does not fan out across many zero-limit destinations competing for a constrained surplus; the existing greedy ordering already covers multi-destination ranking, so this was kept minimal and focused on the dropped-destination regression.
- This only affects **suggestions**. `apply` already handled zero-limit destinations correctly, so no apply-path behavior changed.

## Follow-ups

- The broader AND-549 suggestion suite could pin redistributed amount and destination count more tightly (the weakness that hid this bug) — a small hardening pass, out of scope here.

### DIFF
--- a/Sources/PlaidBarCore/Utilities/BudgetRebalancePlanner.swift
+++ b/Sources/PlaidBarCore/Utilities/BudgetRebalancePlanner.swift
@@ -283,12 +283,19 @@ public enum BudgetRebalancePlanner {
             maxSurplusFraction > 0
         else { return [] }
 
-        let monthRows = schema.budgets.filter { $0.month == month && $0.monthlyLimit > 0 }
+        let monthRows = schema.budgets.filter { $0.month == month }
         guard !monthRows.isEmpty else { return [] }
 
         // Sources: categories under budget, ranked by how much we may pull (a cushion
         // share of the surplus). Destinations: categories over budget, ranked by how
         // far over. Both deterministic (amount desc, then category id asc).
+        //
+        // The zero-limit exclusion applies to *sources only* — you can't pull budget
+        // from a category that has none, and `pullable` below is naturally `0` when
+        // `monthlyLimit == 0`. A zero-limit category is still a legitimate
+        // *destination*: a brand-new or just-zeroed category that is overspent
+        // (`spend > 0`, so `remaining < 0`) must be eligible to receive a share, or
+        // "redistribute within a fixed total" silently drops it (AND-672).
         var sources: [(categoryId: String, available: Double)] = []
         var destinations: [(categoryId: String, overage: Double)] = []
 
@@ -297,6 +304,7 @@ public enum BudgetRebalancePlanner {
             let remaining = row.monthlyLimit - spend
             if remaining > 0 {
                 // Pull at most a fraction of the surplus, but never more than the limit.
+                // `0` when the limit is `0`, so a zero-limit row is never a source.
                 let pullable = min(remaining * maxSurplusFraction, row.monthlyLimit)
                 if pullable > 0 {
                     sources.append((row.categoryId, pullable))
--- a/Tests/PlaidBarCoreTests/BudgetRebalancePlannerTests.swift
+++ b/Tests/PlaidBarCoreTests/BudgetRebalancePlannerTests.swift
@@ -300,6 +300,44 @@ struct BudgetRebalancePlannerTests {
         #expect(BudgetRebalancePlanner.monthTotal(in: current, month: "2026-06") == totalBefore)
     }
 
+    @Test("A zero-limit overspent category is an eligible rebalance destination (AND-672)")
+    func zeroLimitDestinationReceivesShare() {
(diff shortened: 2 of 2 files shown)


ftchvs:
**Independent adversarial review: ✅ APPROVE (high confidence).**

Correctness, edge cases, and the conservation invariant all hold. `monthRows` is local to `suggestRebalances` (only the empty-guard and the source/destination loop consume it); the source side stays excluded by construction (`pullable = min(remaining·frac, monthlyLimit) = 0` at limit 0, gated by `if pullable > 0`), and a zero-limit/zero-spend row falls into neither branch. `monthTotal` filters `schema.budgets` independently and `apply` already supports a zero-limit destination, so the total stays invariant. The new test genuinely **fails on revert** (restoring `&& $0.monthlyLimit > 0` empties `destinations`, the guard returns `[]`, and `#expect(!suggestions.isEmpty)` trips). `suggestRebalances` has no live UI consumer yet (deferred AND-524), so no other call site regresses. Pure Sendable PlaidBarCore logic, Swift Testing — no privacy/concurrency concern.

Non-blocking nit: the test pins that the zero-limit destination receives a share + total invariance, but not the exact moved amount or destination count — the same precision gap that originally hid this bug. Worth a small follow-up to harden the broader AND-549 suggestion assertions (tracked, out of scope here).

Ready for your review.

ftchvs (commented):

The rebalance change is narrowly scoped to `BudgetRebalancePlanner`: zero-limit rows can now participate only as overspent destinations, while the source side still requires positive `pullable` budget and cannot drain a zero-limit category. Applying suggestions remains total-preserving, deterministic, and contained in PlaidBarCore. No Plaid credentials, provider tokens, raw Plaid payloads, local database data, screenshots, or app-target Plaid calls are introduced.

Validation run:
- `swift test --filter zeroLimitDestinationReceivesShare`

PR #690, opened 2026-06-26
fix(server): transient Keychain failure no longer marks item as errored (AND-669)

ftchvs:
## Summary

A transient Keychain / token-vault failure while resolving a Plaid item's access token was being misclassified as a **hard item error**, surfacing a broken connection to the user and potentially triggering needless reconnect/recovery UX.

`PlaidTokenVaultError` is thrown by `PlaidTokenVault.resolve(...)` (via `TokenStore.accessToken(for:)`) when the Keychain is momentarily unreachable — device locked, Keychain temporarily unavailable, an ACL prompt the background server cannot satisfy, etc. These are recoverable hiccups, not genuine item/connection faults. Both `AccountRoutes` and `TransactionRoutes` caught such throws in their per-item catch and ran them through `ItemStatusMapping.status(forAPIError:)`, which — because a `PlaidTokenVaultError` is not a `PlaidError.apiError` — fell through to the hard `.error` state.

This change special-cases `PlaidTokenVaultError` in that single shared helper so it maps to the existing **`.providerOutage`** state: degraded but non-actionable, auto-retried, and never a user-facing "reconnect" prompt.

## Architectural rationale

Both read routes already funnel every per-item failure through one pure helper — `ItemStatusMapping.status(forAPIError:)` (`Sources/PlaidBarServer/Routes/ItemStatusMapping.swift`). `AccountRoutes.refreshAccounts` and `TransactionRoutes.syncItemRecordingStatus` each call `tokenStore.updateItemStatus(..., status: ItemStatusMapping.status(forAPIError: error).rawValue)` in their catch. Fixing the classification at that one seam corrects **both** route files at once with no behavioral drift between them, and keeps the change in a pure, `Sendable`, trivially unit-testable function (per the repo's "put logic in helpers, not routes" doctrine).

## Design decisions

- **Reused `.providerOutage` rather than adding a new status.** That state already encodes exactly the intended semantics — transient, non-actionable, auto-retried, `needsUpdateMode == false`, no Update flow (see `ItemConnectionStatus` in `LinkResponse.swift`, and the existing `INSTITUTION_DOWN`/`PLANNED_MAINTENANCE` mappings). No new enum case, no migration, no new UI wiring.
- **Type-level match (`error is PlaidTokenVaultError`)** rather than enumerating individual Keychain status codes — every `PlaidTokenVaultError` variant (unavailable, load/save/delete failure, invalid stored token) is recoverable-or-retryable from the item's perspective; none of them is a genuine Plaid connection error.
- **Minimal and symmetric:** one guard added before the existing `PlaidError.apiError` switch. Genuine Plaid item errors (and any other unrelated error) still map to the hard `.error` state, so the special-case cannot swallow real connection failures.
- **Resolved at the classifier, not by reordering the catch.** The suggested alternative (resolve the token before the per-item catch) would have required duplicating the change across both route files and risked changing the all-items-failed/`shouldFailRefresh` accounting; the classifier seam is the smaller, lower-risk surface.

## Testing notes

New test: `itemStatusMappingTreatsTokenVaultFailuresAsTransient` in `Tests/PlaidBarServerTests/PlaidBarServerTests.swift`. It feeds every `PlaidTokenVaultError` variant through `ItemStatusMapping.status(forAPIError:)` and asserts the result is `.providerOutage` (and explicitly **not** `.error`, and `needsUpdateMode == false`), while a genuine `PlaidError.apiError` and an unrelated error still map to `.error`.

Red-before-green verified: with the source fix stashed, the test fails with `(status → .error) == .providerOutage` for each variant; with the fix applied it passes.

```
Build complete! (273.81s)

$ swift test --filter itemStatusMappingTreatsTokenVaultFailuresAsTransient
✔ Test "Token-vault/Keychain failures map to transient .providerOutage, not a hard .error" passed after 0.001 seconds.
✔ Test run with 1 test passed.

# neighborhood regression check (status-mapping + all-failure refresh tests)
$ swift test --filter "ItemStatusMapping|itemStatusMapping|webhookItemStatusMapping|AllFailure|RefreshAll"
✔ Test run with 5 tests passed after 0.162 seconds.
```

## Linked issue

- AND-669 (Medium): Transient Keychain/token-vault failure misclassified as a hard error state.
- https://linear.app/andeslab/issue/AND-669

## Known limitations

- The fix is at the status-classification layer. A persistent (non-transient) Keychain failure — e.g. the user permanently denies the ACL, or the stored token is genuinely corrupt (`invalidStoredToken`) — will now also be reported as `.providerOutage` and retried rather than surfaced as a hard error. That is the intended trade-off for AND-669 (favor auto-retry over a false "broken connection"), but it means a truly unrecoverable Keychain state self-heals only if/when the token becomes readable; it will not, on its own, prompt the user to re-link.
- No change to the all-items-failed accounting: a request where *every* item fails on a token-vault error still returns the existing `502 badGateway` ("refresh failed for every linked item"). Only the persisted per-item connection **status** changes from `.error` to `.providerOutage`.

## Follow-ups

- Consider distinguishing genuinely terminal token-vault states (e.g. `invalidStoredToken`) from transient ones if telemetry shows items getting stuck in `.providerOutage` — possibly a bounded retry budget before escalating to `.error`.
- Optional: a server-side log line when a token-vault failure is downgraded to `.providerOutage`, to aid diagnosing Keychain-availability issues in the field (item id only, never token material — consistent with the existing `deleteItem` logging).

### DIFF
--- a/Sources/PlaidBarServer/Routes/ItemStatusMapping.swift
+++ b/Sources/PlaidBarServer/Routes/ItemStatusMapping.swift
@@ -3,6 +3,17 @@ import PlaidBarCore
 
 enum ItemStatusMapping {
     static func status(forAPIError error: Error) -> ItemConnectionStatus {
+        // A token-vault / Keychain failure (device locked, Keychain temporarily
+        // unavailable, an ACL prompt the background server can't satisfy) is
+        // almost always TRANSIENT, not a genuine item/connection fault. Mapping
+        // it to a hard `.error` surfaces a broken connection and can trigger
+        // needless reconnect/recovery UX for a hiccup that resolves itself once
+        // the Keychain is reachable again. Treat it like a provider-side outage
+        // (`.providerOutage`): degraded but non-actionable and auto-retried,
+        // never a user-facing "reconnect" prompt (AND-669).
+        if error is PlaidTokenVaultError {
+            return .providerOutage
+        }
         guard case PlaidError.apiError(_, _, let errorCode, _) = error else {
             return .error
         }
--- a/Sources/PlaidBarServer/Routes/ItemStatusMapping.swift
+++ b/Sources/PlaidBarServer/Routes/ItemStatusMapping.swift
@@ -3,16 +3,19 @@ import PlaidBarCore
 
 enum ItemStatusMapping {
     static func status(forAPIError error: Error) -> ItemConnectionStatus {
-        // A token-vault / Keychain failure (device locked, Keychain temporarily
-        // unavailable, an ACL prompt the background server can't satisfy) is
-        // almost always TRANSIENT, not a genuine item/connection fault. Mapping
-        // it to a hard `.error` surfaces a broken connection and can trigger
-        // needless reconnect/recovery UX for a hiccup that resolves itself once
-        // the Keychain is reachable again. Treat it like a provider-side outage
-        // (`.providerOutage`): degraded but non-actionable and auto-retried,
-        // never a user-facing "reconnect" prompt (AND-669).
-        if error is PlaidTokenVaultError {
-            return .providerOutage
+        // A token-vault / Keychain availability failure (device locked, Keychain
+        // temporarily unavailable, an ACL prompt the background server can't
+        // satisfy) is transient, not a genuine Plaid item fault. Treat those like
+        // provider outages: degraded, non-actionable, and auto-retried. A corrupt
+        // stored token is different: re-reading the same invalid bytes will not
+        // self-heal, so keep it in the hard-error recovery lane.
+        if let tokenVaultError = error as? PlaidTokenVaultError {
+            switch tokenVaultError {
+            case .keychainUnavailable, .keychainSaveFailed, .keychainLoadFailed, .keychainDeleteFailed:
+                return .providerOutage
+            case .invalidStoredToken:
+                return .error
+            }
         }
         guard case PlaidError.apiError(_, _, let errorCode, _) = error else {
             return .error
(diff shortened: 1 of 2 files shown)


ftchvs:
**Independent adversarial review: ✅ APPROVE with nits (high confidence).**

Correct, minimal, and well-scoped: a single type-match guard at the one shared classifier seam (`ItemStatusMapping.status(forAPIError:)`) that both `AccountRoutes` and `TransactionRoutes` funnel per-item errors through. `.providerOutage` already carries the right semantics (`needsUpdateMode == false`, excluded from WeeklyReview reconnect logic); the `credentialsNotConfigured` path precedes it and is unaffected; the all-items-failed `502` accounting is orthogonal and untouched. Genuine `PlaidError.apiError` and unrelated errors still map to `.error`. The new test **fails on revert** (a `PlaidTokenVaultError` would fall through to `.error`) and includes negative cases.

Two non-blocking nits, both already in the PR's Known Limitations:
1. **`.invalidStoredToken` is genuinely terminal**, not transient — a corrupt/empty stored token never self-heals, so re-reads keep throwing; the blanket `is PlaidTokenVaultError` match downgrades it too, leaving such an item in a perpetual "we'll retry automatically" state with no reconnect path.
2. **No server log line** when a failure is downgraded, making a stuck item hard to diagnose.

Both are acceptable AND-669 trade-offs (favor auto-retry over a false "broken connection"). I'll file a follow-up to special-case the terminal variant (bounded retry budget → escalate to `.error`) and add an item-id-only downgrade log.

Ready for your review.

ftchvs (commented):

The original blanket `error is PlaidTokenVaultError` mapping also downgraded `.invalidStoredToken` to `.providerOutage`. That variant is produced after Keychain returns data but the token bytes are empty/corrupt, so retrying the same item will keep failing and the non-actionable provider-outage lane would hide a terminal local-token recovery state from the user. I pushed `c3d52eb3` to keep Keychain availability/load/save/delete failures transient while mapping `.invalidStoredToken` back to `.error`.

Validation run after the fix:
- `git diff --check`
- `swift test --filter itemStatusMappingTreatsTokenVaultFailuresAsTransient`
- `swift build --target PlaidBarServerTests -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors`

No merge performed.

PR #691, opened 2026-06-26
fix(alerts): category-budget alert re-fires on band re-entry (AND-663)

ftchvs:
## Summary

Category-budget alerts (AND-642) never re-fired once spend left an alert band and re-entered it within the same month. The notification kind set `clearsWhenResolved = false`, baking in the assumption that spend climbs monotonically within a month (`under → nearing → over`). Refunds and recategorization break that assumption: they can drop a category's month-to-date spend back below a band it had already crossed. Because the delivered dedup key (`category + month + band`) was never cleared, re-entry into the same band was de-duplicated for the rest of the month, so the user was **not** re-alerted on the second crossing.

This flips `categoryBudgetAlert` to `clearsWhenResolved = true`, making it a stateful band condition that re-arms when its band goes inactive.

## Architectural rationale

The fix reuses the existing resolution mechanism rather than inventing a new one. `NotificationTriggerSelection.evaluate` already computes `resolvedDedupKeys` for any delivered key whose kind `clearsWhenResolved` and whose family was evaluated this pass, **minus** the keys still active this pass. `NotificationService` removes resolved keys from its persisted delivered set, re-arming them. This is exactly how `highUtilization`, `lowBalance`, `syncStale`, etc. already behave: stateful conditions that clear when no longer active.

Conceptually, a category-budget alert *is* a stateful band condition (like credit utilization above a threshold), not a one-shot event (like a single large transaction). It was miscategorized in the original AND-642 implementation. The one-shot kinds (`largeTransaction`, `recurringChargeDetected`, `merchantWatch`, `categoryWatch`) stay `false` — the spend they describe already happened and should not auto-clear. The `CategoryBudgetAlertEvaluator` stays stateless and unchanged; only its classification in `clearsWhenResolved` moved.

## Design decisions

- **Avoiding re-alert spam while in-band:** `resolvedDedupKeys` is `clearableDeliveredKeys.subtracting(activeDedupKeys)`. While spend sits inside a band, that band's key is re-emitted into `activeDedupKeys` every pass, so it is subtracted out and never resolved — and it stays in the delivered set, so the decision is filtered out. The alert resolves and re-arms **only** when the band is genuinely no longer active. No change to in-band behavior; the existing `dedupEscalation` test (same band on repeated refresh → no new decision, `resolvedDedupKeys` empty) still passes unchanged.
- **Per-band grain preserved:** the dedup key is still `category + month + band`, so `nearing` and `over` are independent. Escalation `nearing → over` still fires a distinct alert; de-escalation `over → nearing` resolves the `over` key so a later re-cross of `over` fires again. This is the desired "re-entry fires fresh" behavior, applied per band.
- **Minimal change:** a single `case` move in the `clearsWhenResolved` switch plus doc-comment corrections (the old comments asserted the now-falsified monotonic premise). No signature, storage, or evaluator changes.

## Testing notes

Exact commands and results (run with the sandbox disabled per repo quirk):

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete!** (CI strict-concurrency gate, 0 warnings/errors).
- `swift test --filter CategoryBudgetAlertEvaluatorTests` → **13 tests passed** (includes the new regression test).
- `swift test --filter NotificationTriggerEvaluationTests` → **9 tests passed** (no regression from the `clearsWhenResolved` change).

New test: `refireAfterRefundDropsOutOfBand` in `Tests/PlaidBarCoreTests/CategoryBudgetAlertEvaluatorTests.swift`. It simulates: spend crosses into `nearing` (alert fires) → refund drops spend to `under` (delivered key **resolves**; the caller drops it) → spend rises back into `nearing` (alert fires **again**, same key). Verified it **fails before** the fix (all three assertions fail: the key is never resolved and the re-cross is suppressed) and **passes after** — confirmed by stashing only the core change and re-running.

## Linked issue

- AND-663 (Low) — https://linear.app/andeslab/issue/AND-663

## Known limitations

- Resolution is recompute-driven, not event-driven: the alert re-arms on the **next** trigger evaluation after spend leaves the band, not the instant a refund posts. Matches every other stateful trigger in this file.
- The dedup key carries no spend value, so a tiny dip below the band boundary and back (e.g. 100.5% → 99.5% → 100.5% across two evaluation passes) counts as a genuine re-crossing and re-fires. This is intentional and consistent with the per-band model; if it proves noisy in practice, a hysteresis margin on the band boundary would be the follow-up.
- The 1,000-key cap and per-month re-arm in `NotificationService` are unchanged.

## Follow-ups

- Consider band-boundary hysteresis if real-world refund noise causes flapping near a boundary (out of scope here; no evidence yet).
- The broader `clearsWhenResolved` semantics could be documented once on the enum rather than per-case; not done here to keep the diff minimal.

### DIFF
--- a/Sources/PlaidBarCore/Utilities/CategoryBudgetAlertEvaluator.swift
+++ b/Sources/PlaidBarCore/Utilities/CategoryBudgetAlertEvaluator.swift
@@ -16,9 +16,11 @@ import Foundation
 /// The de-dup contract (see ``NotificationTriggerSelection``) keys an alert on
 /// *category + month + band*, so a category fires **once** when it enters
 /// `nearing` and **once more** if it later escalates to `over` — but not on every
-/// refresh while it sits in the same band. Crossing `under → nearing → over` is
-/// monotonic within a month, so this models the "near or exceed" requirement
-/// without storing prior spend.
+/// refresh while it sits in the same band. This evaluator stays stateless; the
+/// notification layer treats the alert as a stateful band condition that clears
+/// when the band goes inactive (``NotificationTriggerKind/clearsWhenResolved``),
+/// so a refund or recategorization that drops spend out of a band re-arms it and
+/// a later re-crossing fires again within the same month (AND-663).
 ///
 /// ## Privacy
 /// This type emits no amounts — only the category and its band. The notification
--- a/Sources/PlaidBarCore/Utilities/NotificationTriggerSelection.swift
+++ b/Sources/PlaidBarCore/Utilities/NotificationTriggerSelection.swift
@@ -79,17 +79,22 @@ public enum NotificationTriggerKind: String, Codable, CaseIterable, Sendable {
 
     public var clearsWhenResolved: Bool {
         switch self {
+        // A category-budget alert is a *stateful* band condition, like
+        // highUtilization/lowBalance: it auto-resolves when its band is no longer
+        // active. Spend is not monotonic within a month — a refund or
+        // recategorization can drop a category's month-to-date spend back below a
+        // band it had crossed. Clearing the delivered key when the band goes
+        // inactive re-arms it, so re-entering the same band later that month fires
+        // a fresh alert instead of being suppressed for the rest of the month
+        // (AND-663). While spend sits inside a band the key stays active, so it is
+        // never resolved and never re-fires — no in-band spam.
         case .itemError, .providerOutage, .loginRequired, .syncStale, .highUtilization, .lowBalance,
-             .recurringChargeChanged, .recurringChargeDueSoon:
+             .recurringChargeChanged, .recurringChargeDueSoon, .categoryBudgetAlert:
             true
         // A crossed watchlist threshold is a one-shot like largeTransaction:
         // the spend already happened, so it should not auto-clear when the
-        // month-to-date sum later changes. A category-budget alert is keyed on
-        // its band, and within a month spend only climbs (under → nearing →
-        // over), so a delivered band-crossing should likewise not auto-resolve —
-        // each higher band carries its own one-shot key.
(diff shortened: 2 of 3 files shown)


ftchvs:
**Independent adversarial review: ✅ APPROVE (high confidence).**

Correctly reclassifies `categoryBudgetAlert` from a one-shot to a stateful band condition (`clearsWhenResolved = true`), reusing the existing `resolvedDedupKeys = clearableDelivered.subtracting(activeDedupKeys)` machinery exactly like `highUtilization`/`lowBalance`. Because the evaluator emits only the current band, a delivered key resolves **only** when its band genuinely goes inactive (refund/recategorization de-escalation), so it re-arms and re-fires on a true re-crossing while never re-firing in-band (the active band key is re-emitted and subtracted out — no spam). End-to-end re-arm confirmed against `NotificationService.swift:146-149`. The new test `refireAfterRefundDropsOutOfBand` **fails on revert** (step 2's `resolvedDedupKeys.contains(nearingKey)` assertion trips) — not a tautology. No privacy regression (bodies carry no amounts or category names); Sendable enum unchanged.

Non-blocking nit: boundary oscillation across evaluation passes can re-fire — inherent to the per-band stateful model (mirrors how `highUtilization`/`lowBalance` flap around thresholds); hysteresis is the documented follow-up if it proves noisy in practice.

Ready for your review.

ftchvs (commented):

Reclassifying `categoryBudgetAlert` as a clearable, stateful band condition matches the existing delivered-key resolution model: the key stays active while the category remains in-band, clears when the band goes inactive, and can re-fire on a real re-entry. The notification copy remains generic and does not expose category names, merchant names, account identifiers, or amounts on lock-screen-delivered surfaces. No Plaid/token/server boundary changes were introduced.

Validation run:
- `swift test --filter refireAfterRefundDropsOutOfBand`

PR #692, opened 2026-06-26
fix(accounts): account search no longer clears persisted selection (AND-662)

ftchvs:
## Summary

The window Accounts destination's self-heal `onChange` reconciled the dashboard account selection against the **search-filtered** account list instead of the full list. `NavigationModel.reconcileSelection` clears `selectedAccountID` when the selection is not in the visible set, and that value is persisted to UserDefaults. So typing a search query that did not match the currently-selected account **cleared the persisted selection**, which was then lost across relaunch — even after the search was cleared.

This is the AND-625 regression introduced by `d930d2c` (PR #658). The reconcile is only meant to fire on genuine account-set changes (account removed / disconnected), per the code comment, not on a transient text filter.

The fix observes the **full** `accounts` list, so reconcile fires only when accounts actually change. The "which list drives reconcile" decision is centralized in a small pure helper, `NavigationState.reconcileVisibleAccountIDs(allAccountIDs:filteredAccountIDs:)`, so it is unit-testable.

## Architectural rationale

- `AccountsDestinationView` is part of the `PlaidBar` **`@main` executable target**, which cannot be `@testable import`ed (see the note in `Tests/PlaidBarTests/PlaidBarTests.swift`). A one-line view change alone could not be guarded by a test that fails before and passes after.
- Per CLAUDE.md ("put shared logic in `PlaidBarCore`") and the issue's own testing guidance ("the cleanest level to test is `reconcileSelection` behavior **+ the view's choice of which list drives it**"), the list-selection decision was lifted into a pure, `Sendable` static helper on `NavigationState` in `PlaidBarCore`. The view now delegates to it. This turns an untestable SwiftUI binding into a tested pure function.
- `reconcileSelection` itself was already correct and well-tested; it is unchanged.

## Design decisions

- **Minimal behavior change:** the only runtime change is the reconcile input going from `filteredAccounts` → full `accounts`. The popover (`MainPopover`) intentionally keeps reconciling against its filter-driven `filteredAccounts` (it has a separate explicit filter→deselect rule and no free-text search), so it is deliberately untouched.
- **Helper signature keeps `filteredAccountIDs`** (currently unused, `_`-bound) so the call site self-documents that the filtered list is the *wrong* input and the choice is centralized in one place. If a future requirement wants the inspector to react to filtering, the decision can evolve in this one pure function without touching the view.

## Testing notes

All commands run with the SwiftPM sandbox disabled (known repo quirk).

- **CI gate (strict concurrency), passes:**
  `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → `Build complete!`
- **New regression test, passes after fix:**
  `swift test --filter "searchFilterDoesNotClearSelection"` → 1 test passed
- **Fail-before/pass-after verified:** temporarily flipping `reconcileVisibleAccountIDs` to return `filteredAccountIDs` (the regression) makes the test **fail with 3 issues**; restoring it to return `allAccountIDs` makes it **pass**.
- **Suite, passes:** `swift test --filter "NavigationStateTransitionTests"` → 18 tests passed (includes the new test).

Test added: `searchFilterDoesNotClearSelection` — "Account search filter must not clear the persisted selection (AND-662)" in `Tests/PlaidBarCoreTests/RouteNavigationTests.swift` (`@Suite("NavigationState transitions")`). It models the view's case-insensitive name-substring filter, asserts the helper returns the full list, and asserts the end-to-end behavior (full-list reconcile keeps the selection; filtered-list reconcile clears it).

## Linked issue

- AND-662 (Medium) — https://linear.app/andeslab/issue/AND-662
- Regression of AND-625; bug introduced by `d930d2c` (PR #658).

## Known limitations

- The regression is guarded at the `PlaidBarCore` logic level (the `@main` app target is not `@testable`-importable). The test asserts the exact pure decision the view now delegates to plus the resulting reconcile behavior, but it does not exercise the live SwiftUI `onChange` binding. A reviewer reverting *only* the view's `.onChange(of:)` source back to `filteredAccounts` while leaving the helper correct would not be caught by a unit test — the view now routes through the helper specifically to make that revert visible in code review.

## Follow-ups

- None required. Optional: if product later wants the Accounts inspector to collapse when the selected account is filtered out by search (without persisting the clear), decouple the *filter-driven UI collapse* from the *persisted selection* — the centralized helper is the seam for that.

### DIFF
--- a/Sources/PlaidBar/Views/Destinations/AccountsDestinationView.swift
+++ b/Sources/PlaidBar/Views/Destinations/AccountsDestinationView.swift
@@ -128,20 +128,13 @@ struct AccountsDestinationView: View {
             // disconnected), clear it so the inspector returns to its prompt instead
             // of pointing at a vanished account. Mirrors the popover's reconcile.
             //
-            // Observe the FULL account list, not `filteredAccounts` (AND-662). The
-            // cleared selection is persisted to UserDefaults, so reconciling against
-            // the search-filtered list let a transient text filter that excluded the
-            // selected account wipe the persisted selection — lost across relaunch
-            // even after the search was cleared (AND-625 regression). Account-set
-            // changes are the only legitimate trigger; a text filter is not one.
-            // `NavigationState.reconcileVisibleAccountIDs` owns (and tests) that
-            // choice, since this `@main` view cannot be unit-tested directly.
+            // Observe the FULL account list, not `filteredAccounts` (AND-662): the cleared
+            // selection is persisted, so reconciling against the search-filtered list let a
+            // transient text filter that excluded the selected account wipe the persisted
+            // selection — lost across relaunch even after the search cleared (AND-625
+            // regression). Account-set changes are the only legitimate trigger.
             .onChange(of: accounts.map(\.id)) { _, allIDs in
-                let visibleIDs = NavigationState.reconcileVisibleAccountIDs(
-                    allAccountIDs: allIDs,
-                    filteredAccountIDs: filteredAccounts.map(\.id)
-                )
-                navigationModel.reconcileSelection(visibleAccountIDs: visibleIDs)
+                navigationModel.reconcileSelection(visibleAccountIDs: allIDs)
             }
     }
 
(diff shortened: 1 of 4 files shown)


ftchvs (commented):

The fix keeps the persisted account selection reconciled against the full account set rather than the transient search-filtered list, which matches the intended self-heal boundary: account removal/disconnect clears stale selection; a search query does not persistently clear it. The diff stays in the SwiftUI destination/source-invariant test surface and does not alter PlaidBarServer, token custody, raw Plaid payload handling, or local data storage.

Validation run:
- `swift test --filter accountsAdoptsUnifiedShellSearch`

Note: the first attempted filter used the display string and matched zero tests; I reran with the exact function symbol above and it executed/passed. The normal app-target build still emits an existing Swift 6 actor-isolation warning in `WindowActivationPolicy`, not introduced by this PR.

PR #694, opened 2026-06-26
fix(a11y): activity heatmap cells carry per-cell label/help (AND-671)

ftchvs:
## Summary

The year activity heatmap rendered each day's intensity through cell **tint/opacity alone**, with no per-cell hover `help` or accessibility label — so the per-day value was conveyed by a visual channel only (violates ACCESSIBILITY.md: never convey meaning by color/opacity alone). The surface-level VoiceOver summary and the Reduce-Transparency / Privacy-Mask text alternative already existed; this fills the missing **per-cell** affordance.

Each cell in both window grids now carries a textual `.help(...)` (pointer hover) and `.accessibilityLabel(...)` (VoiceOver) derived from the existing `SpendingHeatmap.cellLabel` — date + value + transaction count — with the value **masked under Privacy Mask**.

> **Follow-up (adversarial-review fix, 2nd commit):** the first commit masked only the per-cell `.help`/`.accessibilityLabel`. The same popover view (`BalanceActivityHeatmap`) also renders the **focused-day caption** (`focusedDayCaption(for:)`) from `SpendingHeatmap.focusedDaySummary`, which built its `captionText` (shown via `Text`) and `accessibilityLabel` (announced to VoiceOver) from an **un-masked** amount. Selecting a heatmap cell while Privacy Mask was on therefore still leaked the real value — both visually and to VoiceOver (the cell's onSelect→toggleSelection has no mask guard, so it is reachable live). The second commit closes that remaining half by threading the live mask flag into `focusedDaySummary` too. The header total and the audio graph were already mask-gated; the popover heatmap is now consistent.

## Architectural rationale (single-sourced masked label)

- The per-cell sentence is single-sourced from `PlaidBarCore`'s `SpendingHeatmap.cellLabel`, not a new view-local formatter, so every heatmap surface stays consistent.
- `cellLabel` (and the underlying `amountText`) gained an `isPrivacyMasked` parameter (default `false`, preserving all existing call sites). When masked, the **whole amount token — sign prefix included — collapses to `••••`** (`PrivacyMaskPresentation.compactValue`), while the structural date and the non-financial transaction count remain. This keeps the cell meaningful textually without ever leaking the value.
- `focusedDaySummary` gained the same `isPrivacyMasked` parameter (2nd commit) and threads it into its internal `amountText`/`cellLabel` calls, so its `captionText` and `accessibilityLabel` reuse the exact same mask-collapse path — no new masking logic.
- The two window grids (`DashboardYearHeatmapGrid`, `InsightsActivityHeatmapGrid`) take an `isPrivacyMasked` stored property and thread the live `appState.shouldMaskFinancialValues` in from their call sites, rather than hardcoding a literal.

## Design decisions

- **Both AND-671 surfaces fixed** (`DashboardOverviewColumn.swift`, `InsightsTrendsView.swift`). These grids are only rendered when masking is OFF (the masked branch shows a text alternative), so the flag is `false` in practice — but it's threaded honestly so the affordance is correct by construction if the grid is ever shown while masked (defense in depth).
- **Popover `BalanceHeatmapCell` moved onto the masked source too.** It already had per-cell `.help`/`.accessibilityLabel` (the precedent for this fix) but sourced an **unmasked** `cellLabel`, and unlike the window grids it *does* render its interactive cells while Privacy Mask is on — so its hover/VoiceOver text was a real value leak. Single-line fix onto the same Core source closes it.
- **Popover focused-day caption moved onto the masked source too (2nd commit).** Same popover view, same live-while-masked reachability: `focusedDayCaption(for:)` renders `Text(summary.captionText)` and `.accessibilityLabel(summary.accessibilityLabel)` unconditionally, so a selected cell's value leaked visually and to VoiceOver under Privacy Mask. The `MainPopover` call site now passes `appState.shouldMaskFinancialValues` into `focusedDaySummary`, so both surfaces collapse the amount to `••••`.
- The `ChartAudioGraph.heatmap` caller computes `amountText` only inside its *non-masked* branch (the masked branch already drops the amount and keeps date + count), so it was already safe and is left unchanged.
- Amount-only masking (date + count stay visible) mirrors the existing focused-day caption / audio-graph masking semantics in this codebase.

## Testing notes

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — passes (CI gate).
- Full `swift test`: **2570 tests, 2 failures**, both pre-existing and unrelated to this change — `SyncResponseTests` "Decoding reads present pending cursor fields" and `WebhookDefectAndVerifierTests` "Sticky sync signal clears once the item refresh advances past it" — i.e. exactly the two failures being addressed by the green-main fix in #693. **Zero new failures** from this branch. (Remaining `✘`-glyph lines in the run are Keychain `skipped:` cases, not failures.)
- New tests:
  - `Tests/PlaidBarCoreTests/HeatmapFocusSummaryTests.swift` — per-cell masking (4 cases): unmasked label shows the real amount; **Privacy Mask on** hides the amount (`$`/digits gone, `••••` present) but keeps date + count; masked net-cashflow drops the `+/-` sign prefix entirely; masked zero/empty day still hides `$0` and keeps a zero count. **Focused-day caption masking (2nd commit, 3 cases):** unmasked summary still shows the real amount in both `captionText` and `accessibilityLabel`; **masked** summary hides the amount in **both** surfaces (`$`/digits gone, `••••` present) while keeping date + count; masked net-cashflow caption drops the sign prefix.
  - `Tests/PlaidBarTests/PlaidBarTests.swift` — `activityHeatmapCellsCarryPerCellLabel`: source-invariant guard that both window grids attach `.help(label)` + `.accessibilityLabel(label)` from the mask-aware Core `cellLabel`, declare the `isPrivacyMasked` property, and that both call sites pass the live mask state. **`popoverHeatmapMasksCellHelpAndFocusedCaption` (2nd commit):** source-invariant guard that `MainPopover.swift` threads `isPrivacyMasked`/`shouldMaskFinancialValues` into **both** the popover cell help and the `focusedDaySummary` caption call site (and that `Text(summary.captionText)` / `.accessibilityLabel(summary.accessibilityLabel)` derive from the now-masked summary), so the popover flag-threading can't be silently dropped. These are the regression guards for the `@main` (non-unit-testable) view wiring.
- No existing source-invariant assertion string-matched the heatmap lines that changed (verified by grep), so nothing else needed updating.

## Linked issue

AND-671 (Medium, Accessibility) — https://linear.app/andeslab/issue/AND-671

## Known limitations

- The window grids never render while masked (parent shows a text alternative), so the masked window-cell path is reachable only via the Core unit test, not the live window UI today — the wiring is defensive/future-proof. The popover cell **and the popover focused-day caption** are where masking is exercised live.
- Source-invariant tests assert wiring strings, not rendered AX output (the `@main` app target isn't unit-testable); they string-match the exact `.help(label)` / `.accessibilityLabel(label)` / `focusedDaySummary(...)` lines, so an intentional rename of those will require updating the test alongside the view.

## Follow-ups

- None required. If the legacy popover is removed (Stage 2, AND-598…600), the `BalanceHeatmapCell` masking line and the focused-day-caption masking go away with it; the Core `cellLabel(isPrivacyMasked:)` / `focusedDaySummary(isPrivacyMasked:)` sources stay.

### DIFF
--- a/Sources/PlaidBar/Views/Destinations/DashboardOverviewColumn.swift
+++ b/Sources/PlaidBar/Views/Destinations/DashboardOverviewColumn.swift
@@ -43,7 +43,7 @@ struct DashboardActivityHeatmapCard: View {
             if appState.shouldMaskFinancialValues || reduceTransparency {
                 heatmapTextAlternative(layout: layout)
             } else {
-                DashboardYearHeatmapGrid(layout: layout)
+                DashboardYearHeatmapGrid(layout: layout, isPrivacyMasked: appState.shouldMaskFinancialValues)
             }
         }
         .loadingRedaction(appState.loadState(for: .activityHeatmap))
@@ -449,6 +449,12 @@ private struct DashboardOverviewFallback: View {
 /// under Privacy Mask / Reduce Transparency before reaching this view.
 struct DashboardYearHeatmapGrid: View {
     let layout: SpendingHeatmapLayout
+    /// Whether Privacy Mask is active. The parent only renders this grid when
+    /// masking is OFF (the masked branch shows a text alternative), so this is
+    /// `false` in practice — but it is threaded honestly into the per-cell label
+    /// source so the affordance never leaks a value if the grid is ever shown
+    /// while masked (defense in depth; single Core label source).
+    var isPrivacyMasked: Bool = false
 
     /// Gap between cells. A touch wider than the glance grid's 2pt so the larger
     /// cells read as a clean lattice at desk distance.
@@ -546,9 +552,16 @@ struct DashboardYearHeatmapGrid: View {
     private func cellView(for day: SpendingHeatmapDay?, size: CGFloat) -> some View {
         if let day {
             let intensity = SpendingHeatmap.cellIntensity(for: day, peakValue: layout.peakValue)
+            // Per-cell textual affordance: the cell's meaning otherwise rides on
+            // tint/opacity alone (ACCESSIBILITY.md). The label is the single Core
+            // source (date + masked value + count), so pointer-hover and VoiceOver
+            // get the same sentence and Privacy Mask is honored (AND-671).
+            let label = SpendingHeatmap.cellLabel(for: day, mode: layout.mode, isPrivacyMasked: isPrivacyMasked)
             RoundedRectangle(cornerRadius: Radius.control)
                 .fill(Color.primary.opacity(0.10 + 0.62 * intensity))
                 .frame(width: size, height: size)
+                .help(label)
+                .accessibilityLabel(label)
         } else {
             RoundedRectangle(cornerRadius: Radius.control)
                 .fill(.clear)
--- a/Sources/PlaidBar/Views/Destinations/InsightsTrendsView.swift
+++ b/Sources/PlaidBar/Views/Destinations/InsightsTrendsView.swift
@@ -109,7 +109,7 @@ struct InsightsTrendsView: View {
                     // information without relying on color or translucency.
                     heatmapTextAlternative(layout: layout)
                 } else {
-                    InsightsActivityHeatmapGrid(layout: layout)
+                    InsightsActivityHeatmapGrid(layout: layout, isPrivacyMasked: isMasked)
                 }
             }
             .loadingRedaction(appState.loadState(for: .activityHeatmap))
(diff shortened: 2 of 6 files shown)


ftchvs (commented):

Audited the heatmap accessibility/privacy change against the repo AGENTS guidance. The window heatmap cells now get text equivalents via `.help` and `.accessibilityLabel`, and the live popover path masks both per-cell help/VoiceOver text and the focused-day caption while Privacy Mask is on. The amount token, including +/- sign, collapses to the shared mask glyph while date and transaction count remain. No Plaid credentials, provider tokens, raw Plaid payloads, real account/transaction IDs, local SQLite data, screenshots, or app-side Plaid calls are introduced.

Validation run:
- `swift test --filter HeatmapFocusSummaryTests`

No merge performed.

PR #695, opened 2026-06-26
perf(server): bound unbounded in-memory accumulations (AND-666)

ftchvs:
## Summary

Hardens two low-severity unbounded in-memory accumulations in the localhost server. Both sit behind the auth-gated `127.0.0.1` boundary (so they are contained, not urgent), but both could grow without bound:

1. **Merchant-logo proxy** (`MerchantLogoRoutes`) buffered the whole upstream logo body with no size cap and hardcoded `Content-Type: image/png`.
2. **Stripe billing idempotency set** (`StripeBillingEventStore`) was an insert-only `Set<String>` of processed event ids with no eviction, TTL, or persistence — it grew forever under a stream of distinct events.

## Architectural rationale

Both spots stay in `Sources/PlaidBarServer/` and keep their existing concurrency model — the idempotency store remains an `actor`, the logo proxy stays a `Sendable` struct using its `URLSession`. No new types cross the process/Keychain/localhost boundary, and `PlaidBarCore` is untouched. The fixes are local to the two routes.

The logo size cap is a server-internal proxy detail (not a value shared with the app/widget/CLI), so it lives as a named `static` constant on `MerchantLogoRoutes` rather than in `PlaidBarCore`'s `Constants.swift`. Same reasoning for the idempotency cap on `StripeBillingEventStore`.

## Design decisions

**Logo proxy size cap (2 MB).** Brand logos are a few KB; 2 MB is generous headroom while bounding the in-memory buffer. Enforcement is two-layered: the advertised `Content-Length` is rejected *before* the body is buffered (fast path), and the materialized byte count is re-checked so a chunked or dishonest upstream that omits/understates the header still can't exceed the cap. The cap decision is factored into a pure `exceedsLogoSizeLimit(advertisedContentLength:actualByteCount:)` so it is unit-testable without a live fetch.

**Logo `Content-Type` passthrough.** The served type is now derived from the upstream response: a recognized `image/*` type is passed through; anything missing/odd (e.g. `text/html`) falls back to PNG (the historical default), so a known-good asset never degrades. Cache hits — which store only raw bytes — sniff the type from the payload's magic number (PNG/JPEG/GIF/WebP/SVG), defaulting to PNG.

**Idempotency set bounding (FIFO, default 4096).** Chosen the simplest well-tested option: a FIFO-capped `Set` + insertion-order array on the existing actor. Rationale and correctness argument:
- Stripe's retries for a single event arrive close together (minutes/hours), far inside a 4096-distinct-id window, so de-dup is preserved for the only window that matters in practice.
- The store was already non-persistent across restarts, and that is **correct**: a replay (after eviction or restart) re-applies an *idempotent* `save` of the same normalized status/plan/dates, so billing state is unchanged — only the in-memory footprint is now bounded. Backing it with the Fluent store would add a migration and I/O for no correctness gain over the in-window guarantee, so FIFO is the simpler, lower-risk choice. (Persistence remains a clean follow-up if a longer idempotency window is ever required.)
- ~4096 short `evt_…` ids is at most a few hundred KB — a hard ceiling well above any realistic retry burst.

## Testing notes

New suite `Tests/PlaidBarServerTests/ServerMemoryBoundsTests.swift` (8 tests):
- Idempotency set de-dups within the window; never exceeds capacity across `capacity * 50` distinct inserts; FIFO evicts the oldest id once the window slides; capacity clamped to ≥ 1.
- Logo proxy rejects an over-cap advertised `Content-Length`, rejects an over-cap body even when the header lies or is absent, allows normal-sized logos, passes through `image/*` and falls back to PNG otherwise, and sniffs cached bytes back to the right type.

Commands and results:
```
swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors   # CI gate — PASS
swift test --filter ServerMemoryBoundsTests                                        # 8/8 PASS
swift test                                                                          # full suite
```
Full `swift test`: **2 pre-existing failures only**, both already fixed in pending PR #693 and unrelated to this change:
- `SyncResponseTests` "Decoding reads present pending cursor fields" (date-bomb: expects a 2027 date vs the current system clock).
- `WebhookDefectAndVerifierTests` "Sticky sync signal clears once the item refresh advances past it".

Zero **new** failures introduced by this PR.

## Linked issue

- AND-666 (Low, Tech Debt / Performance) — https://linear.app/andeslab/issue/AND-666

## Known limitations

- The logo size cap relies on `URLSession`'s buffered `data(from:)`, so an upstream that lies about `Content-Length` is still fully downloaded before the byte-count check rejects it (the in-memory buffer is bounded by the cap, but the wire transfer isn't truncated mid-stream). True streaming truncation would require a delegate-based download; not warranted for an allowlisted Plaid CDN behind localhost.
- The idempotency set is still in-memory only; idempotency is guaranteed only within the retained FIFO window, not across restarts (correct by the idempotent-replay argument above).
- The logo size-cap path is covered by a pure decision function rather than an end-to-end fetch (the route's fetch hits a real `URLSession`); the byte/header logic the route uses is exercised directly.

## Follow-ups

- If a longer or restart-durable idempotency window is ever needed, back `StripeBillingEventStore` with the existing Fluent store (add a `CreateStripeBillingEvents` migration + TTL sweep).
- Optional: stream-truncate the logo download via a `URLSessionDataDelegate` to cap wire transfer, not just the buffer.

### DIFF
--- a/Sources/PlaidBarServer/Routes/BillingRoutes.swift
+++ b/Sources/PlaidBarServer/Routes/BillingRoutes.swift
@@ -195,10 +195,48 @@ struct BillingRoutes: Sendable {
     }
 }
 
+/// Bounded, in-memory set of recently processed Stripe webhook event ids used
+/// to make billing-webhook application idempotent.
+///
+/// The set is FIFO-capped at ``capacity`` ids: once full, recording a new id
+/// evicts the oldest. This keeps memory bounded under an unbounded stream of
+/// distinct events while preserving idempotency for the *recent* window — the
+/// only window that matters in practice, since Stripe's retries for a single
+/// event arrive close together (minutes/hours), far inside a multi-thousand-id
+/// window. State is intentionally not persisted across restarts: a replay after
+/// a restart re-applies an idempotent `save` (same normalized status/plan/dates),
+/// so correctness is preserved even when an id is no longer remembered — only the
+/// in-memory footprint is bounded.
 actor StripeBillingEventStore {
-    private var processedEventIDs: Set<String> = []
+    /// Maximum number of recent event ids retained. ~4096 ids (short Stripe
+    /// `evt_…` strings) is a few hundred KB at most, far more than any realistic
+    /// retry window needs, yet a hard ceiling on growth.
+    static let defaultCapacity = 4096
 
-    func recordIfNew(_ eventID: String) -> Bool {
-        processedEventIDs.insert(eventID).inserted
+    private let capacity: Int
+    private var processedEventIDs: Set<String> = []
+    /// Insertion order of `processedEventIDs`, oldest first, for FIFO eviction.
+    private var insertionOrder: [String] = []
+
+    init(capacity: Int = StripeBillingEventStore.defaultCapacity) {
+        self.capacity = max(1, capacity)
     }
+
+    /// Records `eventID` and returns `true` if it had not been seen within the
+    /// retained window. A return of `false` means a duplicate (de-dup hit).
+    func recordIfNew(_ eventID: String) -> Bool {
+        guard processedEventIDs.insert(eventID).inserted else {
+            return false
+        }
+        insertionOrder.append(eventID)
+        if insertionOrder.count > capacity {
+            let evicted = insertionOrder.removeFirst()
+            processedEventIDs.remove(evicted)
+        }
+        return true
+    }
+
+    /// Number of event ids currently retained. Exposed for tests asserting the
+    /// bound holds.
+    var count: Int { processedEventIDs.count }
 }
--- a/Sources/PlaidBarServer/Routes/MerchantLogoRoutes.swift
+++ b/Sources/PlaidBarServer/Routes/MerchantLogoRoutes.swift
@@ -28,9 +28,9 @@ struct MerchantLogoRoutes: Sendable {
     /// Hard ceiling on a single buffered logo response. Brand logos are small
     /// PNG/SVG assets (a few KB); 2 MB leaves generous headroom while bounding
     /// the in-memory buffer so a misbehaving or compromised upstream can't pin
-    /// arbitrary memory in the proxy. Enforced both via the advertised
-    /// `Content-Length` (fast reject before the body streams) and the actual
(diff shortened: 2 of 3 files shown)


ftchvs:
**Independent adversarial review: ✅ APPROVE with nits (high confidence).** Deep analysis confirms correctness: FIFO de-dup is sound (Set insert gates the insertionOrder append, so duplicates never refresh position; eviction removes oldest from both), actor isolation keeps it Sendable-clean, and the double-process risk is benign — `BillingSubscriptionStore.save` is a verified idempotent singleton-row upsert and crossing 4096 distinct events within Stripe's minutes-apart retry window is unrealistic. Logo cap holds against a lying/chunked upstream (advertised Content-Length OR materialized byte-count), and Content-Type passthrough is safe because the sole consumer `NSImage(data:)` ignores the HTTP Content-Type entirely (no SVG-script execution). Tests truly guard the bound (reverting the cap fails the eviction tests). Non-blocking nits, all pre-existing/disclosed: (1) the helper call-sites in `cachedImageData`/`serveLogo` aren't integration-tested (a false-guard surface if the wiring were removed); (2) the advertised-Content-Length 'fast reject before buffering' framing is aspirational — `session.data(from:)` already materializes the body, so the byte-count branch is what actually bounds memory; (3) FIFO eviction slightly widens the pre-existing out-of-order replay window. None block. Ready for your review.

ftchvs (commented):

The merchant-logo cap was still using `URLSession.data(from:)`, so a chunked or understated upstream body could be fully buffered before the size helper ran. That left the logo proxy unbounded in memory for the exact case this PR was meant to harden. I pushed `2e5c963d` to stream via `URLSession.bytes(from:)`, validate the final response before body reads, reject honest oversized `Content-Length` before streaming, and stop appending before the local buffer can exceed `maxLogoBytes`.

Validation run after the fix:
- `git diff --check`
- `swift test --filter ServerMemoryBoundsTests`
- `swift build --target PlaidBarServer --skip-update --disable-keychain -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors`

No merge performed.

PR #696, opened 2026-06-26
fix(investments): holdings returns setup-state instead of misleading empty success (AND-661)

ftchvs:
## Summary

`/api/investments/holdings` (and `/api/investments/transactions`) returned `200` with an empty payload when no Plaid credentials were configured, instead of the `503` setup state every other Plaid-backed route returns. The app then rendered a misleading *empty success* rather than the credential-guidance setup state. This is the AND-644 Investments follow-up.

Two defects, both mirrored from the correct `AccountRoutes` pattern:

1. The `investments` route prefix was **omitted from `SetupStateMiddleware.plaidBackedPathPrefixes`**, so with no items linked the per-item route skipped Plaid entirely and returned an empty `200`.
2. The handler **swallowed every per-item error, including `PlaidError.credentialsNotConfigured`**, so even an already-linked item could not surface the setup state.

## Architectural rationale (mirrors AccountRoutes)

`SetupStateMiddleware` maps the credential-less setup state to a clear `503` so clients can tell "server is up but awaiting credentials" apart from a real failure or a successful empty refresh. It does this with two layers, both of which `AccountRoutes` already uses and `InvestmentRoutes` was missing:

- **Prefix gating** — `plaidBackedPathPrefixes` blocks Plaid-backed paths before the handler runs, covering the no-items-linked case where the handler would otherwise return empty-but-`200`. Added `"/api/investments"` alongside `"/api/link"`, `"/api/accounts"`, `"/api/transactions"`.
- **Handler re-throw (safety net)** — for items already linked, the handler reaches the live Plaid call; `AccountRoutes` re-throws `PlaidError.credentialsNotConfigured` so the middleware's `catch` converts it to the `503`. `InvestmentRoutes` now re-throws it identically (in both `getHoldings` and `getInvestmentTransactions`) instead of folding it into an empty contribution.

## Design decisions

- **Re-throw `credentialsNotConfigured` in both handlers, not just holdings.** The same swallow-all `catch` and the same setup-state contract apply to investment transactions; fixing only holdings would leave `/api/investments/transactions` reporting an empty `200` in setup state. The prefix gate already covers both routes, so re-throwing in both keeps the handler-level safety net consistent with the gate.
- **Other per-item errors still yield an empty contribution.** A missing `investments` product scope (the common case for existing links) remains best-effort: it returns an empty `InvestmentsResponse`/`[]` so a scope gap on one item never fails the whole request. Only `credentialsNotConfigured` — which affects every item identically — propagates. This exactly mirrors `AccountRoutes.listLiabilities` / `refreshAccounts`.
- **Smallest correct change.** One prefix string + two `catch PlaidError.credentialsNotConfigured` clauses; no behavior change for the configured/happy path.

## Testing notes

- `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — passes (CI gate).
- `swift test` (full suite) — 2567 tests, 2 failures, both **pre-existing** and unrelated (fixed in pending PR #693): `SyncResponseTests` `decodePendingCursors` and `WebhookDefectAndVerifier` sticky-sync. **Zero new failures** from this change.
- Tests in `Tests/PlaidBarServerTests/`:
  - Extended `setupStateMiddlewareClassifiesPlaidBackedPaths` (in `PlaidBarServerTests`) to assert `/api/investments/holdings` and `/api/investments/transactions` are classified Plaid-backed.
  - `Investment routes setup state` suite (in `InvestmentRoutesTests`) wires the real `SetupStateMiddleware` + `InvestmentRoutes` through a `Router`/`Application.test`.

### Test-fidelity follow-up (true guard for the handler re-throw)

A review found that the original four setup-state tests all wired `SetupStateMiddleware` with `credentialDiagnosis: .missingBoth`. In that state the `/api/investments` **prefix gate short-circuits to `503` before the handler ever runs**, so the two `seedItem: true` tests that *claimed* to guard the handler re-throw did not: reverting **only** the handler's `catch PlaidError.credentialsNotConfigured` re-throw left all four green. The re-throw was effectively unguarded.

This follow-up commit:

- Parameterizes the suite's `withInvestmentsAPI` helper with `credentialDiagnosis` (default `.missingBoth`, so the existing prefix-gate tests are unchanged) and **renames** the prior `seedItem: true` tests to reflect that they guard the **prefix gate only**.
- Adds two **true-guard** tests (`holdingsRethrowsCredentialsNotConfiguredWhenCredentialsConfigured`, `transactionsRethrowCredentialsNotConfiguredWhenCredentialsConfigured`) with `credentialDiagnosis: .configured`. With credentials present the prefix gate is bypassed, so a linked item actually reaches the credential-less Plaid double — the only path to `503` is then the handler re-throw → the middleware's *outer* `catch`.
- **Red-on-revert verified:** reverting *only* the two handler `catch PlaidError.credentialsNotConfigured` clauses turns these two new tests red (handler swallows the error → `200 OK` empty instead of `503`), while the four `.missingBoth` tests stay green — proving the new tests genuinely guard the re-throw. Production code is unchanged by this commit (the revert was reverted).

## Linked issue

- AND-661 (Medium) — AND-644 Investments follow-up
- https://linear.app/andeslab/issue/AND-661

## Known limitations

- The behavioral suite drives a credential-less `PlaidClientProtocol` double rather than a live Plaid call, so it asserts the server-side contract (status code surfaced to the app), not Plaid's own error wire format.
- The "one linked item" path uses a direct (non-`keychain:`) stored token so token resolution succeeds without macOS Keychain, keeping the test deterministic and Keychain-independent.

## Follow-ups

- None required. The investment-transactions pagination cap is explicitly **out of scope** — already handled by AND-659 / PR #681.

### DIFF
--- a/Sources/PlaidBarServer/Middleware/SetupStateMiddleware.swift
+++ b/Sources/PlaidBarServer/Middleware/SetupStateMiddleware.swift
@@ -43,7 +43,7 @@ struct SetupStateMiddleware<Context: RequestContext>: RouterMiddleware {
     /// credentials. Status and item readiness endpoints are deliberately
     /// absent: they serve local metadata that setup guidance relies on.
     static var plaidBackedPathPrefixes: [String] {
-        ["/api/link", "/api/accounts", "/api/transactions"]
+        ["/api/link", "/api/accounts", "/api/transactions", "/api/investments"]
     }
 
     static func isPlaidBackedPath(_ path: String) -> Bool {
--- a/Sources/PlaidBarServer/Routes/InvestmentRoutes.swift
+++ b/Sources/PlaidBarServer/Routes/InvestmentRoutes.swift
@@ -35,10 +35,16 @@ struct InvestmentRoutes: Sendable {
                 let accessToken = try tokenStore.accessToken(for: item)
                 let response = try await plaidClient.getInvestmentHoldings(accessToken: accessToken)
                 return Self.investmentsResponse(from: response, item: item, itemId: itemId)
+            } catch PlaidError.credentialsNotConfigured {
+                // Setup state affects every item identically: surface the 503
+                // credential guidance (via `SetupStateMiddleware`) instead of
+                // swallowing it into a misleading empty-but-successful refresh.
+                // Mirrors `AccountRoutes`.
+                throw PlaidError.credentialsNotConfigured
             } catch {
-                // A missing `investments` scope (or any per-item failure) yields
-                // an empty contribution; the request still succeeds for items
-                // that do hold investments.
+                // A missing `investments` scope (or any other per-item failure)
+                // yields an empty contribution; the request still succeeds for
+                // items that do hold investments.
                 return InvestmentsResponse()
             }
         }
@@ -62,6 +68,10 @@ struct InvestmentRoutes: Sendable {
                     endDate: endDate
                 )
                 return transactions.map(Self.investmentTransactionDTO(from:))
+            } catch PlaidError.credentialsNotConfigured {
+                // Setup state surfaces the 503 credential guidance instead of an
+                // empty-but-successful refresh. Mirrors `AccountRoutes`.
+                throw PlaidError.credentialsNotConfigured
             } catch {
                 return []
             }
--- a/Tests/PlaidBarServerTests/InvestmentRoutesTests.swift
+++ b/Tests/PlaidBarServerTests/InvestmentRoutesTests.swift
@@ -344,13 +344,20 @@ struct InvestmentRoutesSetupStateTests {
         return headers
     }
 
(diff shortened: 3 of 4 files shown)


ftchvs (commented):

The investments setup-state fix now covers both layers: `/api/investments` is classified as Plaid-backed for the credentialless/no-item path, and both holdings and investment-transaction handlers rethrow `PlaidError.credentialsNotConfigured` once the handler is reached. Other per-item investment errors remain best-effort empty contributions, preserving the missing-scope behavior. No token custody or app-side Plaid boundary changes are introduced.

Validation run:
- `swift test --filter InvestmentRoutesSetupStateTests`

No merge performed.

PR #698, opened 2026-06-26
fix(state): AppState/cache concurrency hardening (AND-670)

ftchvs:
## Summary

Hardens four concurrency/robustness gaps in the AppState refresh/snapshot seam and the disposable PlaidBarCache stores (Linear AND-670, Medium). Each change is surgical and behavior-preserving except the two intended write-wins changes (single-flight + generation guard). All four cited locations were re-verified against current code (line numbers had shifted). No restructuring of the refresh pipeline, the security split, the Plaid client, or the Keychain/localhost boundary.

### 1. `refreshDashboard` had no reentrancy guard — single-flighted
**What:** Added an `inFlightRefreshTask` handle. `refreshDashboard(force:)` now chains off it (awaits any in-flight body, then runs its own as the new handle); the body was extracted unchanged into `performRefreshDashboard(force:)`.
**Why it was a bug:** The refresh loop, ~15 ad-hoc `Task { await refreshDashboard() }` spawns (command palette, recovery actions, popover/glance buttons, settings), and the glance-command path could all run concurrently. Two refresh bodies then interleaved their account/transaction reconciliation, so a slower in-flight refresh could commit *after* — and clobber — a newer result.
**What now wins on a race:** Bodies are serialized (at most one runs at a time); the **most recently requested** refresh is the one that lands last. Awaiting (not cancelling) the prior task preserves the existing semantics that an already-running refresh always completes. Every caller still `await`s to completion, so the "returns once refreshed" contract is unchanged.
**Tested:** Guarded by reasoning + review (see Testing notes) — `AppState` is in the `@main` app target and is not unit-testable directly; the single-flight gate could not be cleanly extracted into a testable seam without restructuring the refresh pipeline (explicitly out of scope). Verified by the strict-concurrency build and the full suite.

### 2. Unguarded finance-snapshot detached save vs generation token — generation-guarded
**What:** `writeGlanceSnapshot` now bumps and captures `glanceSnapshotWriteGeneration` once, up front, and threads it into `writeFinanceSnapshot(updatedAt:generation:)`. The App Group `FinanceSnapshot` save re-checks the token on the MainActor before committing, exactly mirroring the sibling glance-snapshot `Task`.
**Why it was a bug:** The finance save ran in a `Task.detached` that was **not** gated on the generation token, while the glance save was. A clear path (`clearGlanceSnapshot` / `clearPublishedSystemSnapshotsForDemoEntry`) bumps the generation and wipes the App Group files; a stale in-flight finance save could land afterward and resurrect pre-clear figures onto Control Center / Siri / Spotlight / Shortcuts.
**What now wins on a race:** The newest write/clear wins — a stale finance save whose captured generation no longer matches drops itself. File IO still runs off the main actor, preserving the non-blocking contract.
**Tested:** Same untestability constraint as #1 (app target). Guarded by mirroring the already-reviewed glance-snapshot guard pattern + the source-invariant tests in `PlaidBarTests` (which still pass).

### 3. Cache decode failure not contained — corrupt row payload is now a disposable miss
**What:** `ReadModelCacheStore.load(cacheKey:)` wraps the inner per-row `DashboardReadModel` payload decode in do/catch; an undecodable payload purges the row and returns `nil` instead of throwing.
**Why it was a bug:** The whole-store file decode in `loadedRows()` already self-heals to empty, but the per-row payload decode (line ~167) threw on a corrupt/incompatible blob. A disposable cache should treat that as a miss.
**Behavior-preserving:** The `load` signature is unchanged (still `throws` for the other paths); callers using `try?` are unaffected. The disposable cache never permanently disables itself and never loses real data — the authoritative source is the live in-memory data.
**Tested:** New PlaidBarCache test feeds a row with a corrupt inner payload (Snapshot decodes fine, payload does not) → asserts `load` returns `nil` (not throw), the bad row is purged, and a subsequent write rebuilds a clean store that re-opens correctly.

### 4. `BudgetingV2Store` eager init — made lazy
**What:** `rowsByCacheKey` is now optional and lazily hydrated via a `loadedRows()` helper; the initializer is a pure non-throwing handoff (no disk read). An incompatible/corrupt file self-heals into a clean miss. The `throws` init signature is retained for source compatibility.
**Why it was a bug:** The init eagerly read + decoded the on-disk snapshot (and could throw there), so a cold/missing/corrupt store did work — or failed — before the v2 schema was ever needed. This mirrored the old `ReadModelCacheStore` problem AND-656 finding 3 already fixed.
**Behavior-preserving:** Keeping the `throws` signature means the factory (`+Factory.swift`), the app consumer (`AppState+CategoryEditor.swift`, `CategoryEditorModel`), and all existing tests are untouched. v1 budgeting and the opt-in/opt-out round-trip are unchanged.
**Tested:** New PlaidBarCache tests — opening does zero disk I/O (proven via a `fileExists`-counting `FileManager`), and opening over an incompatible file does no work + reads as not-opted-in, then reseeds cleanly.

## Architectural rationale

- **Mirrored existing patterns.** #1 reuses the stored-`Task` handle idiom (cf. `refreshTask`, `localAISummaryRefreshTask`). #2 reuses the `glanceSnapshotWriteGeneration` re-check the glance `Task` already does. #4 reuses `ReadModelCacheStore`'s lazy `loadedRows()` self-heal (AND-656). #3 extends the same disposable-miss philosophy from the whole-store decode to the per-row decode.
- **Disposable caches stay disposable.** Every cache failure degrades to a miss; the live in-memory data is always authoritative.
- **No boundary changes.** PlaidBarCore, the server, the Plaid client, and the Keychain/localhost split are untouched.

## Testing notes

- Strict-concurrency CI gate: `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` → **Build complete** (passes).
- Full `swift test` (not filtered): **`Test run with 2564 tests failed after … with 2 issues`**. The 2 issues are the known pre-existing #693 failures, both outside this diff:
  - `SyncResponseTests.decodePendingCursors` — a date-bomb (`Date(timeIntervalSince1970: 1_800_000_100)` expected 2027, system reads 2058); **proven to fail identically on the clean tree with my changes stashed**.
  - `WebhookDefectAndVerifierTests` "Sticky sync signal clears once the item refresh advances past it" — a 20ms-sleep timing test, untouched by this diff.
- **Zero NEW failures.** All 3 new PlaidBarCache tests pass (the full `PlaidBarCacheTests` suite is 56 tests, green).
- Source-invariant assertions in `PlaidBarTests` that read `AppState.swift` were checked: my edits don't touch any of the asserted strings (the `GlanceSnapshot.make` block, `PrivacyMaskControlCommandReader.peek()?.maskEnabled`, `shouldShowWindowFirstOrientation`, etc.). They still pass.

## Linked issue

AND-670 — https://linear.app/andeslab/issue/AND-670

## Known limitations

- **#1 and #2 are not directly unit-tested.** `AppState` lives in the `@main` `PlaidBar` app target, which is not in the test binary. Extracting a testable generation-gate/single-flight seam would have required restructuring the refresh pipeline (out of scope for this surgical fix). They are guarded by mirroring existing, already-reviewed patterns and validated by the strict build + full suite. No sub-fix was deferred — all four were applied.
- #1 changes timing for previously-concurrent UI refreshes: rapid taps now serialize rather than overlap. This is the intended fix (no interleaved reconciliation) and each tap still completes a real refresh.

## Follow-ups

- If AppState refresh/snapshot logic gets extracted toward `PlaidBarCore` in a future change, add direct unit tests for the single-flight gate (#1) and the generation gate (#2).

### DIFF
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -679,6 +679,16 @@ final class AppState {
     private var localAIInsightsService = LocalAIInsightsService()
     private let notificationService: any NotificationServiceProtocol
     private var refreshTask: Task<Void, Never>?
+    /// Single-flight handle for `refreshDashboard(force:)` (AND-670). The refresh
+    /// loop, ad-hoc `Task { await refreshDashboard() }` spawns (command palette,
+    /// recovery actions, glance/popover buttons), and the glance-command path can
+    /// all request a refresh concurrently. Without a guard, two refresh bodies
+    /// interleave their account/transaction reconciliation, so a slower in-flight
+    /// refresh can commit *after* — and clobber — a newer result. Each call chains
+    /// off this handle (awaits the prior body, then runs its own as the new
+    /// handle), so refresh bodies are serialized: at most one runs at a time and a
+    /// newer request's result is always the one that lands last.
+    private var inFlightRefreshTask: Task<Void, Never>?
     private var localAISummaryRefreshTask: Task<Void, Never>?
     /// Observer tokens for the OS power-state / thermal-state change
     /// notifications that drive energy-aware background refresh (AND-568).
@@ -2861,7 +2871,37 @@ final class AppState {
     /// automatic one (the background loop — fetches only when due per
     /// `automaticRefreshPolicy`). Connectivity is re-probed either way so the
     /// loop keeps self-healing and notifications stay current.
+    ///
+    /// Single-flighted (AND-670): the body runs through `inFlightRefreshTask` so
+    /// the refresh loop, ad-hoc `Task { await refreshDashboard() }` spawns, and the
+    /// glance-command path can never interleave two reconciliations. Each call
+    /// awaits any in-flight body before starting its own, so the result that lands
+    /// last is always the most recently requested one. The `await` here still
+    /// completes only after a refresh body finished, so every caller (button taps,
+    /// recovery actions, the loop) keeps its existing "returns once refreshed"
+    /// contract.
     func refreshDashboard(force: Bool = true) async {
+        // Chain off any in-flight refresh so two bodies never run concurrently.
+        // Capturing the prior handle and awaiting it (rather than cancelling)
+        // preserves the existing semantics: an already-running refresh always
+        // completes, and this newer request runs to completion right after it.
+        let previous = inFlightRefreshTask
+        let task = Task { @MainActor in
+            await previous?.value
+            await performRefreshDashboard(force: force)
+        }
+        inFlightRefreshTask = task
+        await task.value
+        // Clear the handle only when this task is still the latest one, so a
+        // newer chained request keeps the link to its predecessor intact.
+        if inFlightRefreshTask == task {
+            inFlightRefreshTask = nil
+        }
+    }
+
+    /// The actual refresh body. Always invoked through `refreshDashboard(force:)`'s
+    /// single-flight gate (AND-670) so it never runs concurrently with itself.
+    private func performRefreshDashboard(force: Bool) async {
         if refreshDemoDataIfNeeded() { return }
 
         if await consumePendingGlanceCommand() { return }
--- a/Sources/PlaidBar/App/AppState.swift
+++ b/Sources/PlaidBar/App/AppState.swift
@@ -4291,9 +4291,20 @@ final class AppState {
         }
         let systemSurfaceMaskEnabled = shouldMaskFinancialValues
             || ((try? PrivacyMaskControlCommandReader.peek()?.maskEnabled) == true)
+        // Bump and capture the write generation up front so *both* off-actor
+        // publishes below — the App Group finance snapshot and the glance snapshot
+        // — are gated on the same token (AND-670). A clear path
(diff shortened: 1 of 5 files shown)


ftchvs (commented):

The original finance snapshot generation guard checked `glanceSnapshotWriteGeneration` before starting a detached `AppGroupSnapshotStore.save`. A clear could still run after that guard while the detached file write was in flight, then the stale write could land afterward and resurrect pre-clear finance figures on App Group-backed system surfaces. I pushed `f57b1bc9` so the App Group commit now happens immediately after the MainActor generation check on the same executor as clears, and added a source-invariant test that prevents a `Task.detached` gap between the guard and `AppGroupSnapshotStore.save(snapshot)`.

Validation run after the fix:
- `git diff --check`
- `swift test --filter CacheStoreOpenRobustnessTests`
- `swift test --filter financeSnapshotCommitIsGenerationGatedBeforeSave`

No Plaid credential/token custody changes, raw Plaid payload exposure, real account/transaction IDs, local SQLite data movement, screenshots, or app-side Plaid API calls were introduced. No merge performed.

PR #708, opened 2026-06-26
feat(window): persist & restore primary window frame + appearance-before-first-paint test (AND-593)

ftchvs:
## Summary

Completes the remaining slice of **AND-593** — *"Authoritative appearance owner + window restoration (no first-paint flash)"*. The authoritative `NSApp.appearance` owner already shipped with Epic 1 (AND-620); this PR delivers the two residual pieces:

1. **Window position/size persist and restore across relaunch.** There was no `frameAutosaveName` / restoration wiring anywhere in the window scene (verified on `main`). This adds `MainWindowFrameAutosaver`, a zero-size `NSViewRepresentable` that reaches the declarative `Window`'s real `NSWindow` — mirroring the established `PopoverLeadingEdgeAnchor` bridge and the legacy AppKit window controllers (`CategoryDashboardWindowController` et al., which already set a `frameAutosaveName`) — and applies a stable `setFrameAutosaveName("VaultPeekMainWindow")` once. AppKit then writes the frame to `UserDefaults` on move/resize and restores it on the next launch. It's attached to the primary `Window` scene root next to the existing activation-policy bridge.

2. **A launch test asserting appearance is set before first window paint.** Added a source-invariant test (the PlaidBar app target isn't in the unit-test binary) asserting `applyStoredAppearance()` runs inside the synchronous shell `init()` *before* `AppState`/scene construction, that it targets `NSApplication.appearance`, and that the primary window folds into that single authoritative writer.

## Acceptance criteria addressed

- [x] Window position/size persist and restore across relaunch — stable `frameAutosaveName` on the primary `NSWindow`.
- [x] A launch test asserts appearance is set before first window paint — ordering pinned structurally (`applyStoredAppearance()` precedes `AppState()` in `init()`).
- [x] No theme flash on launch (already DONE via AND-620; ordering now regression-pinned).

## Design notes

- The autosaver is **idempotent** (no-ops when the name is unchanged) and **flag-gated**: the primary `Window` only opens behind `WindowFirstFeatureFlag` (`.defaultLaunchBehavior(.suppressed)` + gated opener), so the popover-first escape-hatch path is byte-identical to before.
- `.defaultSize` only supplies the *first-ever* frame; the autosave name is what makes the user's choice durable. `.restorationBehavior(.automatic)` (already present) and the explicit autosave name are complementary — the explicit wiring is what the acceptance criterion calls for and what the test pins.
- The autosave name `"VaultPeekMainWindow"` is distinct from the legacy detached/Category/Review window names so each window keeps its own remembered frame.

## Test evidence

- Strict CI gate: `swift build -Xswiftc -strict-concurrency=complete -Xswiftc -warnings-as-errors` — **Build complete** (clean).
- Full suite: `swift test` — **2628 tests passed**, 0 failures.
- TDD: the frame-autosave test was written first and observed failing (red) before the implementation landed; the appearance-ordering test pins already-shipped wiring.

### DIFF
--- a/Sources/PlaidBar/App/MainWindowFrameAutosaver.swift
+++ b/Sources/PlaidBar/App/MainWindowFrameAutosaver.swift
@@ -1,0 +1,62 @@
+import AppKit
+import SwiftUI
+
+/// Persists and restores the window-first primary `Window`'s position and size
+/// across relaunch (AND-593).
+///
+/// SwiftUI's declarative `Window` scene gives no `NSWindowController` and no
+/// hook for the host window's frame, so — exactly like ``PopoverLeadingEdgeAnchor``
+/// and the legacy AppKit window controllers (`CategoryDashboardWindowController`,
+/// `ReviewTableWindowController`, `DetachedDashboardWindowController`, which all
+/// set a `frameAutosaveName`) — this reaches the real `NSWindow` through a tiny
+/// `NSViewRepresentable` bridge and applies a stable
+/// ``NSWindow/setFrameAutosaveName(_:)`` once. AppKit then writes the frame to
+/// `UserDefaults` whenever the user moves/resizes the window and restores it on
+/// the next launch. `.defaultSize` only supplies the *first-ever* frame; the
+/// autosave name is what makes the choice durable.
+///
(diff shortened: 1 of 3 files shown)


ftchvs:
@codex review

